PR #69, opened 2020-05-14
Add displaying function triggers to tree

### LINKED ISSUE #92: Show list of deployed functions
Would be nice to see list of deployed functions

possibly have a command to navigate to do a quick test
{staticwepapp-defaulthostname}/api/funcname

nturinski:
Fixes #92 

Someone in an email was confused that there wasn't a Functions tree item even though they needed Azure Functions as a dependency.  I thought that this might help with the confusion?

image


EricJizbaMSFT:
Well at least some of this discussion got ironed out offline ("function" vs. "endpoint"). @nturinski is this ready for re-review now that v0.1.0 is out or are you planning on changing any of it?

nturinski:
I think this is okay for review atm.  All it does is list the functions at this point.  I don't think we'd change the UI or language around it (unless the portal does something different now)

PR #112, opened 2020-06-02
Remove dependency, prompt to install

### LINKED ISSUE #108: Remove Functions as a hard dependency
A lot of people won't be using Functions to begin with so change it so that the Functions extension is only installed when a customer goes to add an http function.

nturinski:
Fixes #108

### DIFF
--- a/src/commands/createHttpFunction.ts
+++ b/src/commands/createHttpFunction.ts
@@ -6,7 +6,7 @@
 import * as fse from 'fs-extra';
 import * as path from 'path';
 import * as vscode from 'vscode';
-import { IActionContext } from "vscode-azureextensionui";
+import { IActionContext, UserCancelledError } from "vscode-azureextensionui";
 import { AzureExtensionApiProvider } from 'vscode-azureextensionui/api';
 import { defaultApiName } from '../constants';
 import { ext } from '../extensionVariables';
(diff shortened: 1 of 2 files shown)


src/commands/createHttpFunction.ts
```diff
@@ -65,3 +62,29 @@ function generateSuffixedName(preferredName: string, i: number): string {
...
+async function getFunctionsApi(context: IActionContext): Promise<AzureFunctionsExtensionApi> {
+    const funcExtensionId: string = 'ms-azuretools.vscode-azurefunctions';
+    const funcExtension: vscode.Extension<AzureExtensionApiProvider> | undefined = vscode.extensions.getExtension(funcExtensionId);
+
+    if (funcExtension) {
+        if (!funcExtension.isActive) {
+            await funcExtension.activate();
+        }
+
+        // The Functions DB extension just recently added support for 'AzureExtensionApiProvider' so we should do an additional check just to makes sure it's defined
```
EricJizbaMSFT, line 75: This is not true - it was added a year and a half ago: https://github.com/microsoft/vscode-azurefunctions/pull/812. You don't need the extra check for `funcExtension.exports`

src/commands/createHttpFunction.ts
```diff
@@ -65,3 +62,29 @@ function generateSuffixedName(preferredName: string, i: number): string {
...
+        // tslint:disable-next-line: strict-boolean-expressions
+        if (funcExtension.exports) {
+            return funcExtension.exports.getApi<AzureFunctionsExtensionApi>('^1.0.0');
```
EricJizbaMSFT, line 78: You need to keep the version at `^1.1.0`

`createFunction` wasn't added until 1.1.0: https://github.com/microsoft/vscode-azurefunctions/pull/2077/files#diff-45327f86d4438556066de133327f4ca2R87

src/commands/createHttpFunction.ts
```diff
@@ -65,3 +62,29 @@ function generateSuffixedName(preferredName: string, i: number): string {
...
+        }
+    }
+
+    await ext.ui.showWarningMessage(localize('funcInstall', 'You must have the "Azure Functions" extension installed to perform this operation.'), { title: 'Install' });
+    const commandToRun: string = 'extension.open';
+    vscode.commands.executeCommand(commandToRun, funcExtensionId);
+
+    context.errorHandling.suppressDisplay = true;
+    context.telemetry.properties.installFunctions = 'true';
+    // we still need to throw an error even if the user installs
```
EricJizbaMSFT, line 88: Use `cancelStep` so that this can show up in existing queries, otherwise we might never write a query to check this new property `installFunctions`
```
context.telemetry.properties.cancelStep = 'installFunctions';
throw new UserCancelledError();
```

Also you don't need `suppressDisplay` if you treat this as a cancel.

Issue #95, opened 2020-05-16
Ability to navigate to attached GitHub Repo


anganti:
Is it possible to add a command to navigate to attached GitHub repo for the StaticWebApp

![image](https://user-images.githubusercontent.com/42851022/82109572-f9c2de00-96eb-11ea-92ad-8cdff3f2495f.png)

PR #113, opened 2020-06-04
Add command to view GitHub repo/branch

nturinski:
Fixes https://github.com/microsoft/vscode-azurestaticwebapps/issues/95

Rather than just taking the user to the attached repo, I added the branch information so it'd open the site with that specific branch up.  I think that's actually a little more useful (especially since the command exists on each environment)

### DIFF
--- a/src/commands/viewGitHubRepo.ts
+++ b/src/commands/viewGitHubRepo.ts
@@ -1,0 +1,17 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { IActionContext } from 'vscode-azureextensionui';
+import { ext } from '../extensionVariables';
+import { EnvironmentTreeItem } from '../tree/EnvironmentTreeItem';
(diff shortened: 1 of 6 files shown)


src/commands/viewGitHubRepo.ts
```diff
@@ -0,0 +1,17 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { IActionContext } from 'vscode-azureextensionui';
+import { ext } from '../extensionVariables';
+import { EnvironmentTreeItem } from '../tree/EnvironmentTreeItem';
+import { openUrl } from '../utils/openUrl';
+
+export async function viewGitHubRepo(context: IActionContext, node?: EnvironmentTreeItem): Promise<void> {
```
EricJizbaMSFT, line 11: package.json says `viewItem =~ /azureStatic(WebApp|Environment)$/`, which makes me think the type for `node` should be more than just `EnvironmentTreeItem `

nturinski: Actually, the package.json is wrong.  I still get confused since I changed WebApp to be the overall app, but production is actually an environment.

PR #119, opened 2020-06-06
Use AppSettingClient

nturinski:


src/commands/appSettings/AppSettingsClient.ts
```diff
@@ -0,0 +1,39 @@
...
+
+import { IAppSettingsClient } from 'vscode-azureappservice';
+import { ISubscriptionContext } from 'vscode-azureextensionui';
+import { EnvironmentTreeItem } from '../../tree/EnvironmentTreeItem';
+import { requestUtils } from '../../utils/requestUtils';
+
+export class AppSettingsClient implements IAppSettingsClient {
+
+    public fullName: string;
+    public isLinux: boolean;
```
EricJizbaMSFT, line 14: I don't see you setting `isLinux` in the constructor, but the type here makes it seem required

package.json
```diff
@@ -232,12 +232,12 @@
...
                 {
                     "command": "staticWebApps.viewProperties",
-                    "when": "view == staticWebApps && viewItem =~ /azureStatic(WebApp|Environment|AppSettings)$/",
+                    "when": "view == staticWebApps && viewItem =~ /azureStatic(WebApp|Environment)|applicationSettings$/",
                     "group": "9@1"
                 },
                 {
                     "command": "staticWebApps.openInPortal",
-                    "when": "view == staticWebApps && viewItem =~ /azureStatic(WebApp|AppSettings|Functions)$/",
+                    "when": "view == staticWebApps && viewItem =~ /azureStatic(WebApp|Functions)|applicationSettings$/",
```
EricJizbaMSFT, line 240: If you're going for consistency with the previous stuff, I think you need the `$` in both alternatives: `/azureStatic(WebApp|Functions)$|applicationSettings$/`

That being said, not sure it was actually needed in the first place

src/commands/viewProperties.ts
```diff
@@ -3,25 +3,18 @@
...
 import { ext } from '../extensionVariables';
 import { IAzureResourceTreeItem } from '../tree/IAzureResourceTreeItem';
 import { StaticWebAppTreeItem } from '../tree/StaticWebAppTreeItem';
-import { localize } from '../utils/localize';
-import { nonNullProp } from '../utils/nonNull';
 
-export async function viewProperties(context: IActionContext, node?: IAzureResourceTreeItem): Promise<void> {
-    if (!node) {
-        node = await ext.tree.showTreeItemPicker<StaticWebAppTreeItem>(StaticWebAppTreeItem.contextValue, context);
+export async function viewProperties(context: IActionContext, target?: IAzureResourceTreeItem | AppSettingsTreeItem): Promise<void> {
```
EricJizbaMSFT, line 12: Lol I honestly didn't realize you added "View properties" to application settings. You went hog-wild with "View properties", didn't you? Unfortunately, the name doesn't make sense to me in this scenario. I'm viewing the settings themselves, not properties on the settings. If you want to make it more discoverable, I would maybe do "View as JSON"

src/commands/viewProperties.ts
```diff
@@ -3,25 +3,18 @@
...
     }
 
-    if (!node.data) {
-        if (node.getDataImpl) {
-            await node.getDataImpl();
-        } else {
-            throw new Error(localize('No data exists on resource "{0}"', node.label));
-        }
-    }
+    const data: {} = target instanceof AppSettingsTreeItem ? await target.client.listApplicationSettings() : target.data;
```
EricJizbaMSFT, line 17: I liked it better when this file didn't have to know about any tree item in particular. I suggest you create a simple class that extends `AppSettingsTreeItem` and implements `getDataImpl`

src/commands/viewProperties.ts
nturinski: Hmm, should I just get rid of the `View as Json`?  I realized that this will end up being similar to "Download Remote Settings..." so it seems like a redundancy.

EricJizbaMSFT: Do we already support "Download Remote Settings..."? If yes, you could look at telemetry of both to help us decide. Otherwise yeah I'm fine removing it

nturinski: We don't support it yet, but I believe we have plans to for v0.2.0.

PR #123, opened 2020-06-10
Changelog and readme

nturinski:


CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
 # Change Log
 
+## 0.1.1 - 2020-06-10
+
+### Added
+- Display Function triggers to tree
+
+### Changed
+- Visual Studio Code - Insiders is no longer a requirement
+- Azure Functions extension is no longer a direct dependency
```
EricJizbaMSFT, line 10: I would mention how you changed the default app folder. If CTI was confused by it, I could see other users confused, too

CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
 # Change Log
 
+## 0.1.1 - 2020-06-10
+
+### Added
+- Display Function triggers to tree
```
EricJizbaMSFT, line 6: "to tree" or "in tree"?

CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
 # Change Log
 
+## 0.1.1 - 2020-06-10
+
+### Added
```
EricJizbaMSFT, line 5: What about "Open Repo in GitHub"?

Feels like that plus the functions thing is two good features to me. Aka worth 0.2.0

EricJizbaMSFT: why you gotta be so negative 😂

PR #130, opened 2020-06-24
Add git clone 👨🏼‍🤝‍👨🏼 to static web app

### LINKED ISSUE #33: Context menu action to `git clone`
I think having a context menu action to `git clone` the repo that is connected to that web app and then open that project in VS Code.

I figure the flow could be similar to deploying, where you select a folder, but that's where we'd `git clone` the project to.

nturinski:
Fixes #33 

There were enough GitHub specific actions that I think it's worth throwing them in their own folder.  None of those files were edited besides their imports.

I'm reconsidering where we have this on the context menu...  Today it looks like this

image

image

I made the choice to have Open Repo in Github exist on the Environment (rather than the Static Web App) because I can open it to the _specific_ branch that they have configured to the app.

Clone and Actions don't have that level of granularity, the target is the repo itself.  But I feel like there's a weird split of where the context action menu items are.  Thoughts?

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
         "onCommand:staticWebApps.createHttpFunction",
         "onCommand:staticWebApps.browse",
         "onCommand:staticWebApps.showActions",
+        "onCommand:staticWebApps.cloneRepo",
         "onCommand:staticWebApps.appSettings.add",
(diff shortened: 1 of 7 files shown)


nturinski:
Okay, will leave it for now then.

Issue #101, opened 2020-05-19
Create new Static Webapp not setting local fs changes


orta:
Hey folks! First off, very cool! I'm looking forwards to moving the TypeScript website to static web apps.

I ran through the "Create new Static Webapp" with all the defaults, which created this repo:  https://github.com/orta/devdiv-onboarding/commit/d497d75bb5f17336116ef12bc5d6039d2e2dec50 and also my Azure stuff

![Screen Shot 2020-05-19 at 2 03 46 PM](https://user-images.githubusercontent.com/49038/82361940-9087e700-99d9-11ea-9ca3-aa58f4087c6d.png)

What it didn't do was set up any folders locally, and left me with this screen:

![Screen Shot 2020-05-19 at 2 01 11 PM](https://user-images.githubusercontent.com/49038/82361710-3850e500-99d9-11ea-81bf-f7fa8cefcc7c.png)

It looks fine in the logs:

```sh
1:53:53 PM: Creating resource group "DevDivOnboarding" in location "centralus"...
1:53:55 PM: Successfully created resource group "DevDivOnboarding".
1:53:55 PM: Creating new GitHub repository "devdiv-onboarding"
1:53:57 PM: Created new GitHub repository "devdiv-onboarding"
1:54:02 PM: Created static web app "DevDivOnboarding"
```

---

I guess my expectations were that there would be _something_ after creating. I assume that this did all the GH/Azure bits and then it is on me to clone the repo and do some work, which is fine, but I didn't feel prompted to do so next

EricJizbaMSFT:
@orta can you walk me through your expected steps? You mentioned your "TypeScript website" - does that mean you already tested your app locally before creating a static web app in Azure? And was it a git repo already (just not in GitHub)? I'm trying to figure out how users expect to make the transition from a local app to an app deployed to the cloud.

orta:
That's an aside. 

I was expecting that after running through the template it would clone the repo I had made and open it.

nturinski:
This PR somewhat address this concern https://github.com/microsoft/vscode-azurestaticwebapps/pull/264

It'll tell the user to push to the GitHub repository to trigger another deployment.  There's also a context menu action on the SWA tree item to clone the repo down.  Hopefully this will help guide users on the next steps.

nturinski:
[...]

PR #133, opened 2020-06-26
Add clone 👨‍🤝‍👨 repo button 🔘 after create 

nturinski:
Fixes https://github.com/microsoft/vscode-azurestaticwebapps/issues/101

I'm not sure how commonly we use telemetry as part of a conditional statement, but it made sense to me in this case.  If we don't want to do that, I can always mess around with the contexts settings that are passed around.

I unshifed clone because I feel like it is the more important post-action compared to Show Actions in this case.

### DIFF
--- a/src/commands/createStaticWebApp/createStaticWebApp.ts
+++ b/src/commands/createStaticWebApp/createStaticWebApp.ts
@@ -9,6 +9,7 @@ import { ext } from '../../extensionVariables';
 import { StaticWebAppTreeItem } from '../../tree/StaticWebAppTreeItem';
 import { SubscriptionTreeItem } from '../../tree/SubscriptionTreeItem';
 import { localize } from '../../utils/localize';
+import { cloneRepo } from '../github/cloneRepo';
 import { showActions } from '../github/showActions';
 
 export async function createStaticWebApp(context: IActionContext, node?: SubscriptionTreeItem): Promise<void> {
@@ -22,10 +23,20 @@ export async function createStaticWebApp(context: IActionContext, node?: Subscri
     ext.outputChannel.appendLog(createdSs);
 
     const showActionsMsg: MessageItem = { title: 'Show Actions' };
+    const cloneRepoMsg: MessageItem = { title: localize('cloneRepo', 'Clone Repo') };
+    const msgItems: MessageItem[] = [showActionsMsg];
+
+    if (context.telemetry.properties.gotRemote === 'false') {
+        // only ask to clone if we didn't detect the remote for creation
+        msgItems.unshift(cloneRepoMsg);
+    }
+
     // don't wait
-    window.showInformationMessage(createdSs, showActionsMsg).then(async (result) => {
+    window.showInformationMessage(createdSs, ...msgItems).then(async (result) => {
         if (result === showActionsMsg) {
             await showActions(context, ssNode);
+        } else if (result === cloneRepoMsg) {
+            await cloneRepo(context, ssNode);
         }
     });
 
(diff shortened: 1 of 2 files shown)


src/commands/createStaticWebApp/createStaticWebApp.ts
```diff
@@ -22,11 +23,24 @@ export async function createStaticWebApp(context: IActionContext, node?: Subscri
     ext.outputChannel.appendLog(createdSs);
 
     const showActionsMsg: MessageItem = { title: 'Show Actions' };
+    const cloneRepoMsg: MessageItem = { title: localize('cloneRepo', 'Clone Repo') };
+    const msgItems: MessageItem[] = [showActionsMsg];
+
+    if (context.telemetry.properties.gotRemote === 'false') {
```
EricJizbaMSFT, line 29: > I'm not sure how commonly we use telemetry as part of a conditional statement, but it made sense to me in this case. If we don't want to do that, I can always mess around with the contexts settings that are passed around.

Don't think we ever do it. I don't like it - doesn't have any type safety

src/commands/createStaticWebApp/createStaticWebApp.ts
```diff
@@ -22,11 +23,24 @@ export async function createStaticWebApp(context: IActionContext, node?: Subscri
...
+        msgItems.unshift(cloneRepoMsg);
+    }
+
     // don't wait
-    window.showInformationMessage(createdSs, showActionsMsg).then(async (result) => {
-        if (result === showActionsMsg) {
-            await showActions(context, ssNode);
-        }
+    window.showInformationMessage(createdSs, ...msgItems).then(async (result) => {
+        await callWithTelemetryAndErrorHandling('postCreateStaticWebApp', async (context2: IActionContext) => {
```
EricJizbaMSFT, line 36: I don't quite understand the scenario for "Clone Repo" here. This means users have never tried their code locally before deploying to Azure? That just seems uncommon to me. Even if I created a new GitHub repo in this flow, it seems more likely I need to do a `git remote set-url`, not a `git clone`

nturinski: Well, if you look at the issue this fixes, that user was expecting to have the project in his local workspace.  He made a new repo through our extension, thus he needed to clone it.

That being said, there have been many cases where I linked a repo that I had hosted on GitHub, but not on my local machine.  This is also a way to guide users to the next step, which is having the project opened in their local workspace and commit changes.

EricJizbaMSFT: Asked the user for more info in the issue. Are there any hints here from our telemetry? For example: how often people use "Clone" from the right click menu, the percent of people that create a new repo, the percent that already had a remote

> That being said, there have been many cases where I linked a repo that I had hosted on GitHub, but not on my local machine. 

Well is that because you were forking a bunch of sample apps for testing instead of creating your own?

> This is also a way to guide users to the next step, which is having the project opened in their local workspace and commit changes.

The debate is what's the "next step". I would argue many users already have a project open before they try to deploy. So the next step would either be just commit/push

nturinski: > Are there any hints here from our telemetry? For example: how often people use "Clone" from the right click menu, the percent of people that create a new repo, the percent that already had a remote

Well, clone is a new feature that wouldn't have telemetry yet because it's gonna be included in this next release.  We can check how often the remote is detected though.

> The debate is what's the "next step". I would argue many users already have a project open before they try to deploy. So the next step would either be just commit/push

Well, the clone button only appears _if_ we didn't detect a remote for the user.  So this could be a non-issue.  But I do think if we can detect that they don't have their repo opened, then we should give them the option to bring it down.

That being said, I think a logical feature to add to this is to have the ability of creating a new repo _with_ the local project that they have opened.

nturinski:
Closing until we have a better understanding of customer scenarios

Issue #36, opened 2020-05-08
Add VS Code settings to remember api/app/build folders


EricJizbaMSFT:
I see this helping in two scenarios:
1. I create multiple static web apps from the same project. We should only need to prompt the first time since the project structure probably doesn't change
1. I clone a static web app project from somewhere else. If the VS Code settings are in the repo already, we wouldn't need to prompt. This could be especially helpful for sample projects to avoid user error around selecting the wrong folder

That being said, I'm not sure I want this until we support "Advanced" create so that it's obvious how to bypass the settings if I want to create a static web app for something other than the currently opened workspace.

Related to https://github.com/microsoft/vscode-azurestaticwebapps/issues/19

PR #137, opened 2020-07-02
Save defaults after creation of first SWA app

nturinski:
Fixes https://github.com/microsoft/vscode-azurestaticwebapps/issues/36

Right now, the setting is just being used to set the default value when the prompts come up because I don't want to block users in any way.  But I figured when we have the concept of a basic create, we can auto-populate the value with these settings.

I figured that I should update the setting whenever the user creates the SWA.  Even if they use the default values, I think it makes sense to save it for them to expose the settings.

I think the `apiSubpath` setting should also be set in the case when we create an "api" folder for the user, but that'll be a separate PR.

### DIFF
--- a/src/tree/SubscriptionTreeItem.ts
+++ b/src/tree/SubscriptionTreeItem.ts
@@ -13,10 +13,13 @@ import { GitHubRepoListStep } from '../commands/createStaticWebApp/GitHubRepoLis
 import { IStaticWebAppWizardContext } from '../commands/createStaticWebApp/IStaticWebAppWizardContext';
 import { StaticWebAppCreateStep } from '../commands/createStaticWebApp/StaticWebAppCreateStep';
 import { StaticWebAppNameStep } from '../commands/createStaticWebApp/StaticWebAppNameStep';
+import { apiSubpathSetting, appArtifactSubpathSetting, appSubpathSetting } from '../constants';
 import { getGitHubAccessToken, tryGetRemote } from '../utils/gitHubUtils';
 import { localize } from '../utils/localize';
 import { nonNullProp } from '../utils/nonNull';
 import { requestUtils } from '../utils/requestUtils';
+import { updateWorkspaceSetting } from '../utils/settingsUtils';
+import { getSingleRootFsPath } from '../utils/workspaceUtils';
 import { StaticWebApp, StaticWebAppTreeItem } from './StaticWebAppTreeItem';
 
 export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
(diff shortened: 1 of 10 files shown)


src/tree/SubscriptionTreeItem.ts
```diff
@@ -72,13 +75,21 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
...
+
         await wizard.prompt();
         const newStaticWebAppName: string = nonNullProp(wizardContext, 'newStaticWebAppName');
         wizardContext.newResourceGroupName = newStaticWebAppName;
 
         await wizard.execute();
         context.showCreatingTreeItem(newStaticWebAppName);
 
+        if (wizardContext.fsPath) {
+            await updateWorkspaceSetting(appSubpathSetting, wizardContext.appLocation, wizardContext.fsPath);
```
EricJizbaMSFT, line 88: I'm a little unsure about the experience here. Seems like we can't be sure the current workspace matches up with the repo they're going to use for the static web app. Also would be annoying if I already had something set and you always overwrote the value. A few ideas:

1. Only update settings if you 'gotRemote'
1. Only update settings if I don't already have something set
1. Never update settings automatically and prompt to save defaults

src/tree/SubscriptionTreeItem.ts
```diff
@@ -72,13 +75,21 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
...
+            await updateWorkspaceSetting(apiSubpathSetting, wizardContext.apiLocation, wizardContext.fsPath);
+            await updateWorkspaceSetting(appArtifactSubpathSetting, wizardContext.appArtifactLocation, wizardContext.fsPath);
+        }
```
EricJizbaMSFT, line 91: > But I figured when we have the concept of a basic create, we can auto-populate the value with these settings.

You know you can add basic/advanced create at any time, right? They don't have to be completely different experiences to start with - you can slowly build up the defaults for 'basic' and the extra prompts for 'advanced'. To start with, the only difference might be these three settings and the auto-detected remote logic

src/tree/SubscriptionTreeItem.ts
nturinski: > Seems like we can't be sure the current workspace matches up with the repo they're going to use for the static web app
Yeah, true that.  I think idea 1. would help mitigate that at least.

>Also would be annoying if I already had something set and you always overwrote the value.
I thought about that, but I figured if someone went out of their way to change the value that was previously their default, then it probably wasn't the default that they needed.

EricJizbaMSFT: Sounds good

PR #144, opened 2020-07-10
Don't hard-code the API location

### LINKED ISSUE #107: Remember Backend Location For API Creation
When I created my function app, I selected which folder I want my front end and which I want my back end, the default is:
```
  - app
  - api
```
however, I personally prefer:

```
  - frontend
  - backend
```

and selected those folders during the static web app creation process, and it sets up the github actions based on those folders. However, when I go to later add an api to it, it creates the function in the `api` folder, not the specified `backend` folder.

alexweininger:
When creating a Function, use the workspace setting, and fallback on the default `/api` location.

Fixes #107

### DIFF
--- a/src/commands/createHttpFunction.ts
+++ b/src/commands/createHttpFunction.ts
@@ -8,9 +8,10 @@ import * as path from 'path';
 import * as vscode from 'vscode';
 import { IActionContext, UserCancelledError } from "vscode-azureextensionui";
 import { AzureExtensionApiProvider } from 'vscode-azureextensionui/api';
-import { defaultApiName } from '../constants';
+import { apiSubpathSetting, defaultApiName } from '../constants';
 import { ext } from '../extensionVariables';
 import { localize } from '../utils/localize';
+import { getWorkspaceSetting } from '../utils/settingsUtils';
 import { AzureFunctionsExtensionApi } from '../vscode-azurefunctions.api';
 
 export async function createHttpFunction(context: IActionContext): Promise<void> {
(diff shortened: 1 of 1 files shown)


nturinski:
I think we also want to prompt the user for a folder name (and have the default value be api or the workspaceSetting)

src/commands/createHttpFunction.ts
```diff
@@ -22,7 +23,14 @@ export async function createHttpFunction(context: IActionContext): Promise<void>
     const funcApi: AzureFunctionsExtensionApi = await getFunctionsApi(context);
 
     const endpointName: string = 'endpoint';
-    const folderPath: string = path.join(vscode.workspace.workspaceFolders[0].uri.fsPath, defaultApiName);
+    const apiLocationPlaceholder: string = getWorkspaceSetting(apiSubpathSetting, vscode.workspace.workspaceFolders[0].uri.fsPath) || defaultApiName;
```
EricJizbaMSFT, line 26: > I think we also want to prompt the user for a folder name (and have the default value be api or the workspaceSetting)

The purpose of the setting is to avoid a prompt, not to be the default value for a prompt. Also - I never prompt for a folder before creating a function in the functions extension. I think this experience needs to be ironed out a bit more so that it's not prompting too much

nturinski: Okay then let's roll back to Alex's last commit to where it just replaces 'api' with the workspace setting.  That should at least unblock some users

alexweininger: Rolled back ✔️

Issue #136, opened 2020-07-01
Improve configuration discovery


fiveisprime:
Scenarios

* Failed GH Actions build, point customer to the YAML file
* Tree action to open configuration file (purely for discoverability)

- [x] When creating the app, include the GitHub repo directories in the prompt as selection (e.g. App location, API location)
- [x] Show GitHub Action status in explorer
- [x] Show GitHub Action build status on environment initial creation
- [x] Link to YAML config file with the relevant lines selected (from action or repo?)

PR #146, opened 2020-07-13
Add actions node that lists runs

nturinski:
_CHANGED ARE SMALLER THAN THEY APPEAR:
A lot of the files changed are just the repoUrl (I noticed that I used `node.data.properties.repositoryUrl` quite a bit) and SVG files._

Part 1/4 of this issue https://github.com/microsoft/vscode-azurestaticwebapps/issues/136
Regarding the rerunning/cancelling of jobs, I had two ideas but ultimately decided not to implement it:

1. I thought about having a timeout for the refreshing of the treeItem while we wait for the action to complete.  But I don't think there's any reasonable guess as to how long a rerun will take since that's entirely dependent on their project.

1. I thought about using a cancel token to stifle the "rerun has completed message" but I think relying on GitHub's API response seems reliable enough than to have our own VS Code token tracking all of the different ID's that have been rerun.


nturinski:
The ideas are the two listed below.  Just a timeout and using a cancellation token if the user reruns the action, and then cancels the rerun action.  Previously it would say that the rerun action and the cancel action both completed at the same time.

src/tree/ActionTreeItem.ts
```diff
@@ -0,0 +1,100 @@
...
+        ext.outputChannel.appendLog(cancelRunning);
+
+        await this.waitForRunToFinish(gitHubRequest);
+        const cancelCompleted: string = localize('cancelCompleted', 'Cancel for action "{0}" has completed.', this.data.id);
+        ext.outputChannel.appendLog(cancelCompleted);
+    }
+
+    private async waitForRunToFinish(gitHubRequest: gitHubWebResource): Promise<void> {
+        await requestUtils.sendRequest(gitHubRequest);
+        await this.refresh(); // need to refresh to update the data
```
EricJizbaMSFT, line 93: > I thought about having a timeout for the refreshing of the treeItem while we wait for the action to complete. But I don't think there's any reasonable guess as to how long a rerun will take since that's entirely dependent on their project.

I'm not sure what the answer is, but I think we need a better solution here. I almost feel like you should pull out the auto-refresh logic, so that we can merge the rest of the code and have more time to figure out a solution to this separately. A few reasons I don't like this:
1. Every 2 seconds seems too often to me, since an action could take so long to complete (i.e. our Cosmos nightly build can easily take an hour)
1. The class is re-created if the user refreshes the whole tree, meaning you would continue to refresh in an instance that's no longer even displayed
1. You refresh the tree (causing at least a blip in the UI), even if nothing changes

resources/dark/conclusions/skip.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="14" height="14" viewBox="0 0 14 14" fill="none" xmlns="http://www.w3.org/2000/svg">
```
alexweininger, line 1: Rename this file to "skipped.svg" so that it matches the conclusion response property.

See [here](https://developer.github.com/v3/checks/runs/#parameters)

src/tree/ActionTreeItem.ts
```diff
@@ -0,0 +1,100 @@
...
+
+        await this.waitForRunToFinish(gitHubRequest);
+        if (this.data.conclusion !== 'cancelled') {
+            const rerunCompleted: string = localize('rerunCompleted', 'Rerun for action "{0}" has completed.', this.data.id);
+            ext.outputChannel.appendLog(rerunCompleted);
+        }
+    }
+
+    public async cancelAction(context: IGitHubContext): Promise<void> {
+        const gitHubRequest: gitHubWebResource = await createGitHubRequestOptions(context, this.data.cancel_url, 'POST');
```
EricJizbaMSFT, line 82: I prefer to put command stuff in a file in the 'commands' folder, instead of on the tree item itself. Otherwise it can be easy to accidentally make the tree item classes pretty bloated. That being said, I might put `rerunAction`, `cancelAction`, and `waitForRunToFinish` in the same file since they're related

src/utils/gitHubUtils.ts
```diff
@@ -104,12 +104,14 @@ export async function getGitHubQuickPicksWithLoadMore<T>(cache: ICachedQuickPick
...
 }
 
-export async function createGitHubRequestOptions(context: IStaticWebAppWizardContext, url: string, method: HttpMethods = 'GET'): Promise<gitHubWebResource> {
-    if (!context.accessToken) {
-        context.accessToken = await getGitHubAccessToken();
+export async function createGitHubRequestOptions(context: IGitHubContext | undefined, url: string, method: HttpMethods = 'GET'): Promise<gitHubWebResource> {
+    const accessToken: string = context?.accessToken || await getGitHubAccessToken();
+
+    if (context && !context.accessToken) {
+        context.accessToken = accessToken;
```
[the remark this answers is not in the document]
nturinski: The context is optional and is used when creating StaticWebApps.  I just set it so that way I don't have to keep retrieving it from the VS Code API, but honestly that's probably a safer and quicker way to do it.

resources/dark/conclusions/skip.svg
nturinski: Nice catch!

src/tree/ActionTreeItem.ts
nturinski: Okay, alternatively, I could just keep polling in the background (sort of like how like what's done [here](https://github.com/microsoft/vscode-azurestaticwebapps/blob/master/src/tree/StaticWebAppTreeItem.ts#L126-L154)) and update it once the conclusion populates.

While I think 2 seconds is kind of frequent, I wanted it to feel pretty responsive.  Maybe we could do a incremental increase in polling time?

src/tree/ActionTreeItem.ts
```diff
@@ -0,0 +1,100 @@
...
+        super(parent);
+        this.data = data;
+    }
+
+    public get iconPath(): TreeItemIconPath {
+        return this.data.conclusion ? treeUtils.getThemedIconPath(path.join('conclusions', this.data.conclusion)) : treeUtils.getThemedIconPath(path.join('statuses', this.data.status));
+    }
+
+    public get id(): string {
+        return `${this.parent.parent.id}/${this.data.id}`;
```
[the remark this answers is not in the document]
nturinski: It's not needed, but just lets us know from the id which treeItem it's under. I think all the ids should be unique as is.

src/utils/gitHubUtils.ts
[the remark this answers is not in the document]
nturinski: Ah right, the one annoying thing about this though is that every time we make a GitHub request, we have to get it from VS Code.  That's fine if you set it to never ask me for permissions again, but if a user didn't enable that, it'd be super annoying (and we shouldn't have to ask the user for a token 3 times in one static web app create)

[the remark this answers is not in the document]
EricJizbaMSFT: I'm not opposed to setting accessToken on context - I'm opposed to _where_ you're doing it. I don't like functions that have unrelated side-effects. You should be able to add accessToken to context when you're setting up the wizard (pretty standard procedure to add a few default values to a wizardContext)

Btw, there is a bit of risk the accessToken expires during the course of a single command, but I think that's outweighed by the extra prompts you mentioned.

src/tree/ActionTreeItem.ts
EricJizbaMSFT: Can you clarify what you would do differently based on that example? The biggest difference I see is that function stops polling after a minute

src/tree/ActionTreeItem.ts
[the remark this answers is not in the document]
EricJizbaMSFT: It's just weird that it's the grandparent instead of the parent. That being said, `fullId` already includes the parent id, so there's no value in adding parent id to `id`

src/tree/ActionTreeItem.ts
nturinski: The biggest difference is that the polling is happening independently of the tree.  Right now I'm leveraging the refresh command which is what makes it look super sloppy.

This logic is similar to the `redeploy` where we await `waitForDeploymentToComplete` (which is polling the deployment logs) and then update the tree when it's finished.

EricJizbaMSFT: Sounds like that could help, but I'd have to wait and see what you actually do for this case

src/utils/gitHubUtils.ts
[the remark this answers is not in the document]
nturinski: Actually, it seems like VS Code may have gotten rid of the prompts for GitHub access (unless I can't figure how to turn off the 'Don't Show me Again" setting).  Still, I imagine VS Code's Token API is making a call to GitHub which could potentially increase our latency between steps which seems annoying.

src/tree/ActionTreeItem.ts
nturinski: Should I just make it in another PR?

nturinski:
image

Just showing the timestamps of the actions.  The screenshot doesn't reflect what's actually in the description.

src/tree/ActionTreeItem.ts
```diff
@@ -45,7 +41,7 @@ export class ActionTreeItem extends AzureTreeItem {
     }
 
     public get id(): string {
-        return `${this.parent.parent.id}/${this.data.id}`;
+        return `${this.data.id}`;
```
EricJizbaMSFT, line 44: nit: just use `this.data.id` instead of `` `${this.data.id}` ``

src/tree/ActionsTreeItem.ts
```diff
@@ -0,0 +1,66 @@
...
+            actions.workflow_runs,
+            'invalidActionTreeItem',
+            (act) => new ActionTreeItem(this, act),
+            act => act.head_commit.message
+        );
+    }
+
+    public compareChildrenImpl(_ti1: ActionTreeItem, _ti2: ActionTreeItem): number {
+        // the GitHub API returns the actions in ascending creation order so we can just return 1 to maintain that order
+        return 1;
```
EricJizbaMSFT, line 60: Pretty sure it's better to use "0" to maintain order

> If the result is negative a is sorted before b.
> 
> If the result is positive b is sorted before a.
> 
> If the result is 0 no changes are done with the sort order of the two values.

https://www.w3schools.com/js/js_array_sort.asp

src/utils/gitHubUtils.ts
```diff
@@ -104,25 +104,22 @@ export async function getGitHubQuickPicksWithLoadMore<T>(cache: ICachedQuickPick
...
 
-export async function createGitHubRequestOptions(context: IStaticWebAppWizardContext, url: string, method: HttpMethods = 'GET'): Promise<gitHubWebResource> {
-    if (!context.accessToken) {
-        context.accessToken = await getGitHubAccessToken();
-    }
-
-    const requestOptions: gitHubWebResource = await requestUtils.getDefaultRequest(url, new TokenCredentials(context.accessToken), method);
+// a context should be passed in if many subsequential GitHub requests are made (i.e. the create SWA wizard)
+export async function createGitHubRequestOptions(context: IStaticWebAppWizardContext | undefined, url: string, method: HttpMethods = 'GET'): Promise<gitHubWebResource> {
+    const requestOptions: gitHubWebResource = await requestUtils.getDefaultRequest(url, new TokenCredentials(await getGitHubAccessToken(context)), method);
```
EricJizbaMSFT, line 109: I think I would prefer if you made `accessToken` a parameter. Thus it would look like one of these two options when you call it:

```
let requestOptions: gitHubWebResource = await createGitHubRequestOptions(context.accessToken, `${githubApiEndpoint}/user`);
```

```
const token: string = await getGitHubAccessToken();
const gitHubRequest: gitHubWebResource = await createGitHubRequestOptions(token, node.data.cancel_url, 'POST');
```

A few reasons I don't like the current stuff:
- It doesn't feel right to have "gitHubUtils" know about "IStaticWebAppWizardContext" (I think you agree, otherwise you wouldn't have felt the need to add a comment)
- I try to avoid cases where I have to explicitly pass `undefined` as a parameter
- It bothers me that `getGitHubAccessToken ` doesn't always get an access token

nturinski: Can't believe you're using my own comment against me...

nturinski: I do feel better about this though 😁 Thanks. I'm not sure why I didn't think of this as a solution...

PR #149, opened 2020-07-17
Detect if current branch is a SWA

nturinski:
image

Waiting on this PR https://github.com/microsoft/vscode-azurestaticwebapps/pull/146 to be merged so context doesn't have to passed to `tryGetRemote()`.

I couldn't think of a great way to convey that your local workspace is a project linked to a SWA in your subscription.  I thought of the words active or linked, but... yeah.

Also feels like it wouldn't be easy to notice either since it's just description is already so subtle.

I was thinking about whether or not we should support multi-root.  I know usage is kind of low so I'm not sure if the complexity is worth the trade off.  Thoughts?


src/tree/EnvironmentTreeItem.ts
```diff
@@ -58,8 +61,9 @@ export class EnvironmentTreeItem extends AzureParentTreeItem implements IAzureRe
         return this.data.properties.buildId === 'default' ? productionEnvironmentName : `#${this.name} - ${this.data.properties.pullRequestTitle}`;
     }
 
-    public get description(): string | undefined {
-        return this.data.properties.sourceBranch;
+    public get description(): string {
+        const linkedTag: string = this.inWorkspace ? '(linked)' : '';
+        return `${this.data.properties.sourceBranch} ${linkedTag}`;
```
EricJizbaMSFT, line 66: Use localize here with "linked"

src/utils/gitHubUtils.ts
```diff
@@ -148,6 +150,22 @@ export async function tryGetRemote(): Promise<string | undefined> {
     return;
 }
 
+export async function tryGetBranch(): Promise<string | undefined> {
```
EricJizbaMSFT, line 153: For "try" functions, I prefer if the "try/catch" wraps like the entire function. For example, maybe `git(localProjectPath);` could throw an error 🤷‍♂️

src/tree/StaticWebAppTreeItem.ts
```diff
@@ -84,11 +84,16 @@ export class StaticWebAppTreeItem extends AzureParentTreeItem implements IAzureR
     public async loadMoreChildrenImpl(_clearCache: boolean, _context: IActionContext): Promise<AzExtTreeItem[]> {
         const requestOptions: requestUtils.Request = await requestUtils.getDefaultAzureRequest(`${this.id}/builds?api-version=2019-12-01-preview`, this.root);
         const envs: StaticEnvironment[] = (<{ value: StaticEnvironment[] }>JSON.parse(await requestUtils.sendRequest(requestOptions))).value;
+        const remote: string | undefined = await tryGetRemote();
+        const branch: string | undefined = remote ? await tryGetBranch() : undefined;
```
EricJizbaMSFT, line 88: How often does `tryGetRemote` get called related to this PR? Like - if I have a bunch of static apps or environments - does it get called a lot? If so, we might want to find a more efficient way of caching this information since it's tied to the workspace and is unlikely to change

src/tree/EnvironmentTreeItem.ts
```diff
@@ -96,4 +100,10 @@ export class EnvironmentTreeItem extends AzureParentTreeItem implements IAzureRe
 
         return undefined;
     }
+
+    public async refreshImpl(): Promise<void> {
+        const remote: string | undefined = await tryGetRemote();
+        const branch: string | undefined = remote ? await tryGetBranch() : undefined;
+        this.inWorkspace = this.parent.data.properties.repositoryUrl === remote && this.data.properties.sourceBranch === branch;
```
EricJizbaMSFT, line 107: You're doing this here and in `loadMoreChildrenImpl`. Can we share that code somehow?

```
this.parent.data.properties.repositoryUrl === remote && this.data.properties.sourceBranch === branch;
```

src/utils/gitHubUtils.ts
nturinski: 🧤

src/tree/EnvironmentTreeItem.ts
nturinski: I don't know if we even need this tbh.  I did it in case we ended up supporting multi-root, but you wouldn't be able to open a new folder without reactivating the extension so it'd have to regenerate the entire tree anyway.

src/tree/StaticWebAppTreeItem.ts
nturinski: Right now it gets called:
- When an `EnvironmentTreeItem` is refreshed
- When a `StaticWebAppTreeItem` is expanded (only called once per expanding, not per environment)
- When creating a Static Web App (during the wizard)

I see the concern, but I don't think it'll really impact users too much.  It's a pretty fast GET request, the bottleneck would probably be making the request to Azure for the children.  

I'm also thinking we can remove the check in `EnvironmentTreeItem` (though we should probably support multi-root at some point)

src/tree/EnvironmentTreeItem.ts
nturinski: Never mind, I remember why I thought this was necessary now.  Users can easily change the branch of their project so we at least need a refresh so we can reflect that.

src/commands/isWorkspaceInAzure.ts
```diff
@@ -0,0 +1,14 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { StaticEnvironment } from "../tree/EnvironmentTreeItem";
+import { StaticWebApp } from "../tree/StaticWebAppTreeItem";
+import { tryGetBranch, tryGetRemote } from "../utils/gitHubUtils";
+
+export async function isWorkspaceInAzure(swa: StaticWebApp, env: StaticEnvironment): Promise<boolean> {
```
EricJizbaMSFT, line 10: I meant more like this in terms of sharing code: 
https://github.com/microsoft/vscode-azurefunctions/blob/efe121205711eafe30adfcd88e494a85745e15c0/src/tree/remoteProject/RemoteFunctionsTreeItem.ts#L24

Feels a bit weird to have this random function in "src/commands". If you don't want to do the above, at least move it to the "tree" folder, because that's where it's used.

src/tree/EnvironmentTreeItem.ts
```diff
@@ -58,8 +62,9 @@ export class EnvironmentTreeItem extends AzureParentTreeItem implements IAzureRe
         return this.data.properties.buildId === 'default' ? productionEnvironmentName : `#${this.name} - ${this.data.properties.pullRequestTitle}`;
     }
 
-    public get description(): string | undefined {
-        return this.data.properties.sourceBranch;
+    public get description(): string {
+        const linkedTag: string = this.inWorkspace ? localize('linkedTag', '(linked)') : '';
```
EricJizbaMSFT, line 66: Not quite - you need to put the localized content in context whenever possible. Depending on the language, the ordering of the words might change.

localize('linkedTag', '{0} (linked)', branch)

src/commands/isWorkspaceInAzure.ts
nturinski: Ah, I see what you mean.  Yeah, I can do something like that instead.

PR #152, opened 2020-07-20
Show workflow jobs and steps

alexweininger:
Not sure how Nathan's commits got on this PR, but if that's an issue I will try to fix it.

image

A lot of added code here is added to deal with displaying the times and date completed of the jobs and steps. We might try to rely on an npm package to do this for us, however I didn't immediately find one that suited our needs.

I also feel that I need to write tests for the util functions I added. I will write these soon I just wanted to get eyes on the code asap.


src/utils/timeUtils.ts
```diff
@@ -0,0 +1,68 @@
...
+*  Licensed under the MIT License. See License.txt in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import { localize } from './localize';
+
+/**
+ * Get the elapsed time since `date` as a pretty printed string.
+ * @param date Date of event.
+ */
+export function timeDifferenceString(date: Date): string {
```
EricJizbaMSFT, line 12: > A lot of added code here is added to deal with displaying the times and date completed of the jobs and steps. We might try to rely on an npm package to do this for us, however I didn't immediately find one that suited our needs.

We use [moment](https://www.npmjs.com/package/moment) for this kind of thing. Specifically this: https://momentjs.com/docs/#/displaying/fromnow/

alexweininger: Thanks! Now I can delete my code :)

src/tree/ActionTreeItem.ts
```diff
@@ -56,6 +57,21 @@ export class ActionTreeItem extends AzureTreeItem implements IAzureResourceTreeI
         return this.data.event;
     }
 
+    public async loadMoreChildrenImpl(_clearCache: boolean, _context: IActionContext): Promise<AzExtTreeItem[]> {
+        const { owner, name } = getRepoFullname(this.parent.repositoryUrl);
+        const token: string = await getGitHubAccessToken();
+        const requestOption: gitHubWebResource = await createGitHubRequestOptions(token, `${githubApiEndpoint}/repos/${owner}/${name}/actions/runs/${this.data.id}/jobs`);
+        const githubResponse: IncomingMessage & { body: string } = await requestUtils.sendRequest(requestOption);
+        const gitHubJobs: { jobs: GitHubJob[] } = <{ jobs: GitHubJob[] }>JSON.parse(githubResponse.body);
+        return gitHubJobs.jobs.map((job => {
```
nturinski, line 66: Like Eric mentioned before, we should use the `createTreeItemsWithErrorHandling` especially since it's a custom call that we're making.

src/constants.ts
```diff
@@ -11,3 +11,8 @@ export const productionEnvironmentName: string = 'Production';
 export const appSubpathSetting: string = 'appSubpath';
 export const apiSubpathSetting: string = 'apiSubpath';
 export const appArtifactSubpathSetting: string = 'appArtifactSubpath';
+
+// Doc for these parameter values: https://developer.github.com/v3/checks/runs/#parameters
+export type Conclusion = 'success' | 'failure' | 'skipped' | 'cancelled' | null;
```
nturinski, line 16: @EricJizbaMSFT I think we should make this an enum.  See why in a comment below!

src/tree/JobTreeItem.ts
```diff
@@ -0,0 +1,101 @@
...
+        super(parent);
+        this.data = data;
+    }
+
+    public get iconPath(): TreeItemIconPath {
+        return treeUtils.getActionIconPath(this.data.status, this.data.conclusion);
+    }
+
+    public get id(): string {
+        return `${this.parent.id}/${this.data.id}`;
```
nturinski, line 53: I think that you don't need the parent.id.  Each id should be unique and `fullId` will contain the entire parent.

src/utils/gitHubUtils.ts
```diff
@@ -157,3 +158,29 @@ export function isUser(orgData: gitHubOrgData | undefined): boolean {
     // if there's no orgData, just assume that it's a user (but this shouldn't happen)
     return orgData ? orgData.type === 'User' : true;
 }
+
+export function convertConclusionToVerb(conclusion: Conclusion): string {
+    switch (conclusion) {
```
nturinski, line 163: I think this can be cleaned up a bit if you use an enum instead of a type.  It'll look like this: https://github.com/microsoft/vscode-azurefunctions/blob/67c8ebc6b8931e6ab379dfe42c9c747de2c5b03e/src/tree/remoteProject/RemoteFunctionTreeItem.ts#L63-L73

src/utils/timeUtils.ts
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * Get a string in the `#m #s` format. E.g. `2m 34s`.
+ *
+ * @param start Starting Date
+ * @param end Ending Date
+ *
+ * **NOTE:** If the difference between `start` and `end` is negative, `end` will be treated as the current time.
+ */
+// tslint:disable-next-line: export-name
+export function getTimeElapsedString(start: Date, end: Date): string {
```
nturinski, line 15: I think you can probably replace this with this https://momentjs.com/docs/#/displaying/difference/

alexweininger: I can't see any way to cleanly format a moment object to `2m 31s`.

It seems like it's a missing feature that people have wanted for a while and is still not implemented in moment: https://github.com/moment/moment/issues/463

I've looked at both
* https://momentjs.com/docs/#/displaying/format/
* and https://momentjs.com/docs/#/durations/humanize/

And neither can do the job.

EricJizbaMSFT: IMO it feels like you're trying to pack too much info into the description. This piece feels the least important relative to it's code complexity and I would get rid of it.

src/utils/gitHubUtils.ts
EricJizbaMSFT: FWIW, I don't care about "enum" vs. "type". It's six of one and half a dozen of the other

src/constants.ts
```diff
@@ -11,3 +11,17 @@ export const productionEnvironmentName: string = 'Production';
 export const appSubpathSetting: string = 'appSubpath';
 export const apiSubpathSetting: string = 'apiSubpath';
 export const appArtifactSubpathSetting: string = 'appArtifactSubpath';
+
+// Doc for these parameter values: https://developer.github.com/v3/checks/runs/#parameters
+
+export enum Conclusion {
```
EricJizbaMSFT, line 17: These types should stay in the same file as the type for `GitHubAction` since they're so closely related. If you don't like having it in `ActionTreeItem`, make a separate file for GitHub stuff. The separate file would somewhat represent a drop-in replacement for the GitHub npm package that I assume you guys will switch to soon

src/tree/StepTreeItem.ts
```diff
@@ -0,0 +1,69 @@
...
+import * as moment from 'moment';
+import { AzureTreeItem, TreeItemIconPath } from "vscode-azureextensionui";
+import { Conclusion, Status } from '../constants';
+import { convertConclusionToVerb, convertStatusToVerb } from '../utils/gitHubUtils';
+import { getTimeElapsedString } from '../utils/timeUtils';
+import { treeUtils } from "../utils/treeUtils";
+import { IAzureResourceTreeItem } from './IAzureResourceTreeItem';
+import { JobTreeItem } from './JobTreeItem';
+
+export type GitHubStep = {
```
EricJizbaMSFT, line 15: Is there any reason you guys are waiting to use the octokit packages? It bothers me that we're adding more technical debt that we know we're going to switch soon anyways

src/tree/ActionTreeItem.ts
```diff
@@ -4,28 +4,29 @@
...
 import { requestUtils } from '../utils/requestUtils';
 import { treeUtils } from "../utils/treeUtils";
 import { ActionsTreeItem } from "./ActionsTreeItem";
 import { IAzureResourceTreeItem } from './IAzureResourceTreeItem';
+import { GitHubJob, JobTreeItem } from './JobTreeItem';
 
 export type GitHubAction = {
     id: number;
-    conclusion: 'success' | 'failure' | 'skip' | 'cancelled' | null;
+    conclusion: Conclusion;
```
nturinski, line 18: This can be null as well.

src/constants.ts
```diff
@@ -11,3 +11,17 @@ export const productionEnvironmentName: string = 'Production';
...
+    Success = 'success',
+    Failure = 'failure',
+    Skipped = 'skipped',
+    Cancelled = 'cancelled'
+}
+
+export enum Status {
```
nturinski, line 24: I missed it before, but there's also a completed status.

src/tree/StepTreeItem.ts
nturinski: I'm in the process of converting it now.  There's been some hiccups so it's taking a little longer than I thought.  Should be able to get it done today.

EricJizbaMSFT: I would think new code (aka this PR) could use it without needing to convert old stuff

src/constants.ts
nturinski: I think you missed this comment-- we probably could move this to the gitHubTypings.  Annoying that they don't have their own enum for this, but alas

alexweininger: Oops! Will do.

src/tree/ActionsTreeItem.ts
```diff
@@ -44,13 +44,11 @@ export class ActionsTreeItem extends AzureParentTreeItem {
         const { owner, name } = getRepoFullname(this.repositoryUrl);
         const branch: string = this.parent.data.properties.sourceBranch;
 
-        const token: string = await getGitHubAccessToken();
-        const requestOption: gitHubWebResource = await createGitHubRequestOptions(token, `${githubApiEndpoint}/repos/${owner}/${name}/actions/runs?branch=${branch}`);
-        const githubResponse: IncomingMessage & { body: string } = await requestUtils.sendRequest(requestOption);
+        const octokitClient: Octokit = await createOctokitClient();
```
nturinski, line 47: Thanks for converting some of my stuff to consume Octokit for me 😁

src/tree/JobTreeItem.ts
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    public get label(): string {
+        return this.name;
+    }
+
+    public get description(): string {
+        if (this.data.conclusion !== null) {
+            const elapsedMs: number = this.completedDate.getTime() - this.startedDate.getTime();
+            return `${convertConclusionToVerb(<Conclusion>this.data.conclusion)} ${moment(this.completedDate).fromNow()} in ${prettyMs(elapsedMs)}`;
```
nturinski, line 50: So I wanted to confirm with GitHub, but apparently it doesn't support any language other than English: https://github.blog/2010-07-13-github-in-your-language/

I feel like we might want to localize this string because I'm not sure this format is standard in say, Chinese or something.

nturinski, line 50: Rather than casting this.data.conclusion, you could make a get property for conclusion that confirms it's one of the enums and then returns it casted.

https://stackoverflow.com/questions/43804805/check-if-value-exists-in-enum-in-typescript

src/tree/JobTreeItem.ts
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    public hasMoreChildrenImpl(): boolean {
+        return false;
+    }
+
+    public async refreshImpl(): Promise<void> {
+        const { owner, name } = getRepoFullname(this.parent.parent.repositoryUrl);
+        const octokitClient: Octokit = await createOctokitClient();
+        this.data = (await octokitClient.actions.getJobForWorkflowRun({ job_id: this.parent.data.id, owner: owner, repo: name })).data;
```
nturinski, line 80: Is this supposed to be using the parent.data.id rather than it's own?  That seems odd to me.

alexweininger: Good catch

src/tree/StepTreeItem.ts
```diff
@@ -0,0 +1,59 @@
...
+    public get label(): string {
+        return this.name;
+    }
+
+    public get description(): string {
+        if (this.data.conclusion !== null) {
+            const elapsedMs: number = this.completedDate.getTime() - this.startedDate.getTime();
+            return `${convertConclusionToVerb(<Conclusion>this.data.conclusion)} in ${prettyMs(elapsedMs)}`;
+        } else {
+            return `${convertStatusToVerb(<Status>this.data.status)} ${this.startedDate.getTime() === 0 ? '' : moment(this.startedDate).fromNow()}`;
```
nturinski, line 48: What does fromNow() return if the startedDate is 0 (ie now)?  

I feel like instead of having a blank, saying queued now is more similar to what GitHub does anyway.

src/tree/JobTreeItem.ts
alexweininger: So in the getter, if it's not in the enum should I throw an error?

src/tree/StepTreeItem.ts
alexweininger: This check is making sure the `this.startedDate` isn't the epoch (which is what it is if github hasn't given us a started it yet. 

From now when the startedDate when `startedDate.getTime() === 0` returns like 31 years ago or something since that's the epoch.

src/tree/JobTreeItem.ts
nturinski: Yeah, probably something along the lines of "Received unexpected conclusion."  I'm pretty sure it'll mess up the getActionIconPath code anyway if we don't (since the path won't be found if it's not an enum we expect.)

src/tree/StepTreeItem.ts
nturinski: Ohh, like the beginning of the ms clock?  I didn't realize that's what it was called.  Okay, gotcha.

src/tree/JobTreeItem.ts
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    public get label(): string {
+        return this.name;
+    }
+
+    public get description(): string {
+        if (this.data.conclusion !== null) {
+            const elapsedMs: number = this.completedDate.getTime() - this.startedDate.getTime();
+            return `${convertConclusionToVerb(this.conclusion)} ${moment(this.completedDate).fromNow()} in ${prettyMs(elapsedMs)}`;
```
EricJizbaMSFT, line 51: > Actually, moment is writing complete phrases like 21 days ago or 21 minutes ago, so I'm not sure how we would localize that unless moment has language support.

Moment supports localization. However, I don't think "pretty-ms" does. Even if it did - my original comment was that you're putting too much stuff in the description and I stand by that. The biggest reason I love VS Code is because of how lightweight/simple it is - and I want our extensions to reflect that. Pick only the most important information to display and throw the rest in a "View Properties" command

src/tree/JobTreeItem.ts
```diff
@@ -0,0 +1,103 @@
...
+
+    private get completedDate(): Date {
+        return new Date(this.data.completed_at);
+    }
+
+    private get status(): Status {
+        if (this.data.status in Status) {
+            return <Status>this.data.conclusion;
+        } else {
+            throw new Error(localize('statusNotRecognized', 'Status not recognized.'));
```
EricJizbaMSFT, line 69: Add `this.data.status` to the error, so that we have a chance of fixing the error based on telemetry alone. This is what I usually do:
```
throw new RangeError(localize('invalidStatus', 'Invalid status "{0}".', this.data.status));
```

src/tree/StepTreeItem.ts
```diff
@@ -0,0 +1,75 @@
...
+        super(parent);
+        this.data = data;
+    }
+
+    public get iconPath(): TreeItemIconPath {
+        return treeUtils.getActionIconPath(this.status, this.conclusion);
+    }
+
+    public get id(): string {
+        return `${this.parent.id}/${this.data.name}`;
```
EricJizbaMSFT, line 33: No need to add `this.parent.id`. It's automatically included in `this.fullId`

src/tree/StepTreeItem.ts
```diff
@@ -0,0 +1,75 @@
...
+    public get description(): string {
+        if (this.data.conclusion !== null) {
+            const elapsedMs: number = this.completedDate.getTime() - this.startedDate.getTime();
+            return `${convertConclusionToVerb(this.conclusion)} in ${prettyMs(elapsedMs)}`;
+        } else {
+            return `${convertStatusToVerb(this.status)} ${this.startedDate.getTime() === 0 ? 'now' : moment(this.startedDate).fromNow()}`;
+        }
+    }
+
+    private get startedDate(): Date {
```
EricJizbaMSFT, line 53: nit: private variables should be prefixed with an underscore

src/tree/JobTreeItem.ts
```diff
@@ -0,0 +1,103 @@
...
+        return new Date(this.data.started_at);
+    }
+
+    private get completedDate(): Date {
+        return new Date(this.data.completed_at);
+    }
+
+    private get status(): Status {
+        if (this.data.status in Status) {
+            return <Status>this.data.conclusion;
```
EricJizbaMSFT, line 67: this.data.status

src/tree/StepTreeItem.ts
```diff
@@ -0,0 +1,75 @@
...
+        return new Date(this.data.started_at);
+    }
+
+    private get completedDate(): Date {
+        return new Date(this.data.completed_at);
+    }
+    private get status(): Status {
+        if (this.data.status in Status) {
+            return <Status>this.data.conclusion;
```
EricJizbaMSFT, line 62: You copied/pasted a bug (conclusion instead of status). This is why we try to avoid copied code haha. Maybe stick a util function (`parseStatus` and `parseConclusion`?) for this in gitHubUtils

src/tree/JobTreeItem.ts
nturinski: I say we at least keep how long ago it ran (i.e. 4 days ago).  I don't know if we need the verbs (like completed for instance) in there since that's kind of the point of icons (though I'm pretty sure that's not accessible, so I dunno).  

I personally like the descriptions, and to your point about being lightweight/simple, while I agree, this level of description only appears if they drill down pretty far into the Actions node, in which case I feel like the user would be trying to get more detailed information about the steps.

The data object (the GitHub response type) doesn't include the times in a human friendly format (they just have completed/started timestamps) so I see some value in displaying that in a easily read manner.

EricJizbaMSFT: I'm fine with how long ago it ran and the conclusion/status (because the icons _aren't_ accessible, as you mentioned). The conclusion/status is most important (gotta know if it worked) and how long ago is second most important (if it ran a month ago, I might need to update my site). I just don't see as much value in the elapsed time - why do I care if it took a minute or an hour?

> this level of description only appears if they drill down pretty far into the Actions node, in which case I feel like the user would be trying to get more detailed information about the steps.

so if it looks gross already - we're allowed to make it even more gross? It's not that far down :P

> The data object (the GitHub response type) doesn't include the times in a human friendly format (they just have completed/started timestamps) so I see some value in displaying that in a easily read manner.

Doesn't mean you have to display it in the description. There's no reason the "properties" object we display has to be the exact object we get from GitHub. We can modify and make it look better as much as we want (still needs to be localizable, though - and out of scope for this PR).

alexweininger:
💯

PR #154, opened 2020-07-21
Start using Octokit API

nturinski:
Using the Octokit API for our GitHub org requests.  I'll start propagating the changes everywhere, but wanted to get this in so Alex could make the changes to his PR https://github.com/microsoft/vscode-azurestaticwebapps/pull/152

~~The way the API uses the accessToken makes me think that it doesn't expire.  I'd see some documentation implying this, but I don't think I've ever seen definitively where that is said.~~

~~But FWIW the GitHub Actions extension also creates a Octokit client on activation and doesn't ever create a new one afterwards.~~

EDIT: lol never mind


src/utils/gitHubUtils.ts
```diff
@@ -113,8 +111,19 @@ export async function createGitHubRequestOptions(gitHubAccessToken: string, url:
 
 export async function getGitHubAccessToken(): Promise<string> {
     const scopes: string[] = ['repo', 'workflow', 'admin:public_key'];
-    return (await vscode.authentication.getSession('github', scopes, { createIfNone: true })).accessToken;
-
+    // https://github.com/microsoft/vscode-azurestaticwebapps/issues/153
+    // tslint:disable-next-line:no-constant-condition
+    while (true) {
```
EricJizbaMSFT, line 116: Do we still need this since you're not getting the token on activation? Would like to avoid a loop like this if possible

nturinski: It could probably still happen, but considering that I had never seen it prior to this PR, it's probably safe to remove.

package.json
```diff
@@ -400,6 +400,8 @@
         "webpack-cli": "^3.3.7"
     },
     "dependencies": {
+        "@octokit/rest": "^18.0.1",
+        "@octokit/types": "^5.1.0",
```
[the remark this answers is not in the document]
EricJizbaMSFT: The point of making it a dev dependency is so that we don't ship it with our vsix and put it on every user's machine. That's more important than some tslint rule

The quick/dirty method would be to disable the rule line-by-line, but the best way to do it is to add "@octokit/types" to the list of allowed implicit dependencies here: https://github.com/microsoft/vscode-azurestaticwebapps/blob/master/tslint.json#L59

PR #156, opened 2020-07-23
Get gitTree 🌲 to display in app/api location steps

nturinski:
Resolves checkbox 1 in https://github.com/microsoft/vscode-azurestaticwebapps/issues/136

This is what it looks like: 
image

image

If there's a subpath setting, if that entry is found, it is moved to the top.

GetTreeData recursively gets all of the files from a repo-- I haven't tested which a huge repo how long that takes, but GitHub has set a limit of 1000 files/folders in its response.

I'm not sure I like how the list looks.  The original order is actually depth first rather than breadth first (how I organized it), but that makes the most sense to me since I imagine the app and api folder would be near  the root (rather than really nested in)

Local experience is to come later, just wanted to check this code in first.


src/commands/createStaticWebApp/ApiLocationStep.ts
```diff
@@ -4,19 +4,27 @@
...
 import { addLocationTelemetry } from "./addLocationTelemetry";
 import { IStaticWebAppWizardContext } from "./IStaticWebAppWizardContext";
 
 export class ApiLocationStep extends AzureWizardPromptStep<IStaticWebAppWizardContext> {
     public async prompt(wizardContext: IStaticWebAppWizardContext): Promise<void> {
-        wizardContext.apiLocation = (await ext.ui.showInputBox({
-            value: getWorkspaceSetting(apiSubpathSetting, wizardContext.fsPath) || defaultApiName,
-            prompt: localize('apiLocation', "Enter the location of your Azure Functions code (Leave blank for no Azure Functions code)"),
-        })).trim();
+        const placeHolder: string = localize('apiLocation', "Enter the location of your Azure Functions code");
```
EricJizbaMSFT, line 18: nit: "Select" instead of "Enter" since it's a quick pick now

src/commands/createStaticWebApp/ApiLocationStep.ts
```diff
@@ -4,19 +4,27 @@
...
+        const input: string = (await ext.ui.showQuickPick(
+            await getGitTreeQuickPicks(
+                nonNullProp(wizardContext, 'gitTreeData'),
+                getWorkspaceSetting(apiSubpathSetting, wizardContext.fsPath),
+                true),
+            { placeHolder, suppressPersistence: true })).label.trim();
```
EricJizbaMSFT, line 24: You should always use `data` instead of `label` for storing the actual value. Too many bugs are caused by a display value (label) being slightly different than the actual value (data). That would mean you don't have to check for `skipForNowQuickPickItem.label` below since you can just make `data` whatever it needs to be in that case

src/utils/gitHubUtils.ts
```diff
@@ -167,3 +169,38 @@ export function isUser(orgData: UsersGetAuthenticatedResponseData | OrgForAuthen
...
     return !!orgData && 'type' in orgData && orgData.type === 'User';
 }
+
+export async function getGitHubTree(repositoryUrl: string, branch: string): Promise<GitTreeData[]> {
+    const octokitClient: Octokit = await createOctokitClient();
+    const { owner, name } = getRepoFullname(repositoryUrl);
+    const branchRes: OctokitResponse<ReposGetBranchResponseData> = await octokitClient.repos.getBranch({ owner, repo: name, branch });
+    const getTreeRes: OctokitResponse<GitGetTreeResponseData> = await octokitClient.git.getTree({ owner, repo: name, tree_sha: branchRes.data.commit.sha, recursive: 'true' });
+
+    return getTreeRes.data.tree.filter(file => file.type === 'tree').sort((f1, f2) => {
```
EricJizbaMSFT, line 179: Two things - 
1. I'd prefer you create a function `getFolderDepth` to avoid copied code
1. When using `sort` it doesn't have to be exactly 1 or -1. It just has to be a positive or negative number. Thus it's more traditional to just subtract the two values. This is shorter and more accurately handles the "0" case.

```
// sort descending by the depth of subfolder
return getTreeRes.data.tree.filter(file => file.type === 'tree').sort((f1, f2) => {
    // tslint:disable-next-line: strict-boolean-expressions
    function getFolderDepth(path: string): number { return (path.match(/\//g) || []).length; }
    return getFolderDepth(f1.path) - getFolderDepth(f2.path);
});
```

src/commands/createStaticWebApp/GitHubBranchListStep.ts
```diff
@@ -23,6 +23,9 @@ export class GitHubBranchListStep extends AzureWizardPromptStep<IStaticWebAppWiz
         } while (!branchData);
 
         context.branchData = branchData;
+
+        // load the gitTree data now as it is used for the next two steps
+        context.gitTreeData = await getGitHubTree(nonNullProp(context, 'repoHtmlUrl'), nonNullValueAndProp(context.branchData, 'name'));
```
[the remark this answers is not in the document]
nturinski: Hmm, I feel like the likelihood of someone selecting the wrong app location isn't that unlikely.  It seems like it'd be more annoying that we have to make this API call and load every time they go back to the app location step.  I do understand it looks weird for the user though-- is there anyway we can preserve a property after hitting the back button?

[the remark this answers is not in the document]
EricJizbaMSFT: Yeah there's lots of ways to do that. One way would be to change `gitTreeData` to `gitTreeDataTask` - aka start the task here, but don't await it until you're inside `showQuickPick`

src/commands/createStaticWebApp/ApiLocationStep.ts
```diff
@@ -3,25 +3,31 @@
...
 
 export class ApiLocationStep extends AzureWizardPromptStep<IStaticWebAppWizardContext> {
     public async prompt(wizardContext: IStaticWebAppWizardContext): Promise<void> {
-        wizardContext.apiLocation = (await ext.ui.showInputBox({
-            value: getWorkspaceSetting(apiSubpathSetting, wizardContext.fsPath) || defaultApiName,
-            prompt: localize('apiLocation', "Enter the location of your Azure Functions code (Leave blank for no Azure Functions code)"),
-        })).trim();
-        addLocationTelemetry(wizardContext, 'apiLocation', defaultApiName);
+        const placeHolder: string = localize('apiLocation', "Select the location of your Azure Functions code");
+        (await ext.ui.showQuickPick(getGitTreeQuickPicks(wizardContext, true), { placeHolder, suppressPersistence: true })).data.trim();
```
EricJizbaMSFT, line 18: Looks like you forgot to set this to `wizardContext.apiLocation`

src/utils/gitHubUtils.ts
```diff
@@ -167,3 +171,34 @@ export function isUser(orgData: UsersGetAuthenticatedResponseData | OrgForAuthen
...
+    // sort descending by the depth of subfolder
+    return getTreeRes.data.tree.filter(file => file.type === 'tree').sort((f1, f2) => {
+        // tslint:disable-next-line: strict-boolean-expressions
+        function getFolderDepth(path: string): number { return (path.match(/\//g) || []).length; }
+        return getFolderDepth(f1.path) - getFolderDepth(f2.path);
+    });
+}
+
+export async function getGitTreeQuickPicks(wizardContext: IStaticWebAppWizardContext, isSkippable?: boolean): Promise<IAzureQuickPickItem<string>[]> {
+    if (!wizardContext.gitTreeData) {
```
EricJizbaMSFT, line 190: You can remove `gitTreeData` from the wizard context. No matter how many times you call `await` on the same task, it's still only executed once, so there's no performance benefit from caching the result on the wizardContext

src/commands/createStaticWebApp/EnterApiLocationStep.ts
```diff
@@ -0,0 +1,24 @@
...
+import { ext } from "../../extensionVariables";
+import { localize } from "../../utils/localize";
+import { getWorkspaceSetting } from "../../utils/settingsUtils";
+import { IStaticWebAppWizardContext } from "./IStaticWebAppWizardContext";
+
+export class EnterApiLocationStep extends AzureWizardPromptStep<IStaticWebAppWizardContext> {
+    public async prompt(context: IStaticWebAppWizardContext): Promise<void> {
+        context.apiLocation = (await ext.ui.showInputBox({
+            value: getWorkspaceSetting(apiSubpathSetting, context.fsPath) || defaultApiLocation,
+            prompt: localize('enterApiLocation', "Enter the location of your Azure Functions code (Leave blank for no Azure Functions code)"),
```
EricJizbaMSFT, line 17: Do we need "(Leave blank for no Azure Functions code)" still? They had "Skip for now" on the previous step

src/commands/createStaticWebApp/ApiLocationStep.ts
```diff
@@ -3,25 +3,31 @@
...
+
+        wizardContext.manuallyEnterApi = wizardContext.apiLocation === enterInputQuickPickItem.data;
```
EricJizbaMSFT, line 20: Here's how I would do this:
- Remove `manuallyEnterApi` from the wizard context
- Set `data` to `undefined` in the "manually enter" case
- Below in `getSubWizard`, check `wizardContext.apiLocation === undefined`

It feels weird to have a "QuickPickItem" as a "constant" (normally the constants.ts file is just for strings/numbers). Plus technically this blocks people from using "manuallyEnter" as the location name (I know - why would they 😝). Finally, it just cleans up the wizardContext a bit

src/commands/createStaticWebApp/ApiLocationStep.ts
```diff
@@ -3,25 +3,30 @@
...
+
+        addLocationTelemetry(wizardContext, 'apiLocation', defaultApiLocation);
     }
 
     public shouldPrompt(wizardContext: IStaticWebAppWizardContext): boolean {
         return !wizardContext.apiLocation;
     }
 
+    public async getSubWizard(wizardContext: IStaticWebAppWizardContext): Promise<IWizardOptions<IStaticWebAppWizardContext> | undefined> {
+        return !wizardContext.apiLocation ? { promptSteps: [new EnterApiLocationStep()] } : undefined;
```
EricJizbaMSFT, line 28: This might work because `EnterApiLocationStep` checks `wizardContext.apiLocation === undefined`, but I'd prefer to check undefined here, too. No need to add the step (and possibly mess up the step count?) if apiLocation is an empty string

Issue #73, opened 2020-05-15
Add commands for Production and Environment in command palette


v-ruizh:
**OS:** Windows
**Build Version:** [20200514.4](https://dev.azure.com/ms-azuretools/AzCodePrivate/_build/results?buildId=19336&view=results)

**Repro Steps:**
Check the commands in command palette.

**Expect:**
There are commands for Production and Environment.

**Actual:**
There is no command for Production and Environment.
![image](https://user-images.githubusercontent.com/45614674/82021922-55767280-96be-11ea-8f0d-3462aa23db11.png)

nturinski:
Implement ways to do command palette stuff on environments other than production.

PR #158, opened 2020-07-23
Browse 👁‍🗨 selects an environment rather than app

nturinski:
Fixes https://github.com/microsoft/vscode-azurestaticwebapps/issues/73

I think the original issue wanted the command names to include "Environment" in them, but I don't think most of the commands would make sense if renamed that way.

I made browse exclusive to environments because I'm trying to really separate the SWA from the environment. 

As for other commands that accept multiple tree items: openInPortal, viewProperties, refresh, etc. the command palette is looking for Static Web Apps because that's what made sense to me (and is the current behavior of other extensions.)

### DIFF
--- a/package.json
+++ b/package.json
@@ -235,7 +235,7 @@
                 },
                 {
                     "command": "staticWebApps.browse",
-                    "when": "view == staticWebApps && viewItem == azureStaticEnvironment",
+                    "when": "view == staticWebApps && viewItem =~ /azureStatic(WebApp|Environment)$/",
                     "group": "1@1"
                 },
                 {
--- a/package.json
+++ b/package.json
@@ -235,7 +235,7 @@
                 },
                 {
                     "command": "staticWebApps.browse",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -235,7 +235,7 @@
                 },
                 {
                     "command": "staticWebApps.browse",
-                    "when": "view == staticWebApps && viewItem =~ /azureStatic(WebApp|Environment)$/",
+                    "when": "view == staticWebApps && viewItem == azureStaticEnvironment",
```
EricJizbaMSFT, line 238: Idk browse seems super important to me, so I worry about making it less discoverable. I think it makes sense what you did for the command palette, but can we keep the context menu on both?

nturinski: I'm kind of mixed on this because I think guiding the user to think that SWA === production environment could lead to some confusion.  The SWA itself doesn't have a website or host, everything in the portal forces you to select an environment before you can configure something.

image

You can see that browse is on the production environment here as well  That being said, this is their overview pane:

image

But with all that said, I'm fine with having browse on both, just some considerations.

Issue #93, opened 2020-05-16
Show the Status of Environment similar to Azure Portal


anganti:
Is it possible display the status of Build/Environment and show the state inline with the Environment/Build node

export enum BuildStatus {
    WaitingForDeployment = 'WaitingForDeployment',
    Uploading = 'Uploading',
    Deploying = 'Deploying',
    Ready = 'Ready',
    Failed = 'Failed',
    Deleting = 'Deleting',
    Detached = 'Detached',
}

![image](https://user-images.githubusercontent.com/42851022/82109234-448f2680-96e9-11ea-8ce6-b7026ca796d4.png)

PR #159, opened 2020-07-24
Add env 🌎 status to description

nturinski:
Fixes https://github.com/microsoft/vscode-azurestaticwebapps/issues/93

The issue mentioned building out an enum for it, but I don't think we need it in the context of this PR.  Also when we start using the SDK, hopefully all of that will be defined for us.

image

image

Also this doesn't auto-refresh yet... Polling anyone? 😐


src/tree/EnvironmentTreeItem.ts
```diff
@@ -69,8 +70,13 @@ export class EnvironmentTreeItem extends AzureParentTreeItem implements IAzureRe
     }
 
     public get description(): string {
-        const linkedDesc: string = localize('linkedTag', '{0} (linked)', this.data.properties.sourceBranch);
-        return this.inWorkspace ? linkedDesc : this.data.properties.sourceBranch;
+        if (this.data.properties.status !== 'Ready') {
+            // if the environment isn't ready, the status has priority over displaying its linked
+            return localize('statusTag', '{0} ({1})', this.data.properties.sourceBranch, this.data.properties.status);
```
EricJizbaMSFT, line 75: Can you paste a screenshot of the portal? Feels like we're getting in the range of too much information on these tree items, too. Maybe we get rid of PR number? The nice thing about the status is it should be hidden for most cases...

src/tree/EnvironmentTreeItem.ts
```diff
@@ -69,8 +70,13 @@ export class EnvironmentTreeItem extends AzureParentTreeItem implements IAzureRe
...
+        }
+
+        const linkedTag: string = localize('linkedTag', '{0} (linked)', this.data.properties.sourceBranch);
+        return this.inWorkspace ? linkedTag : this.data.properties.sourceBranch;
```
EricJizbaMSFT, line 79: > Also this doesn't auto-refresh yet... Polling anyone? 😐

> Do you think it's worth creating a separate issue for the polling stuff?

@wwlorey I think (and hope) he's just joking about the polling because I was giving him a hard time yesterday re polling lol. We generally take a manual-refresh approach for our tree views. Auto-refresh can be quite expensive and unnecessary. Can you imagine if a user had all of our extensions installed and they were all polling on slow internet? The value vs. cost just isn't there

nturinski: Eric is correct ^

nturinski:
image

Issue #128, opened 2020-06-18
[Suggestion] Add a command 'Delete...' for Environment


v-mengwe:
**OS:** Win10
**Build Version:** 20200611.12

**Repro Steps:**
Check the context menu for Environment.

**Actual:**
There is no "Delete..." action.
![image](https://user-images.githubusercontent.com/62053665/84972306-2a65bf80-b151-11ea-9db9-78974bd4c7e6.png)

**Suggestion:**
Add a command 'Delete...' for Environment.

**More info:**
There is "Delete" action in portal.
![image](https://user-images.githubusercontent.com/62053665/84972850-503f9400-b152-11ea-989c-db1269f9cca2.png)

PR #161, opened 2020-07-24
Add delete environment

nturinski:
Fixes https://github.com/microsoft/vscode-azurestaticwebapps/issues/128

I decided to leave the label `Delete...` as is because `Delete Static Web App...` looks terrible.  Maybe I could just shorten it to App or Static App?

I'm using label for Environment's output messages because that's what the user is used to seeing.  The name property is just the build number (that's the name Azure gives it).

Moved the AzureAsyncRequest code to requestUtils, but we should be able to remove this once we move over to the new SDK.

### DIFF
--- a/package.nls.json
+++ b/package.nls.json
@@ -1,6 +1,7 @@
 {
     "staticWebApps.createStaticWebApp": "Create Static Web App...",
     "staticWebApps.deleteStaticWebApp": "Delete...",
+    "staticWebApps.deletedeleteEnvironment": "Delete Environment...",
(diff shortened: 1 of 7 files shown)


package.nls.json
```diff
@@ -1,6 +1,7 @@
 {
     "staticWebApps.createStaticWebApp": "Create Static Web App...",
     "staticWebApps.deleteStaticWebApp": "Delete...",
+    "staticWebApps.deletedeleteEnvironment": "Delete Environment...",
```
EricJizbaMSFT, line 4: deletedeleteEnvironment -> deleteEnvironment

nturinski: What the heck, I remember fixing this...

PR #166, opened 2020-07-29
Open yml file in GitHub

nturinski:
Part 4/4 of https://github.com/microsoft/vscode-azurestaticwebapps/issues/136

I chose "edit" rather than "open" in the command name because we open them in edit mode.  I chose to do this so hopefully users feel empowered to edit this file since users tend not to mess with auto-generated files.

Also making these changes made my head hurt and I think I want to create a `SiteClient` file for these tree items now.


package.nls.json
```diff
@@ -28,5 +28,6 @@
     "staticWebApps.appArtifactSubpath": "The default value for the location of your build output relative to your application code location",
     "staticWebApps.action.rerun": "Rerun Action...",
     "staticWebApps.action.cancel": "Cancel Action...",
-    "staticWebApps.showDocumentation": "Documentation"
+    "staticWebApps.showDocumentation": "Documentation",
+    "staticWebApps.openYAMLConfigFile": "Edit YAML Configuration File..."
```
EricJizbaMSFT, line 32: What about "Edit Configuration..."? Not sure we need to know it's yaml up front

src/commands/openYAMLConfigFile.ts
```diff
@@ -0,0 +1,31 @@
...
+    if (node instanceof EnvironmentTreeItem && node.inWorkspace) {
+        const ymlUri: Uri[] = await workspace.findFiles(ymlFileName);
+        // if we couldn't find it, then try opening it in GitHub
+        if (ymlUri.length > 0) {
+            await window.showTextDocument(await workspace.openTextDocument(ymlUri[0]));
+            return;
+        }
+    }
+
+    await openUrl(`${node.repositoryUrl}/edit/${node.branch}/.github/workflows/azure-static-web-apps-${defaultHostname.replace('.azurestaticapps.net', '')}.yml`);
```
EricJizbaMSFT, line 30: Leverage `ymlFileName` here, too

src/commands/openYAMLConfigFile.ts
```diff
@@ -0,0 +1,31 @@
...
+import { StaticWebAppTreeItem } from "../tree/StaticWebAppTreeItem";
+import { openUrl } from "../utils/openUrl";
+
+export async function openYAMLConfigFile(context: IActionContext, node?: StaticWebAppTreeItem | EnvironmentTreeItem): Promise<void> {
+    if (!node) {
+        node = await ext.tree.showTreeItemPicker<EnvironmentTreeItem>(EnvironmentTreeItem.contextValue, context);
+    }
+
+    const defaultHostname: string = node instanceof StaticWebAppTreeItem ? node.data.properties.defaultHostname : node.parent.data.properties.defaultHostname;
+    const ymlFileName: string = `.github/workflows/azure-static-web-apps-${defaultHostname.replace('.azurestaticapps.net', '')}.yml`;
```
EricJizbaMSFT, line 19:
```
defaultHostname.replace('.azurestaticapps.net', '')
```

This won't work with other Azure environments (Azure China, etc.). Either refactor the logic now (maybe something like `defaultHostname.split('.')[0]`?) or file an issue to fix later (this stuff is easy to forget and hard to find in testing)

src/commands/openYAMLConfigFile.ts
```diff
@@ -0,0 +1,31 @@
...
+
+    if (node instanceof EnvironmentTreeItem && node.inWorkspace) {
+        const ymlUri: Uri[] = await workspace.findFiles(ymlFileName);
```
EricJizbaMSFT, line 22: I'd use `fse.pathExists` instead of `workspace.findFiles`

nturinski: Any particular reason why?  Performance reasons?

EricJizbaMSFT: Yeah should be more efficient

PR #177, opened 2020-07-30
Add telemetry about users configuration

nturinski:
Some telemetry about the user's workspace so we can better understand how users are creating static web apps.

### DIFF
--- a/src/commands/createStaticWebApp/addWorkspaceTelemetry.ts
+++ b/src/commands/createStaticWebApp/addWorkspaceTelemetry.ts
@@ -8,24 +8,21 @@ import { workspace } from 'vscode';
 import { IStaticWebAppWizardContext } from "./IStaticWebAppWizardContext";
 
 export function addWorkspaceTelemetry(wizardContext: IStaticWebAppWizardContext): void {
-    try {
-        wizardContext.telemetry.properties.isGitInstalled = 'true';
+    wizardContext.telemetry.properties.isGitInstalled = 'true';
+    const localGit: git.SimpleGit = git(wizardContext.fsPath);
 
-        const localGit: git.SimpleGit = git(wizardContext.fsPath);
-
-        // tslint:disable-next-line:no-floating-promises
-        localGit.status().then(() => {
-            wizardContext.telemetry.properties.isGitProject = 'true';
-        }).catch((error) => {
-            // tslint:disable: no-unsafe-any
-            if (error.message.indexOf('spawn git ENOENT') >= 0) {
-                wizardContext.telemetry.properties.isGitInstalled = 'false';
-            } else if (error.message.indexOf('fatal: Not a git repository') >= 0) {
-                wizardContext.telemetry.properties.isGitProject = 'false';
-            }
-        }).finally(() => {
-            wizardContext.telemetry.properties.numberOfWorkspaces = !workspace.workspaceFolders ? String(0) : String(workspace.workspaceFolders.length);
-            wizardContext.telemetry.properties.gotRemote = String(!!wizardContext.repoHtmlUrl);
-        });
-    }
+    // tslint:disable-next-line:no-floating-promises
+    localGit.status().then(() => {
+        wizardContext.telemetry.properties.isGitProject = 'true';
+    }).catch((error) => {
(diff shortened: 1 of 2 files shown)


src/commands/createStaticWebApp/addWorkspaceTelemetry.ts
```diff
@@ -0,0 +1,28 @@
...
+import * as git from 'simple-git/promise';
+import { workspace } from 'vscode';
+import { IStaticWebAppWizardContext } from "./IStaticWebAppWizardContext";
+
+export async function addWorkspaceTelemetry(wizardContext: IStaticWebAppWizardContext): Promise<void> {
+    try {
+        wizardContext.telemetry.properties.isGitInstalled = 'true';
+
+        const localGit: git.SimpleGit = git(wizardContext.fsPath);
+        await localGit.status();
```
EricJizbaMSFT, line 15: We prioritize user experience over telemetry, so you should use `.then` instead of `await` to make sure this doesn't delay anything

Something like this: https://github.com/microsoft/vscode-azurefunctions/blob/2fd8a606a339439637f883dd1d2786635dece81d/src/funcCoreTools/getLocalFuncCoreToolsVersion.ts#L32

nturinski: Okay, FWIW this shouldn't take very long despite it being async since it's just a local `git status` command rather than an API call to a remote server.

src/commands/createStaticWebApp/addWorkspaceTelemetry.ts
```diff
@@ -0,0 +1,28 @@
...
+        wizardContext.telemetry.properties.isGitProject = 'true';
+    }).catch((error) => {
+        // tslint:disable: no-unsafe-any
+        if (error.message.indexOf('spawn git ENOENT') >= 0) {
+            wizardContext.telemetry.properties.isGitInstalled = 'false';
+        } else if (error.message.indexOf('fatal: Not a git repository') >= 0) {
+            wizardContext.telemetry.properties.isGitProject = 'false';
+        }
+    }).finally(() => {
+        wizardContext.telemetry.properties.numberOfWorkspaces = !workspace.workspaceFolders ? String(0) : String(workspace.workspaceFolders.length);
```
EricJizbaMSFT, line 25: nit: Not quite sure why you need a `finally` for these. Feels like you could just stick them before `localGit.status()`

src/commands/createStaticWebApp/addWorkspaceTelemetry.ts
```diff
@@ -0,0 +1,28 @@
...
+export function addWorkspaceTelemetry(wizardContext: IStaticWebAppWizardContext): void {
+    wizardContext.telemetry.properties.isGitInstalled = 'true';
+    const localGit: git.SimpleGit = git(wizardContext.fsPath);
+
+    // tslint:disable-next-line:no-floating-promises
+    localGit.status().then(() => {
+        wizardContext.telemetry.properties.isGitProject = 'true';
+    }).catch((error) => {
+        // tslint:disable: no-unsafe-any
+        if (error.message.indexOf('spawn git ENOENT') >= 0) {
```
EricJizbaMSFT, line 19: Always use parseError instead of disabling `no-unsafe-any`. Protects against null-ref exceptions

src/commands/createStaticWebApp/addWorkspaceTelemetry.ts
```diff
@@ -0,0 +1,28 @@
...
+            wizardContext.telemetry.properties.isGitInstalled = 'false';
+        } else if (error.message.indexOf('fatal: Not a git repository') >= 0) {
```
EricJizbaMSFT, line 21: I prefer `message.includes('fatal: Not a git repository')`. Just feels more readable than `>= 0`

src/commands/createStaticWebApp/addWorkspaceTelemetry.ts
EricJizbaMSFT: It's the principle of the thing. Protects us for cases that seem fast on our machines, but are actually noticeable on slower machines

PR #180, opened 2020-07-31
Update npm packages

nturinski:


nturinski:
I think it's fine since npm update shouldn't pick up any major updates that would be breaking.  That being said, this is why we're doing it post-release in case it does break the extension somehow.

But I'll wait for Eric to say something too 😏

EricJizbaMSFT:
Yeah what Nathan said. I wouldn't do this kind of change right before a release, but otherwise it doesn't worry me too much

PR #191, opened 2020-08-31
Convert any reference to master to main in repos we own

nturinski:


src/commands/createStaticWebApp/GitHubRepoListStep.ts
```diff
@@ -31,8 +31,8 @@ export class GitHubRepoListStep extends AzureWizardPromptStep<IStaticWebAppWizar
 
         context.repoData = repoData;
         context.repoHtmlUrl = repoData.html_url;
-        // if this is a new repo, the only branch that will have been created is 'master'
-        context.branchData = repoData.name === createNewRepo ? { name: 'master' } : undefined;
+        // if this is a new repo, the only branch that will have been created is 'main'
+        context.branchData = repoData.name === createNewRepo ? { name: 'main' } : undefined;
```
EricJizbaMSFT, line 35: Hmm do we need to worry about the timing on this one?

Docs say GitHub won't change the default branch name on new repos to "main" until Oct 1

https://github.com/github/renaming#on-october-1-2020-newly-created-repositories-will-default-to-main

nturinski: Well, we'll be releasing this mid/end September, might as well just do it early.  Though I can try not specifying any branch name and see if automatically selects the default.

Issue #185, opened 2020-08-18
Expand Azure Functions support


dariagrigoriu:
Expand Azure Functions support to include .NET Core and Python.

nturinski:
Just to verify, support .NET core and Python in this case is to just enable users to be able to create those function triggers locally, right?  Is the work to support this done in the platform already?

anthonychu:
@nturinski Yes the platform work should be completed to allow .NET and Python function apps. Thanks. /cc @miwebst

nturinski:
So for project languages, we don't support .NET Core by name because it's ambiguous as it could C# or F#.  Would it be sufficient to expand Azure Functions support to include Python and C#?

We only support .NET Core in Functions if you're picking a runtime for the app in Azure.

anthonychu:
Yep this sounds good to me thanks.

PR #198, opened 2020-09-09
Add C# and Python 🐍 as supported language triggers

nturinski:
Fixes https://github.com/microsoft/vscode-azurestaticwebapps/issues/185
Relies on https://github.com/microsoft/vscode-azurefunctions/pull/2355/files

I changed `return !(await fse.pathExists(path.join(folderPath, newName));` to `await fse.access(path.join(folderPath, newName));` because access will throw an error if it doesn't exist.  I felt like it made the logic a little simpler rather than doing pathExists, returning it if it's true, if it's false, go into another statement to check if the file exists, then try/catch that and return the result.

### DIFF
--- a/src/commands/createHttpFunction.ts
+++ b/src/commands/createHttpFunction.ts
@@ -3,7 +3,6 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import * as fse from 'fs-extra';
 import * as path from 'path';
 import * as vscode from 'vscode';
 import { IActionContext, UserCancelledError } from "vscode-azureextensionui";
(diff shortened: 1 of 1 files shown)


src/commands/createHttpFunction.ts
```diff
@@ -61,7 +61,18 @@ function generateSuffixedName(preferredName: string, i: number): string {
 }
 
 async function isNameAvailable(folderPath: string, newName: string): Promise<boolean> {
-    return !(await fse.pathExists(path.join(folderPath, newName)));
+    // if the file or folder doesn't exist, it will throw an error so we have to wrap these in try/catches
+    try {
+        await fse.access(path.join(folderPath, newName));
+    } catch (error) {
```
EricJizbaMSFT, line 67: >  I felt like it made the logic a little simpler rather than doing pathExists, returning it if it's true, if it's false, go into another statement to check if the file exists, then try/catch that and return the result.

IMO, the problem with this is that we don't know why the error occurred. _One_ of the possible errors is that it doesn't exist, but there's more possible errors. It's safer to keep using pathExists

src/commands/createHttpFunction.ts
```diff
@@ -61,7 +61,18 @@ function generateSuffixedName(preferredName: string, i: number): string {
...
+        try {
+            return !(await fse.readdir(folderPath)).find(file => { return file.includes(`${newName}.`); });
```
EricJizbaMSFT, line 69: I think this has several edge cases it doesn't handle well. For example, maybe I have a file named "HttpTrigger.test.cs" and I want to add "HttpTrigger.cs". Remind me again why you have to prompt for name in this extension instead of letting the Functions extension do it? It's just so much easier to validate the name once the Functions extension knows which language it is

nturinski: I think it was because we wanted to have the generic name "endpoint1", but honestly, I think it makes a lot more sense to just let Functions handle it.

src/commands/createHttpFunction.ts
```diff
@@ -22,60 +21,18 @@ export async function createHttpFunction(context: IActionContext): Promise<void>
...
-        }
-        count += 1;
-    }
-
-    newName = await ext.ui.showInputBox({
-        value: newName, prompt: localize('enterHttpFuncName', 'Provide a unique HTTP Function name for your API'),
-        validateInput: async value => await validateFunctionName(folderPath, value)
-    });
-
     await funcApi.createFunction({
```
EricJizbaMSFT, line 27: We might want to check the api version before passing C# as an option, since it won't work unless they have the latest functions extension

src/commands/createHttpFunction.ts
```diff
@@ -22,60 +21,18 @@ export async function createHttpFunction(context: IActionContext): Promise<void>
 
     const funcApi: AzureFunctionsExtensionApi = await getFunctionsApi(context);
 
-    const endpointName: string = 'endpoint';
```
EricJizbaMSFT, line 25: If you want to add this as an option to the api, I'm fine with that

nturinski: You mean like a default name that gets run through `getUniqueFunctionName`

EricJizbaMSFT: Yeah - just something that replaces "HttpTrigger"

src/commands/createHttpFunction.ts
```diff
@@ -85,7 +42,7 @@ async function getFunctionsApi(context: IActionContext): Promise<AzureFunctionsE
             await funcExtension.activate();
         }
 
-        return funcExtension.exports.getApi<AzureFunctionsExtensionApi>('^1.1.0');
+        return funcExtension.exports.getApi<AzureFunctionsExtensionApi>('^1.2.0');
```
EricJizbaMSFT, line 45: Just to be clear, you still have to update this api version in the functions extension

EricJizbaMSFT (approved):

src/commands/createHttpFunction.ts
nturinski: Yeah, I'm going to fix that weird bug we saw in stand-up before I update the api version.

PR #208, opened 2020-09-21
Move "skip for now" to appear near the top, so the user can see it.

johnpapa:
otherwise, even with a small repo, the "skip for now" may be after 20 or 30 items. The user may not scroll that far and may never see it as an option.

@fiveisprime


src/utils/gitHubUtils.ts
```diff
@@ -191,15 +191,21 @@ export async function getGitTreeQuickPicks(wizardContext: IStaticWebAppWizardCon
...
     const quickPicks: IAzureQuickPickItem<string | undefined>[] = gitTreeData.map((d) => { return { label: d.path, data: d.path }; });
+
+    // the "skip for now" should appear after the root, so we add that first.
+    // the "skip for now" should appear near the top, so the user can see it.
+    const skipForNowQuickPickItem: IAzureQuickPickItem<string> = {
+      label: localize('skipForNow', '$(clock) Skip for now'),
+      data: '',
+    };
+    if (isSkippable) {
+      quickPicks.unshift(skipForNowQuickPickItem);
```
nturinski, line 202: Regarding having the skip for now button closer to the top, do you think it'd be better if we just had it as the first option?  That way if a user is hitting enter to get through all the prompts, they'd automatically hit skip rather than root.

johnpapa: I debated on that. I didnt put it there because I wasnt sure if there is a unwritten convention for what goes first.

But yeah, I think making "skip" first makes a lot of sense.

nturinski: I agree.  Could you update the PR to reflect that change? 😃

johnpapa: done

johnpapa:
I had to catch up to main and merge. Then I had to try and get the formatting correct. This was difficult as there is not formatting file in this project. Let me know if this is good.

I will make a prettier PR separately. This is important for OSS to have some formatter (any formatter) with whatever settings you prefer (any settings). Otherwise, PRs become difficult.

nturinski:
Seems like the merge did some odd things.  I'll go ahead and make the changes we discussed in another PR that's branching off main's current state since this isn't that huge of a change.

johnpapa:
I'm happy to close this and reopen. Would you like that?

nturinski:
It's okay, I'm drafting up a new PR now 👍.  I'll go ahead and close this when I open the new PR.

nturinski:
Closing in favor of https://github.com/microsoft/vscode-azurestaticwebapps/pull/222

EricJizbaMSFT:
> This is important for OSS to have some formatter (any formatter) with whatever settings you prefer (any settings). Otherwise, PRs become difficult.

Fwiw, these settings exist [here](https://github.com/microsoft/vscode-azurestaticwebapps/blob/main/.vscode/settings.json), the only caveat is you have to have the root of the repo as your workspace folder in VS Code. Also, some of the formatting relies on the tslint extension (I know tslint is deprecated in favor of eslint - we'll make the move eventually), which should be recommended when you open our repo based on [this](https://github.com/microsoft/vscode-azurestaticwebapps/blob/main/.vscode/extensions.json#L3)

johnpapa:
@EricJizbaMSFT thanks. I didnt see those, and I'm not sure why they didnt work for me. I opened in the root. oh well

I didnt see them because most OSS I work with use prettier or tools that are editor agnostic.

PR #214, opened 2020-09-25
ci: add Azure Static Web Apps workflow file

MohammadFaleh:
on-behalf-of: @Azure opensource@microsoft.com


nturinski (commented):
Sorry, but could you tell me what the purpose of this workflow file is?  It looks like something that was copied from a static web app's workflow folder.

PR #215, opened 2020-09-30
Remove any custom GitHub typings and API calls in favor of OctoKit

nturinski:


src/utils/gitHubUtils.ts
```diff
@@ -4,42 +4,35 @@
...
-// tslint:disable-next-line:no-reserved-keywords
-export type gitHubRepoData = { name: string; url: string; html_url: string; clone_url?: string; default_branch?: string };
-export type gitHubBranchData = { name: string };
-export type gitHubLink = { prev?: string; next?: string; last?: string; first?: string };
-export type gitHubWebResource = requestUtils.Request & { nextLink?: string };
+type gitHubLink = { prev?: string; next?: string; last?: string; first?: string };
 
-export async function getGitHubJsonResponse<T>(requestOptions: gitHubWebResource): Promise<T> {
+// tslint:disable:no-any no-unsafe-any
+export async function getGitHubJsonResponse<T, U>(gitHubApiCb: (params: U) => Promise<OctokitResponse<any>>, params: U & { page?: number }): Promise<T> {
```
EricJizbaMSFT, line 24: How about we get rid of `getGitHubJsonResponse`. Seems like it's only used once

src/utils/gitHubUtils.ts
```diff
@@ -4,42 +4,35 @@
...
-    // Note: blank after user implies look up authorized user
-    const gitHubResponse: Response = await requestUtils.sendRequest(requestOptions);
-    if (gitHubResponse.headers.link) {
-        const headerLink: string = <string>gitHubResponse.headers.link;
-        const linkObject: gitHubLink = parseLinkHeaderToGitHubLinkObject(headerLink);
-        requestOptions.nextLink = linkObject.next;
+    const res: OctokitResponse<any> = await gitHubApiCb(params);
+    if (res.headers.link) {
+        const linkObject: gitHubLink = parseLinkHeaderToGitHubLinkObject(res.headers.link);
+        const page: string | null | undefined = linkObject.next ? new URL(linkObject.next).searchParams.get('page') : undefined;
```
EricJizbaMSFT, line 30: Is 'page' officially supported/documented? I feel like I trust a next link more than I trust a page query param within a next link

src/commands/createStaticWebApp/GitHubBranchListStep.ts
```diff
@@ -3,23 +3,28 @@
...
 
+type ReposListBranchesParameters = RestEndpointMethodTypes['repos']['listBranches']['parameters'];
+
 export class GitHubBranchListStep extends AzureWizardPromptStep<IStaticWebAppWizardContext> {
     public async prompt(context: IStaticWebAppWizardContext): Promise<void> {
-        let branchData: gitHubBranchData | undefined;
+        let branchData: BranchData | undefined;
         const { owner, name } = getRepoFullname(nonNullProp(context, 'repoHtmlUrl'));
         const placeHolder: string = localize('chooseBranch', 'Choose branch for repository "{0}/{1}"', owner, name);
-        const requestOption: gitHubWebResource = await createGitHubRequestOptions(context.accessToken, `${githubApiEndpoint}/repos/${owner}/${name}/branches`);
```
EricJizbaMSFT, line 19: We use `context.accessToken` here because we want to use the same token for the whole wizard instead of re-getting a token each time (which could affect performance and/or prompt the user for password unless they pick "Always Allow"). Can you find a way to replicate that with the new octokit stuff? A few options:
1. Add an optional 'token' param to `createOctokitClient` that would bypass `getGitHubAccessToken`
1. Add the client itself to the wizard context

src/commands/github/actionCommands.ts
```diff
@@ -3,24 +3,23 @@
...
     }
 
-    const token: string = await getGitHubAccessToken();
-    const gitHubRequest: gitHubWebResource = await createGitHubRequestOptions(token, node.data.rerun_url, 'POST');
     const rerunRunning: string = localize('rerunRunning', 'Rerun for action "{0}" has started.', node.data.id);
     ext.outputChannel.appendLog(rerunRunning);
 
-    await requestUtils.sendRequest(gitHubRequest);
+    const client: Octokit = await createOctokitClient();
+    await client.actions.reRunWorkflow({ owner: node.data.repository.owner.login, repo: node.data.repository.name, run_id: node.data.id });
```
EricJizbaMSFT, line 22: Just double-checking - does this `owner.login` thing work for both user and org repositories?

src/gitHubTypings.ts
```diff
@@ -3,12 +3,14 @@
...
 
-import { ActionsGetJobForWorkflowRunResponseData, GitGetTreeResponseData, OrgsListForAuthenticatedUserResponseData } from "@octokit/types";
+import { ActionsGetJobForWorkflowRunResponseData, GitGetTreeResponseData, OrgsListForAuthenticatedUserResponseData, ReposListBranchesResponseData, ReposListForOrgResponseData } from "@octokit/types";
 
 // we'll export the types until this is fixed: https://github.com/octokit/types.ts/issues/120
 export type OrgForAuthenticatedUserData = OrgsListForAuthenticatedUserResponseData[0];
 export type GitTreeData = GitGetTreeResponseData['tree'][0];
 export type ActionWorkflowStepData = ActionsGetJobForWorkflowRunResponseData['steps'][0];
+export type RepoData = ReposListForOrgResponseData[0];
+export type BranchData = ReposListBranchesResponseData[0];
```
wwlorey, line 13: Is the zero index on these types just a way of converting an array to a single element of that type? I haven't seen that pattern before

src/utils/gitHubUtils.ts
nturinski: https://developer.github.com/v3/#link-header

The page is a query parameter for the API call itself.  Unless we keep using custom requests, I don't think I can use the next link that is retrieved directly anymore.

It's also annoyingly a number parameter (which is why I have to convert the query param from a string to a number).  I didn't see any other way to retrieve the next-page from the API response.

src/commands/github/actionCommands.ts
nturinski: Yep yep, owner.login will be what you need to pass in for both cases.  This is an org:

image

src/gitHubTypings.ts
nturinski: Yeah... this is a weird thing.

image
image

Notice that the typing ends with a [].  Dunno _why_ that they did this, but basically the data type that I want to extract lives in the 0 index.  Rather than use `ReposListBranchesResponseData[0]`, I just extracted the type that I wanted and renamed it to what it is (which is just RepoData)

EricJizbaMSFT (approved):

PR #222, opened 2020-10-16
Move skip/manual enter closer to the top of qp list

nturinski:
Co-authored with @johnpapa 

As mentioned in this PR https://github.com/microsoft/vscode-azurestaticwebapps/pull/208, Skip for Now and Manual Enter being at the bottom really hurts users with repos more than 20-30 folders.  Which, since our logic looks into subfolders as well, seems very likely to happen.

I also changed the label for root because it looked so strange to just be a slash.

Before:
image

After:
image


johnpapa:
Looking at this, I am wondering if we should have put the root slash after the manually enter option. That way the root is right before the rest of the folders. Thoughts?

nturinski:
I'd like to know @fiveisprime 's thoughts on it, but to me, root should be more commonly picked than manual input so it makes sense for it to be higher.  It also looks kind of odd to me in the quickPick list:

image

That being said, I don't have any strong objection to it.

johnpapa:
I could go either way

EricJizbaMSFT (approved):
Either option is fine with me, but I think I lean towards putting the root folder after "Manually enter" like John said. Just feels like it makes more sense if all folder options are together

johnpapa:
Yay!

Thanks y'all!

Issue #195, opened 2020-09-07
The branch name of the Environment should be "branchname" instead of "GitHubaccount:branchname"


v-mengwe:
**OS:** Win10
**Build Version:** 20200906.1

**Repro Step:**

1. Expand a Static Web App.
2. Create an Environment.
3. Check the branches name of Production and Environment.

**Expect:**
The branch name of the Environment is "**branchname**", e.g. `master`.

**Actual:**
The branch name of the Environment is "**GitHubaccount**:**branchname**".
![image](https://user-images.githubusercontent.com/62053665/92373469-ef13d500-f130-11ea-9181-b2aa6c81d878.png)

nturinski:
Sorry, I'm unclear here.  The test environment's branch should _also_ be master?

v-mengwe:
Create an Environment by the following commands:
```
git checkout -b new
git add .
git commit -m 'api'
git push origin head
```
The branch name of the Environment should be "new" instead of "v-mengwe:new".

PR #233, opened 2020-10-30
Remove github username from source branch

nturinski:
Fixes https://github.com/microsoft/vscode-azurestaticwebapps/issues/195

The API now returns source branch as `githubUsername:sourceBranch` for any build that is not the production.  Really strange behavior, but you can see it here
image
And here on the Portal as well
image

I don't have any older SWAs with an environment to see if this was a retroactive change or not.  Branch names are not allowed to have a colon though, so my split should be safe.

### DIFF
--- a/src/tree/EnvironmentTreeItem.ts
+++ b/src/tree/EnvironmentTreeItem.ts
@@ -58,6 +58,12 @@ export class EnvironmentTreeItem extends AzureParentTreeItem implements IAzureRe
         this.branch = nonNullProp(this.data, 'sourceBranch');
 
         this.isProduction = this.buildId === 'default';
+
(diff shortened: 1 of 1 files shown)


wwlorey:
Do you think its worth fixing this since it shows `username:branch` on the portal too? I wonder if there's a reason the API changed like this

nturinski:
It's worth fixing in the sense that it breaks a lot of things in the extension (such as view in github, retrieving the actions, etc.) because those rely on the branch name to be retrieved.  We don't have to change how we display it though.

But yeah, it's broken on the portal too.  The reason they changed it is most likely because if the intent is a SWA to be a collaborative project, only seeing the branch name is kind of unhelpful if people don't label their branches with their initials (like we do).  That'd be my guess, at least.

wwlorey (approved):

Issue #19, opened 2020-05-07
[Discussion] Reduce number of prompts for basic create


EricJizbaMSFT:
We've discussed this offline, but I didn't see an issue for it. 8 feels like a lot of prompts for basic create. We discussed various ideas like using the GitHub remote from their local project or automatically defaulting some of the folders, but each of those could be problematic. Will need further thought

EDIT: Seems like the GitHub org _is_ defaulted to local project, but we're still at 7 steps

PR #236, opened 2020-11-03
Use default branch for basic create scenario

nturinski:
Fixes https://github.com/microsoft/vscode-azurestaticwebapps/issues/19

This further reduces the basic create prompts by 1.  I think that realistically, this is as small as we'll get until we can reliably detect the location of the app/api folders.

I don't see other extensions that have advanced create in the wizardContext, but it seemed unavoidable to me. 😐

### DIFF
--- a/src/commands/createStaticWebApp/IStaticWebAppWizardContext.ts
+++ b/src/commands/createStaticWebApp/IStaticWebAppWizardContext.ts
@@ -13,6 +13,7 @@ export type CreateNewResource = { name?: string; html_url?: string };
 export interface IStaticWebAppWizardContext extends IResourceGroupWizardContext {
     accessToken: string;
     client: WebSiteManagementClient;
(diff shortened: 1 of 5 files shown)


src/commands/createStaticWebApp/IStaticWebAppWizardContext.ts
```diff
@@ -13,6 +13,7 @@ export type CreateNewResource = { name?: string; html_url?: string };
 export interface IStaticWebAppWizardContext extends IResourceGroupWizardContext {
```
EricJizbaMSFT, line 13: Rather than adding "advancedCreation", just do this:

```suggestion
export interface IStaticWebAppWizardContext extends IResourceGroupWizardContext, ICreateChildImplContext {
```

EricJizbaMSFT (approved):

Issue #117, opened 2020-06-05
Improve experience when there is no branch for selected repo


nturinski:
Today if you select a repo that has no branch, then you will see this:

![image](https://user-images.githubusercontent.com/5290572/83909371-8544f300-a71d-11ea-957a-0c079b1da9ab.png)

This is a pretty bad experience especially if you are just "entering" through it, because it will just exit without any indicator as to why.  We could possibly give the option of initializing the repo with whatever local project that the user currently has opened.

PR #244, opened 2020-11-04
Check for no remote branches

nturinski:
Fixes https://github.com/microsoft/vscode-azurestaticwebapps/issues/117

While ideally, we could just create a remote branch for the user, it's a complicated UI experience due to the fact that we can't guarantee that the repo they selected is what they currently have in the local workspace.  GitHub doesn't have an API call to create a new branch (as far as I know) so an error message seems better than nothing.

If anyone knows how to make it more of an error message or change the color of the message/icon, let me know.  I looked through the VS Code API and didn't see anything about it.

Before: 
image

After:
image

### DIFF
--- a/src/commands/createStaticWebApp/GitHubBranchListStep.ts
+++ b/src/commands/createStaticWebApp/GitHubBranchListStep.ts
@@ -38,6 +38,8 @@ export class GitHubBranchListStep extends AzureWizardPromptStep<IStaticWebAppWiz
 
     private async getBranchPicks(context: IStaticWebAppWizardContext, params: ReposListBranchesParameters, picksCache: ICachedQuickPicks<BranchData>): Promise<IAzureQuickPickItem<BranchData | undefined>[]> {
         const client: Octokit = await createOctokitClient(context.accessToken);
-        return await getGitHubQuickPicksWithLoadMore<BranchData, ReposListBranchesParameters>(picksCache, client.repos.listBranches, params, 'name');
+        const picks: IAzureQuickPickItem<BranchData | undefined>[] = await getGitHubQuickPicksWithLoadMore<BranchData, ReposListBranchesParameters>(picksCache, client.repos.listBranches, params, 'name');
+        const noBranch: string = localize('noBranch', ' $(stop) No branches detected. Go back to select a different repository or create a remote branch');
+        return picks.length > 0 ? picks : [{ label: noBranch, data: undefined }];
     }
 }


nturinski:
I guess instead of a pick item, I could change the placeHolder though I don't know how immediately obvious that would be.

src/commands/createStaticWebApp/GitHubBranchListStep.ts
```diff
@@ -38,6 +38,8 @@ export class GitHubBranchListStep extends AzureWizardPromptStep<IStaticWebAppWiz
 
     private async getBranchPicks(context: IStaticWebAppWizardContext, params: ReposListBranchesParameters, picksCache: ICachedQuickPicks<BranchData>): Promise<IAzureQuickPickItem<BranchData | undefined>[]> {
         const client: Octokit = await createOctokitClient(context.accessToken);
-        return await getGitHubQuickPicksWithLoadMore<BranchData, ReposListBranchesParameters>(picksCache, client.repos.listBranches, params, 'name');
+        const picks: IAzureQuickPickItem<BranchData | undefined>[] = await getGitHubQuickPicksWithLoadMore<BranchData, ReposListBranchesParameters>(picksCache, client.repos.listBranches, params, 'name');
+        const noBranch: string = localize('noBranch', ' $(stop) No branches detected. Go back to select a different repository or create a remote branch');
```
wwlorey, line 42: If only the 🛑 would work 😔

src/commands/createStaticWebApp/GitHubBranchListStep.ts
```diff
@@ -38,6 +38,8 @@ export class GitHubBranchListStep extends AzureWizardPromptStep<IStaticWebAppWiz
...
+        return picks.length > 0 ? picks : [{ label: noBranch, data: undefined }];
```
wwlorey, line 43: If `noBranch` is selected, does the wizard try and continue with the next prompt?

wwlorey (approved):
> I guess instead of a pick item, I could change the placeHolder though I don't know how immediately obvious that would be.

I think the pick item is good. It seems to stand out better than the placeholder text

src/commands/createStaticWebApp/GitHubBranchListStep.ts
nturinski: No, because the data is `undefined` it'll be stuck in the while loop and try to get the branches again.  If a user _did_ actually create a new remote branch, it should pick it up next time.

Issue #206, opened 2020-09-18
[Suggestion] Add a message to remind that the special characters will replace with '-' when creating a new GitHub repository


v-mengwe:
**OS:** All
**Build Version:** 20200917.3

**Repro steps:**

1. Click "Create Static Web App..." button -> Enter a name for the new Static Web App -> Select a organization -> Choose "+ Create a new GitHub repository" -> Enter the new GitHub repository with name "@@@@" -> Enter "app" -> Enter "api" -> Enter the default publish directory -> Select a location.
2. Check the result.

**Actual:**
The repository name of next to the Static Web App displays "-" instead of  "@@@@".
![image](https://user-images.githubusercontent.com/62053665/93580045-d1b4f580-f9d1-11ea-9933-87689714f407.png)

**Suggestion:**
Add a message to remind that the special characters will replace with '-', such as:
![image](https://user-images.githubusercontent.com/62053665/93580489-76cfce00-f9d2-11ea-8d2c-54cb064c4102.png)

**More Info:**

1. The repository name displayed in the output logs is "@@@@".
 ![image](https://user-images.githubusercontent.com/62053665/93580805-d8903800-f9d2-11ea-8d9d-301753263b78.png)
2. This issue also reproduces for other special characters.

Issue #181, opened 2020-08-03
Deployment of a React site fails when creating new repo from Wizard


chrisdias:
VS Code Stable
Extension version `0.3.0`
Running on MacOS

```
npx create-react-app myreactapp
cd myreactapp
yarn build
code . -r
```

try to deploy the app using the extension:

- site name: `cdias-react-app...`
- organization: `chrisdias`
- created a new github repository `https://github.com/chrisdias/cdias-react-app-repo`

- location of app folder: `/`
- location of azure functions code: << blank >>
- path of build outputs: `build`
- location: `West US2`

Creating the repository, resource group, and site all work, the action is started.  

Deployment fails on the `Build and Deploy` job, with this:

```
[...]

PR #246, opened 2020-11-09
Publish new repo directly to GitHub

nturinski:
This is a big'un.  I'll demo the full experience on Tuesday so don't worry too much about reviewing it until after then, just wanted to get it up and have eyes on it.

Fixes https://github.com/microsoft/vscode-azurestaticwebapps/issues/181
Fixes https://github.com/microsoft/vscode-azurestaticwebapps/issues/206

Though rather than alert the user, we just sanitize it and reflect it in the quickpick item for them (this is what VS Code does)


src/commands/createStaticWebApp/RepoCreateStep.ts
```diff
@@ -16,26 +20,83 @@ import { IStaticWebAppWizardContext } from './IStaticWebAppWizardContext';
...
+        if (!wizardContext.repo?.state.HEAD?.commit) {
+            // needs to have an initial commit
+            await wizardContext.repo?.commit(localize('initCommit', 'Initial commit'), { all: true });
+        }
+
+        let remoteName: string = 'origin';
+        // ask this before we create the repo so that the user can cancel
+
+        if (wizardContext.repo?.state.remotes.find((remote) => { return remote.name === remoteName; })) {
+            const rewriteMsgBtn: MessageItem = { title: localize('rename', 'Rename') };
```
nturinski, line 50: Rename doesn't feel accurate in this context... What's a better verb?

src/utils/workspaceUtils.ts
```diff
@@ -3,10 +3,47 @@
...
+            canSelectMany: false,
+            defaultUri: workspace.workspaceFolders && workspace.workspaceFolders.length > 0 ? workspace.workspaceFolders[0].uri : undefined,
+            openLabel: localize('select', 'Select')
+        },
+        getSubPath);
+}
+
+export async function selectWorkspaceItem(placeHolder: string, options: OpenDialogOptions, getSubPath?: (f: WorkspaceFolder) => string | undefined | Promise<string | undefined>): Promise<string> {
+    let folder: IAzureQuickPickItem<string | undefined> | undefined;
+    if (workspace.workspaceFolders) {
```
EricJizbaMSFT, line 33: The scenario where I have no folders open was particularly jarring for me. Maybe for static web apps we need to show a quick pick even if "Browse..." is the only option - just to make sure users see the placeholder

src/commands/createStaticWebApp/RepoCreateStep.ts
```diff
@@ -16,26 +20,83 @@ import { IStaticWebAppWizardContext } from './IStaticWebAppWizardContext';
...
     }
 
     public shouldExecute(wizardContext: IStaticWebAppWizardContext): boolean {
-        return !!(wizardContext.accessToken && wizardContext.newRepoName);
+        return !!(wizardContext.accessToken && wizardContext.newRepo);
+    }
+
+    private async getRemoteName(wizardContext: IStaticWebAppWizardContext, projectPath: string): Promise<string> {
+        return await ext.ui.showInputBox({
+            placeHolder: localize('enterRemote', 'Enter a unique remote name'), validateInput: async (value) => {
```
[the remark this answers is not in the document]
nturinski: I'll try using remote shortname ala 
> To add a new remote Git repository as a shortname you can reference easily, run git remote add [shortname] [url]:

https://git-scm.com/book/be/v2/Git-Basics-Working-with-Remotes

nturinski (commented):
So I'm trying out a new design pattern of instead of specifying at the wizard level which steps to push into the promptSteps based on advanced creation, we just push everything and let shouldPrompt handle the behavior.  What do you think?  It feels harder to understand/trace, but less prone to user error.

src/commands/createRepo/WorkspaceListStep.ts
```diff
@@ -0,0 +1,52 @@
...
+
+        // this command doesn't work if the project is not in the workspace, however after running "init" on it, it will generate the Repository object without affecting the repo itself
+        // the downside to this is that it is creating a repo if it didn't exist already
+        // tslint:disable-next-line: strict-boolean-expressions
+        wizardContext.repo = git.getRepository(uri) || await git.init(uri);
+    }
+
+    public shouldPrompt(_wizardContext: IStaticWebAppWizardContext): boolean {
+        // regardless of prompting or not, this has to run to get the workspace/repo
+        return true;
```
ejizba, line 43: I think this will mess with the step count, though, and I don't like how many explanatory comments you've had to use in this file. Don't get me wrong - I like explanatory comments when things are confusing, but I'm saying there's a bit too much confusing stuff going on.

I feel like the fix would be to move most of the code into `DeploymentMethodStep.getSubWizard` and clean up `shouldPrompt` for the other steps. Happy to help work out the kinks offline

src/commands/createStaticWebApp/DeploymentMethodStep.ts
```diff
@@ -0,0 +1,54 @@
...
+        const existingRepo: IAzureQuickPickItem<DeploymentMethods> = { label: localize('existingRepo', '$(github) Use existing GitHub repository'), data: 'existingRepo' };
+
+        const placeHolder: string = localize('createMethod', "How do you want to create a static web app?");
+        const learnMore: IAzureQuickPickItem<DeploymentMethods> = { label: localize('learnMore', '$(book) Learn more...'), data: 'learnMore' };
+        let pick: DeploymentMethods;
+
+        do {
+            pick = (await ext.ui.showQuickPick([localCode, existingRepo, learnMore], { placeHolder, suppressPersistence: true })).data;
+            if (pick === learnMore.data) {
+                await openUrl('https://aka.ms/AAaee7c');
```
ejizba, line 31: Not sure the link you picked is super helpful. I think I'd prefer you write up some of your own documentation and stick it in our wiki because I'm not sure you're going to find an existing doc specific to our scenario.

src/commands/createStaticWebApp/DeploymentMethodStep.ts
```diff
@@ -0,0 +1,54 @@
...
+            }
+        } while (pick === learnMore.data);
+
+        wizardContext.deploymentMethod = pick;
+    }
+
+    public shouldPrompt(wizardContext: IStaticWebAppWizardContext): boolean {
+        return !wizardContext.apiLocation;
```
ejizba, line 39: This should check against `deploymentMethod`, not `apiLocation `

src/commands/createStaticWebApp/DeploymentMethodStep.ts
```diff
@@ -0,0 +1,54 @@
...
+import { localize } from "../../utils/localize";
+import { openUrl } from "../../utils/openUrl";
+import { RepoCreateStep } from "../createRepo/RepoCreateStep";
+import { WorkspaceListStep } from "../createRepo/WorkspaceListStep";
+import { GitHubBranchListStep } from "./GitHubBranchListStep";
+import { GitHubOrgListStep } from "./GitHubOrgListStep";
+import { GitHubRepoListStep } from "./GitHubRepoListStep";
+import { IStaticWebAppWizardContext } from "./IStaticWebAppWizardContext";
+
+export type DeploymentMethods = 'local' | 'existingRepo' | 'learnMore';
```
ejizba, line 17: The 'learnMore' option is only used in this file, so I would not include it in the type exported to other files. I'd probably do `IAzureQuickPickItem<DeploymentMethods | undefined>` and make data `undefined` for the learnMore pick

src/commands/createStaticWebApp/IStaticWebAppWizardContext.ts
```diff
@@ -6,14 +6,18 @@
...
 import { BranchData, GitTreeData, OrgForAuthenticatedUserData, RepoData } from '../../gitHubTypings';
+import { DeploymentMethods } from './DeploymentMethodStep';
 
 // creating a dummy repoData/branchData would be an annoying amount of work, so use this type to recognize when users have selected create new repo
 export type CreateNewResource = { name?: string; html_url?: string };
 export interface IStaticWebAppWizardContext extends IResourceGroupWizardContext, ICreateChildImplContext {
     accessToken: string;
     client: WebSiteManagementClient;
 
+    deploymentMethod?: DeploymentMethods;
```
ejizba, line 19: `deploymentMethod` feels too similar to "local git" vs. "zip deploy". In fact, I would say the deploymentMethod in both cases is "GitHub"

How about we call this property `createScenario` with `publishToNewRepo` and `connectExistingRepo` as the options

src/tree/SubscriptionTreeItem.ts
```diff
@@ -63,11 +61,8 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
             executeSteps.push(new ResourceGroupCreateStep());
         }
 
-        promptSteps.push(new GitHubOrgListStep());
-        promptSteps.push(new GitHubRepoListStep());
-        if (context.advancedCreation) {
-            promptSteps.push(new GitHubBranchListStep());
-        }
+        // DeploymentMethodStep will add a sub-wizard to determine how to deploy the code that will get run before BuildPresetListStep
+        promptSteps.push(new DeploymentMethodStep());
```
ejizba, line 65: What would you think of moving this step first? I just kind of feel like it makes more sense to ask what you're deploying before you ask where you're deploying it to

src/commands/createRepo/GitignoreListStep.ts
```diff
@@ -0,0 +1,39 @@
...
+        result.forEach(file => ignored.delete(file.label));
+
+        if (ignored.size > 0) {
+            const data: string = [...ignored].map(i => `/${i}`).join(os.EOL);
+            await fse.writeFile(gitignorePath, data);
+        }
+    }
+
+    public shouldPrompt(wizardContext: IStaticWebAppWizardContext): boolean {
+        const fsPath: string = nonNullProp(wizardContext, 'fsPath');
```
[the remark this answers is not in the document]
nturinski: Well, this is separate from the advancedCreation check.  Do you think that's okay? 

As far as this goes, yeah I agree with that.  I guess I could set a boolean in the context and check it immediately after the user selects the project that they intend to deploy?

src/commands/createRepo/RepoCreateStep.ts
```diff
@@ -8,34 +8,50 @@ import { ReposCreateForAuthenticatedUserResponseData, ReposCreateInOrgResponseDa
...
-        const name: string = nonNullProp(wizardContext, 'newRepoName');
+        const repo: Repository = nonNullProp(wizardContext, 'repo');
+        if (!repo.state.HEAD?.commit) {
+            // needs to have an initial commit
+            await repo.commit(localize('initCommit', 'Initial commit'), { all: true });
+        }
 
         const client: Octokit = await createOctokitClient(wizardContext.accessToken);
-        const gitHubRepoRes: RepoCreateData = (isUser(wizardContext.orgData) ? await client.repos.createForAuthenticatedUser({ name }) : await client.repos.createInOrg({ org: nonNullProp(wizardContext, 'orgData').login, name })).data;
+        const gitHubRepoRes: RepoCreateData = (isUser(wizardContext.orgData) ? await client.repos.createForAuthenticatedUser({ name: newRepoName, private: newRepoPrivacy }) :
```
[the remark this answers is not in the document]
nturinski: So I actually had the opposite thought as you.  Of course, this is as someone who's making a _lot_ of test repos, but it's pretty annoying to delete them.

From a user's perspective, I agree that it'd be annoying to have local repo changes but we don't actually do anything unless they don't have a repo with a commit themselves.  I would anticipate most users would, but you lay out a compelling argument.

src/commands/createRepo/WorkspaceListStep.ts
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        // calling to verify the user has git enabled so they don't go through the whole process and then it fails
+        const git: API = await getGitApi();
+        const uri: Uri = Uri.file(wizardContext.fsPath);
+
+        // this command doesn't work if the project is not in the workspace, however after running "init" on it, it will generate the Repository object without affecting the repo itself
+        // the downside to this is that it is creating a repo if it didn't exist already
+        // tslint:disable-next-line: strict-boolean-expressions
+        wizardContext.repo = git.getRepository(uri) || await git.init(uri);
```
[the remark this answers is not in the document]
nturinski: I agree-- it's a limitation on the API I'm using.  I think what I'll do alternatively is just use cpUtils to check if a repo exists there.  That should work even if the project isn't in the active workspace.

src/commands/createRepo/WorkspaceListStep.ts
nturinski: I think I have a solution for this now that you bring it up!  I'll let you know if I run into any snags and need some help engineering 🤓

src/commands/createStaticWebApp/IStaticWebAppWizardContext.ts
nturinski: Works for me.  We'll have another one called `deployToAzure` or `zipdeploy` or something, but should all make sense still.

src/tree/SubscriptionTreeItem.ts
nturinski: Like before the name step?  It's a little tricky since I'm using subwizards though, because ideally it'd be pick how you're deploying, pick what you're deploying, name, and then all the other steps (like selecting the repo if you are using GitHub or picking a name for the repo if you're publishing to a new one).

How are you envisioning the order?

src/commands/createStaticWebApp/IStaticWebAppWizardContext.ts
nturinski: I ran into a _real_ weird feeling scenario.  If you have workspace that has an existing GitHub repo and do a basic create, you'll pick connect to an existing GitHub repo, but it'll skip all of the org/repo/branch steps and jump immediately to preset builds which made me super confused.

I'm thinking that we should get rid of that auto-detect logic for basic create and just add it to the list of potential CreateScenarios.  Like if the project that you have opened is on GitHub and you are owner to it, put it in the quicklist as "Create static web app with {owner}/{repo}" or maybe use it's .git url?

ejizba: Oh I guess I thought basic create would automatically select `connectExistingRepo` in that case. It would be less "weird" without the prompt, right?

nturinski: I guess that makes sense.  I never really loved the auto-select though.  I always felt like while useful, it was kind of jarring because of the sudden logic jump.  What do you think about adding another createScenario pick in this case?

src/commands/createRepo/WorkspaceListStep.ts
```diff
@@ -0,0 +1,52 @@
...
+
+    public shouldPrompt(_wizardContext: IStaticWebAppWizardContext): boolean {
+        // regardless of prompting or not, this has to run to get the workspace/repo
+        return true;
+    }
+
+    public async getSubWizard(_wizardContext: IStaticWebAppWizardContext): Promise<IWizardOptions<IStaticWebAppWizardContext> | undefined> {
+        // some shouldPrompts depend on having a workspace, so put all of the steps as a subwizard instead
+        const promptSteps: AzureWizardPromptStep<IStaticWebAppWizardContext>[] = [];
+        promptSteps.push(new GitHubOrgListStep(), new RepoNameStep(), new RepoPrivacyStep(), new RemoteShortnameStep(), new GitignoreListStep());
```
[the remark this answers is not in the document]
nturinski: We can, but the step numbers will change after selecting a workspace.  Not sure if that's an issue or not.

src/commands/createRepo/RepoCreateStep.ts
```diff
@@ -22,29 +24,46 @@ export class RepoCreateStep extends AzureWizardExecuteStep<IStaticWebAppWizardCo
...
         await repo.addRemote(remoteName, gitHubRepoRes.clone_url);
         const branch: Branch = await repo.getBranch('HEAD');
+
+        const pushingBranch: string = localize('pushingBranch', 'Pushing local branch "{0}" to GitHub repository "{1}"...', branch.name, wizardContext.newRepoName);
+        ext.outputChannel.appendLog(pushingBranch);
+        progress.report({ message: pushingBranch });
         await repo.push(remoteName, branch.name, true);
 
-        const createdGitHubRepo: string = localize('createdGitHubRepo', 'Created new {0} GitHub repository "{1}"', privacy, newRepoName);
+        const createdGitHubRepo: string = newRepoIsPrivate ? localize('createdPrivateGitHubRepo', 'Created new private GitHub repository "{0}".', newRepoName) :
```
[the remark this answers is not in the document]
nturinski: Hmm, well I don't really think the repo is considered "created" (at least from our extension's perspective-- an empty repo will exist) until after the push completes.  Think we should just say push completed then?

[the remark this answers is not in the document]
ejizba: What do you mean it's not considered "created"? Just because a resource group is empty, doesn't mean we wait to tell users we created it until we fill it with resources

src/commands/createRepo/WorkspaceListStep.ts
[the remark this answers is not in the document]
ejizba: Yeah I would prefer that, and I think we can modify `shouldPrompt` in a few steps to fix any super weird counts

nturinski:
> What are the main differences between basic/advanced without auto-selecting the remote?

So with basic, we assume if it's one workspace, that is the workspace that they want to use.  If we can find the remote in their GitHub account, then we automatically set the org/repo to that as well.  We also assume the branch to just be the default branch.  We automatically make a rg for them as well.

But I think since it's not abundantly clear that SWA/GitHub are so intrinsically tied together, that the logic jump of "oh my current workspace is a Github repo that I own" is kind of hard to piece together so I thought physically showing them the remote repo and letting them just select it as a "How do I want to create" prompt could more easily fit those pieces together.

src/commands/createRepo/RepoCreateStep.ts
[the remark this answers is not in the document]
nturinski: Yeah, I see what you mean. I just added a log signifying when the push is complete as well.

src/commands/createStaticWebApp/DeploymentMethodStep.ts
nturinski: Will start working on this component now (I didn't forget 😁)

ejizba: Yeah I mean the advantage of an aka.ms link is that you can change it after the PR is merged (but that doesn't mean put it off forever)

src/commands/createStaticWebApp/CreateMethodStep.ts
```diff
@@ -0,0 +1,75 @@
...
+import { RepoPrivacyStep } from "../createRepo/RepoPrivacyStep";
+import { WorkspaceListStep } from "../createRepo/WorkspaceListStep";
+import { GitHubBranchListStep } from "./GitHubBranchListStep";
+import { GitHubOrgListStep } from "./GitHubOrgListStep";
+import { GitHubRepoListStep } from "./GitHubRepoListStep";
+import { IStaticWebAppWizardContext } from "./IStaticWebAppWizardContext";
+
+export type CreateScenario = 'publishToNewRepo' | 'connectToExistingRepo';
+
+export class CreateMethodStep extends AzureWizardPromptStep<IStaticWebAppWizardContext> {
```
ejizba, line 25: CreateScenarioStep? or CreateScenarioListStep?

ejizba (approved):
Didn't get a chance to go much into the details for the last few set of changes, but the general idea looks good to me - definitely good enough to merge and start getting it tested

src/commands/createStaticWebApp/DeploymentMethodStep.ts
nturinski: https://github.com/microsoft/vscode-azurestaticwebapps/wiki/Creating-a-Static-Web-App

Issue #239, opened 2020-11-04
"Application Settings" tree item should be moved under "Functions"


wwlorey:
I thought that since my app had deployed, I shouldn't be seeing an error here:
![image](https://user-images.githubusercontent.com/22795803/98161996-82e5f100-1e95-11eb-9c32-2ff7a98a4321.png)

As discussed offline - since that error is related to functions, it may be best to move the whole "Application Settings" tree item under "Functions"

PR #247, opened 2020-11-09
Move app settings tree node under functions

nturinski:
Fixes https://github.com/microsoft/vscode-azurestaticwebapps/issues/239

You can't have app settings without a function app, so it makes sense to hide the node unless you have deployed a function project.  The default error when expanding the app settings node from Azure isn't super clear either, especially if you aren't familiar with what a function is.

I thought about creating the tree to look like this:

>Functions
>-> Application Settings
> ---> test = value
>-> Triggers
> ---> HttpTrigger1

But it ultimately felt like overkill to have another level of nodes to expand.  I like how it looks now, but if anyone has dissenting opinions, am always open to feedback!

Old:
image

New: 
image

### DIFF
--- a/src/tree/EnvironmentTreeItem.ts
+++ b/src/tree/EnvironmentTreeItem.ts
@@ -91,19 +91,20 @@ export class EnvironmentTreeItem extends AzureParentTreeItem implements IAzureRe
         return treeUtils.getIconPath('Azure-Static-Apps-Environment');
     }
 
-    public async loadMoreChildrenImpl(_clearCache: boolean, _context: IActionContext): Promise<AzExtTreeItem[]> {
+    public async loadMoreChildrenImpl(_clearCache: boolean, context: IActionContext): Promise<AzExtTreeItem[]> {
         const client: WebSiteManagementClient = createAzureClient(this.root, WebSiteManagementClient);
         const functions: WebSiteManagementModels.StaticSiteFunctionOverviewCollection = await client.staticSites.listStaticSiteBuildFunctions(this.parent.resourceGroup, this.parent.name, this.buildId);
-        client.staticSites.listStatuc;
         if (functions.length === 0) {
+            context.telemetry.properties.hasFunctions = 'false';
             return [this.actionsTreeItem, new GenericTreeItem(this, {
-                label: localize('noFunctions', 'Learn more about Functions in Azure Static Web Apps...'),
+                label: localize('noFunctions', 'Learn how to add an API with Azure Functions...'),
                 contextValue: 'noFunctions',
                 commandId: 'staticWebApps.showFunctionsDocumentation',
-                iconPath: new ThemeIcon('question')
+                iconPath: new ThemeIcon('book')
             })];
         }
 
+        context.telemetry.properties.hasFunctions = 'true';
         return [this.actionsTreeItem, this.appSettingsTreeItem, this.functionsTreeItem];
     }
 
--- a/src/tree/EnvironmentTreeItem.ts
+++ b/src/tree/EnvironmentTreeItem.ts
@@ -4,9 +4,9 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { WebSiteManagementClient, WebSiteManagementModels } from "@azure/arm-appservice";
-import { ProgressLocation, window } from "vscode";
+import { ProgressLocation, ThemeIcon, window } from "vscode";
 import { AppSettingsTreeItem, AppSettingTreeItem } from "vscode-azureappservice";
-import { AzExtParentTreeItem, AzExtTreeItem, AzureParentTreeItem, createAzureClient, IActionContext, TreeItemIconPath } from "vscode-azureextensionui";
+import { AzExtTreeItem, AzureParentTreeItem, createAzureClient, GenericTreeItem, IActionContext, TreeItemIconPath } from "vscode-azureextensionui";
 import { AppSettingsClient } from "../commands/appSettings/AppSettingsClient";
 import { productionEnvironmentName } from "../constants";
 import { ext } from "../extensionVariables";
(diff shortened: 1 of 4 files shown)


src/tree/EnvironmentTreeItem.ts
```diff
@@ -92,7 +89,7 @@ export class EnvironmentTreeItem extends AzureParentTreeItem implements IAzureRe
     }
 
     public async loadMoreChildrenImpl(_clearCache: boolean, _context: IActionContext): Promise<AzExtParentTreeItem[]> {
-        return [this.actionsTreeItem, this.appSettingsTreeItem, this.functionsTreeItem];
+        return [this.actionsTreeItem, this.functionsTreeItem];
```
EricJizbaMSFT, line 92: I think you're confusing "function"/"function app"/"trigger". If you were going to do the "overkill" tree structure, this is how I would word it:

> Function App
> -> Application Settings
> ---> test = value
> -> Functions
> ---> HttpTrigger1

I would prefer the existing UI or the slightly "overkill" UI over the proposed changes in this PR, which I think is even more confusing.

Another option (not sure if this is possible) would be to detect if the app has any functions deployed to it ahead of time. If not, hide both "functions" and "application settings" and instead show a generic node like "Add a function..."/"Deploy a function app..."/"Learn how to add functions..."

nturinski: It's definitely possible, it'll just take be slightly longer to load the build's children (from instant to an Azure API get call).

As far as the wording, I think the portal purposely avoids using the word Function App because you don't have full control over it like you would if you were actually using a function app proper.

image

I think since the API call is on the build itself (and not when you're loading the children of the subscription) I'm okay with the latency there.

nturinski: I played with the idea of putting a tree item for creating a function if the user didn't have one, but I didn't really like how it looked.

Also, we can't actually _create_ a function app for them.  Even if they do have their project opened in the local workspace, we'd still have to commit/push the repo for them.  I guess we could take them to this page though https://docs.microsoft.com/en-us/azure/static-web-apps/apis

image

EricJizbaMSFT: > it'll just take be slightly longer to load the build's children (from instant to an Azure API get call).

This would be my biggest concern as well and it definitely needs to be considered. As you mentioned, though, it's probably okay assuming it only happens for one build at a time when you expand it.

>  but I didn't really like how it looked.

I think it looks weird because there's no icon - just add an icon.

> Also, we can't actually create a function app for them. Even if they do have their project opened in the local workspace, we'd still have to commit/push the repo for them. I guess we could take them to this page though https://docs.microsoft.com/en-us/azure/static-web-apps/apis

I'm fine with just a learn more link if that's easier

> Oh yeah, but they also don't nest Application Settings in Functions either, but they also have a terrible error message that's super hard to notice as well, not really sure it's a great pattern to follow

First of all - not sure why you think that error message is so terrible... Secondly, nesting app settings under functions is the only option I don't like. It's just not consistent with other UI (in the portal or the functions extension) and the labeling doesn't makes sense to me. If you don't want to do any of the other options, I'm totally fine keeping the UI as-is (with app settings and functions at the same level).

nturinski: Well, it's more that the message is hard to see (it's just a small ribbon at the top), but all the UI still functions so it's hard to notice that it's not actually doing anything.  It's not consistent with the other UI, but also Functions/App Service app settings affect the entire app whereas App Settings only applies to the Function app that lives in the SWA.  The portal actually calls it "Configuration" which I don't think is specific enough to help the user, but what if we called it Function Settings instead then?

EricJizbaMSFT: I'm not a big fan of inventing new terminology for platform features. The portal might use "Configuration", but [plenty of docs](https://docs.microsoft.com/azure/static-web-apps/application-settings) and our other extensions use "Application Settings"

nturinski: Okay, then I'll leave everything the same but just have a learn more about Azure Functions in Static Web Apps... node or something that directs them the doc you just linked.  Seem 👌?

nturinski:
New look: 
image

Clicking on it leads [here](https://aka.ms/AAacf3z)

src/tree/EnvironmentTreeItem.ts
```diff
@@ -91,7 +91,18 @@ export class EnvironmentTreeItem extends AzureParentTreeItem implements IAzureRe
         return treeUtils.getIconPath('Azure-Static-Apps-Environment');
     }
 
-    public async loadMoreChildrenImpl(_clearCache: boolean, _context: IActionContext): Promise<AzExtParentTreeItem[]> {
+    public async loadMoreChildrenImpl(_clearCache: boolean, _context: IActionContext): Promise<AzExtTreeItem[]> {
+        const client: WebSiteManagementClient = createAzureClient(this.root, WebSiteManagementClient);
+        const functions: WebSiteManagementModels.StaticSiteFunctionOverviewCollection = await client.staticSites.listStaticSiteBuildFunctions(this.parent.resourceGroup, this.parent.name, this.buildId);
```
EricJizbaMSFT, line 96: Hmm there's not a simpler API we can call to detect if they have functions set up? For example, technically I could have a function app, but just no functions. Would be nice if there's just like a property on the site config or something

nturinski: As far as I know, SWA only creates a function app if there are functions in the designated API folder.  If you don't have functions, it'll give you the error that a function must be first deployed.  Even if you previously had a function and then remove the api folder, it'll go back to throwing that error.

I checked the properties on the build itself, but don't see anything that we could use for that.

nturinski: I could use the list function app settings API call if you think that'd be faster.  It'll throw an error if it can't detect a function app, so in case the scenario you described is possible, it should be a more reliable indicator of whether or not there's a function app.

src/tree/EnvironmentTreeItem.ts
```diff
@@ -91,7 +91,19 @@ export class EnvironmentTreeItem extends AzureParentTreeItem implements IAzureRe
...
+        client.staticSites.listStatuc;
```
EricJizbaMSFT, line 97: lol

src/tree/EnvironmentTreeItem.ts
EricJizbaMSFT: 🤷‍♂️

src/tree/EnvironmentTreeItem.ts
```diff
@@ -91,7 +91,19 @@ export class EnvironmentTreeItem extends AzureParentTreeItem implements IAzureRe
...
+        if (functions.length === 0) {
```
EricJizbaMSFT, line 98: I'd add telemetry to track how many people do/don't have functions

src/tree/EnvironmentTreeItem.ts
```diff
@@ -91,7 +91,19 @@ export class EnvironmentTreeItem extends AzureParentTreeItem implements IAzureRe
...
+            return [this.actionsTreeItem, new GenericTreeItem(this, {
+                label: localize('noFunctions', 'Learn more about Functions in Azure Static Web Apps...'),
```
EricJizbaMSFT, line 100: How about this (modeled after the title of the article you link to):

> Learn how to add an API with Azure Functions...

src/tree/EnvironmentTreeItem.ts
```diff
@@ -91,7 +91,19 @@ export class EnvironmentTreeItem extends AzureParentTreeItem implements IAzureRe
...
+                contextValue: 'noFunctions',
+                commandId: 'staticWebApps.showFunctionsDocumentation',
+                iconPath: new ThemeIcon('question')
```
EricJizbaMSFT, line 103: Other places use the little book to indicate documentation

Screen Shot 2020-11-18 at 11 21 49 AM

EricJizbaMSFT (approved):

src/tree/EnvironmentTreeItem.ts
nturinski: lol it was late 😭

PR #251, opened 2020-11-10
Add git to getExtensionApi file

nturinski:
Refactored from: https://github.com/microsoft/vscode-azurestaticwebapps/pull/246

I need the git api for some other things and thought it would just make code review a little simpler.  While I'm not using the gitApi in the codebase yet, another PR will follow shortly (plus there's #246 if you need an example of how it'll be used)

### DIFF
--- a/src/getExtensionApi.ts
+++ b/src/getExtensionApi.ts
@@ -7,6 +7,7 @@ import { commands, Extension, extensions } from "vscode";
 import { IActionContext, UserCancelledError } from "vscode-azureextensionui";
 import { AzureExtensionApiProvider } from "vscode-azureextensionui/api";
 import { ext } from "./extensionVariables";
+import { API, GitExtension } from "./git";
 import { localize } from "./utils/localize";
 import { AzureFunctionsExtensionApi } from "./vscode-azurefunctions.api";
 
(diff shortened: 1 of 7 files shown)


src/git.d.ts
```diff
@@ -0,0 +1,304 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { Disposable, Event, ProviderResult, Uri } from 'vscode';
+export { ProviderResult } from 'vscode';
+
```
EricJizbaMSFT, line 8: Can you add a comment in this file linking to the source code for this

src/getExtensionApi.ts
```diff
@@ -30,3 +31,15 @@ export async function getFunctionsApi(context: IActionContext): Promise<AzureFun
     // we still need to throw an error even if the user installs
     throw new UserCancelledError();
 }
+
+export async function getGitApi(): Promise<API> {
+    try {
+        const gitExtension: Extension<GitExtension> | undefined = extensions.getExtension<GitExtension>('vscode.git');
+        if (gitExtension) {
```
EricJizbaMSFT, line 38: Should probably do the same 'isActive' check done in `getFunctionsApi` (even if the git extension almost always activates before us - better to be safe). Might even be worth it to add a helper function that both `getGitApi` and `getFunctionsApi` since it seems like there's a bit of duplicate stuff

EricJizbaMSFT (approved):

Issue #197, opened 2020-09-08
After Revoking GitHub access no longer useable. Need to change GitHub Org.


LarouexSoftwareDesign:
- Configuring for the first time, opened Edge, signed into of my GitHub orgs.
- Authorized.
- Realized it was not the Org I wanted.
- Un-authorized the VS-Code App.
- Uninstalled extension and re-installed.
- Cannot get beyond "Bad Credentials" error in extension.

I checked extension settings and it was not obvious how to fix and uninstall does not get me to re-enter creds.

nturinski:
Hi @LarouexSoftwareDesign 

Sorry you're having issues.  How did you unauthorize VS Code? via Github?  You'll have to sign-out of GitHub from VS Code as well.  You can do that by clicking this profile button at the bottom left:
![image](https://user-images.githubusercontent.com/5290572/92518171-c93c0e00-f1cc-11ea-94b0-9b08137d42ac.png)
![image](https://user-images.githubusercontent.com/5290572/92518194-d0fbb280-f1cc-11ea-8d09-bb817c0ffca0.png)

I agree that it's not super obvious how to fix. Hope this resolves your problem.

nturinski:
Will direct users to sign out of GitHub command if we encounter this error.

nturinski:
Moving this to next milestone until logout command is added back to VS Code API.

nturinski:
Tracking here: https://github.com/microsoft/vscode/issues/115761

nturinski:
Logout seems unlikely, so we can just direct users to reauthorize their GitHub credentials for now.

nturinski:
Looking into this further, I don't think the tradeoff for directing users to reauthorize their GitHub credentials or logging out is worth the cost.  The reasons being:

1. In order to detect stale credentials, we would have to check VS Code's access token against the GitHub API every time we retrieve it.  This could be somewhat expensive for all users with what seems to be a pretty edge case.
2. This only seems to happen if you go to GitHub's settings and unauthorize GitHub for VSCode ![image](https://user-images.githubusercontent.com/5290572/117368484-72c8c180-ae78-11eb-822e-8e8adbabdd65.png)
3. This is resolved if you restart VS Code.  On the next attempt, VS Code will verify the token, see that it's invalid, and prompt for the user to login again.

PR #252, opened 2020-11-13
Check for bad credentials/ sign-out automatically

nturinski:
Fixes https://github.com/microsoft/vscode-azurestaticwebapps/issues/197

Doesn't seem like this happens frequently, but basically if you revoke your token access in GitHub Developer's Settings while you have VS Code opened, VS Code won't recognize that the token is invalid now so any action throws "Bad credentials".  This is resolved by either logging out of session or by restarting VS Code, but it's kind of hard to discover that.

I only validate when no token is passed in since it'd be annoying to add an extra API call for each step of the wizard that uses Octokit client.

### DIFF
--- a/src/commands/github/createOctokitClient.ts
+++ b/src/commands/github/createOctokitClient.ts
@@ -4,15 +4,34 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { Octokit } from "@octokit/rest";
-import { appendExtensionUserAgent } from "vscode-azureextensionui";
-import { getGitHubAccessToken } from "../../utils/gitHubUtils";
+import { appendExtensionUserAgent, parseError } from "vscode-azureextensionui";
+import { getGitHubAccessToken, logoutOfGitHubSession } from "../../utils/gitHubUtils";
 
 // token should only be passed in during a wizard; otherwise retrieve a new token for the request
 export async function createOctokitClient(token?: string): Promise<Octokit> {
+    const validateCredentials: boolean = !token;
     token = token || await getGitHubAccessToken();
-    return new Octokit(
+    const client: Octokit = new Octokit(
         {
             userAgent: appendExtensionUserAgent(),
             auth: token
         });
+
+    if (validateCredentials) {
+        try {
+            await client.users.getAuthenticated();
+        } catch (error) {
+            if (parseError(error).message.includes('Bad credentials')) {
+                await logoutOfGitHubSession();
+                return new Octokit(
+                    {
+                        userAgent: appendExtensionUserAgent(),
+                        auth: await getGitHubAccessToken()
+                    });
+            }
+
+            throw error;
+        }
+    }
+    return client;
 }
--- a/src/utils/gitHubUtils.ts
+++ b/src/utils/gitHubUtils.ts
@@ -103,9 +103,19 @@ export async function getGitHubQuickPicksWithLoadMore<TResult, TParams>(
     }
 }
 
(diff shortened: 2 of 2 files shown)


src/utils/gitHubUtils.ts
```diff
@@ -103,9 +103,19 @@ export async function getGitHubQuickPicksWithLoadMore<TResult, TParams>(
...
-    const scopes: string[] = ['repo', 'workflow', 'admin:public_key'];
-    return (await authentication.getSession('github', scopes, { createIfNone: true })).accessToken;
+    return (await authentication.getSession(githubProviderId, scopes, { createIfNone: true })).accessToken;
+}
+
+export async function logoutOfGitHubSession(): Promise<void> {
+    const sessionId: string | undefined = (await authentication.getSession(githubProviderId, scopes))?.id;
+
+    if (sessionId) {
+        await authentication.logout(githubProviderId, sessionId);
```
EricJizbaMSFT, line 117: I don't see logout in the official API and I thought we were supposed to have moved off of the proposed API a long time ago

src/commands/github/createOctokitClient.ts
```diff
@@ -4,15 +4,34 @@
...
         {
             userAgent: appendExtensionUserAgent(),
             auth: token
         });
+
+    if (validateCredentials) {
+        try {
+            await client.users.getAuthenticated();
+        } catch (error) {
+            if (parseError(error).message.includes('Bad credentials')) {
```
EricJizbaMSFT, line 24: This doesn't give you a better error code like '401'?

src/commands/github/createOctokitClient.ts
```diff
@@ -4,15 +4,34 @@
...
+                await logoutOfGitHubSession();
```
EricJizbaMSFT, line 25: In the issue you said:

> Will direct users to sign out of GitHub command if we encounter this error.

I like that better than logging out _for_ them. It just seems a bit too magic-y, plus the account is managed at the VS Code-level, not just for our extension.

nturinski: Hmm, how should I direct them?  It feels weird explaining it in a notification that the user profiles is in the bottom right corner of their screen.

src/utils/gitHubUtils.ts
nturinski: Yeah, you're right.  It's actually deprecated in the proposed api... 😬

src/commands/github/createOctokitClient.ts
nturinski: I would like to have a button to sign them out, but I can't find the commandId to execute to initiate the sign out process.  I've looked through their repo and see the actual function, but I don't see anywhere it gets registered and used.

EricJizbaMSFT: Maybe VS Code just doesn't want an extension like us signing users out of something affecting VS Code as a whole. If that's the case, I say we just won't-fix the bug and move on. Like... I think it's fair to expect a "Bad Credentials" error if you revoke access to a key. All you gotta do is sign out and sign back in

nturinski: According to Rachel, other extensions have requested to bring back logout so I'll close this PR.  As for the bug, I do feel like it's a very rare error and the fix is pretty easy, the only issue is it's hard to figure out how to sign out/in.  I think restarting VS Code resolves this problem now though too, so I'm leaning towards just closing it as wont-fix.

Issue #237, opened 2020-11-04
Canceling out of GitHub sign-in throws an error


wwlorey:
![image](https://user-images.githubusercontent.com/22795803/98161512-b70ce200-1e94-11eb-967d-bbcf569b14be.png)

PR #253, opened 2020-11-13
Swallow VS Code 'User did not consent' error and throw UserCancelledError

nturinski:
Fixes https://github.com/microsoft/vscode-azurestaticwebapps/issues/237

### DIFF
--- a/src/utils/gitHubUtils.ts
+++ b/src/utils/gitHubUtils.ts
@@ -10,7 +10,7 @@ import gitUrlParse = require('git-url-parse');
 import * as git from 'simple-git/promise';
 import { URL } from 'url';
 import { authentication, QuickPickItem } from 'vscode';
-import { IAzureQuickPickItem } from 'vscode-azureextensionui';
+import { IActionContext, IAzureQuickPickItem, parseError, UserCancelledError } from 'vscode-azureextensionui';
 import { IStaticWebAppWizardContext } from '../commands/createStaticWebApp/IStaticWebAppWizardContext';
 import { createOctokitClient } from '../commands/github/createOctokitClient';
 import { GitTreeData, OrgForAuthenticatedUserData } from '../gitHubTypings';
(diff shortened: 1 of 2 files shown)


src/utils/gitHubUtils.ts
```diff
@@ -103,9 +103,20 @@ export async function getGitHubQuickPicksWithLoadMore<TResult, TParams>(
...
 }
 
-export async function getGitHubAccessToken(): Promise<string> {
+export async function getGitHubAccessToken(context?: IActionContext): Promise<string> {
     const scopes: string[] = ['repo', 'workflow', 'admin:public_key'];
-    return (await authentication.getSession('github', scopes, { createIfNone: true })).accessToken;
+    try {
+        return (await authentication.getSession('github', scopes, { createIfNone: true })).accessToken;
+    } catch (error) {
+        if (parseError(error).message === 'User did not consent to login.') {
```
nturinski, line 111: There's no error code or anything that I can refer to:
image

But I don't think this solution will work for other languages since it should be localized...

src/utils/gitHubUtils.ts
```diff
@@ -103,9 +103,20 @@ export async function getGitHubQuickPicksWithLoadMore<TResult, TParams>(
     }
 }
 
-export async function getGitHubAccessToken(): Promise<string> {
+export async function getGitHubAccessToken(context?: IActionContext): Promise<string> {
```
EricJizbaMSFT, line 106: Why is context optional?

src/utils/gitHubUtils.ts
EricJizbaMSFT: It's not just about other languages - the english text can easily change and break this logic. Ideally you would file an issue on VS Code to give us a better way to detect this error and add a link in the code to that issue. That being said, in this case I don't care that much since the downside isn't that bad (aka they just see the "consent" message)

src/utils/gitHubUtils.ts
nturinski: Because it's kind of a pain to make it required since it's used in a lot of code, some of which does not context.  But most of those are one-off calls (like getting the Actions).  In the case the context _is_ passed in is during the create static web app wizard, where I thought we'd be more interested in the cancel step telemetry.

src/utils/gitHubUtils.ts
nturinski: Filed here https://github.com/microsoft/vscode/issues/110758

EricJizbaMSFT (approved):

src/utils/gitHubUtils.ts
nturinski: Will be handled by _someone_ _after_ release

Issue #249, opened 2020-11-10
[Suggestion] Make it known that you have to wait for GH Action to finish before accessing app


wwlorey:
From https://github.com/MicrosoftDocs/azure-docs/issues/65544#issuecomment-724732460

> Ultimately, when you first create the app it has to build and the banner shown in the portal does a good job of notifying developers that they need to wait for the Action to finish before the app is available.

PR #255, opened 2020-11-13
Change msg when creating swa and reveal actions node

nturinski:
Fixes https://github.com/microsoft/vscode-azurestaticwebapps/issues/249

Based off the feedback, I'd like to make it painfully obvious that the action needs to finish before you can access the SWA.  @fiveisprime you might have to help me with the wording of the message a bit though.

I thought it'd help to just reveal the tree item to the user as well so they can see that the action is running.

### DIFF
--- a/src/commands/createStaticWebApp/createStaticWebApp.ts
+++ b/src/commands/createStaticWebApp/createStaticWebApp.ts
@@ -20,7 +20,7 @@ export async function createStaticWebApp(context: IActionContext & Partial<ICrea
 
     const swaNode: StaticWebAppTreeItem = await node.createChild(context);
 
-    const createdSs: string = localize('createdSs', 'Successfully created new static web app "{0}".  The GitHub Action must complete before your Static Web App will receive the deployed content.', swaNode.name);
+    const createdSs: string = localize('createdSs', 'Successfully created new static web app "{0}".  GitHub Actions is building and deploying your app, it will be available once the deployment completes.', swaNode.name);
     ext.outputChannel.appendLog(createdSs);
 
     const showActionsMsg: MessageItem = { title: localize('openActions', 'Open Actions in GitHub') };
--- a/src/commands/createStaticWebApp/createStaticWebApp.ts
+++ b/src/commands/createStaticWebApp/createStaticWebApp.ts
@@ -6,9 +6,11 @@
 import { MessageItem, window } from 'vscode';
 import { IActionContext, ICreateChildImplContext } from 'vscode-azureextensionui';
 import { ext } from '../../extensionVariables';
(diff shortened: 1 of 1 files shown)


src/commands/createStaticWebApp/createStaticWebApp.ts
```diff
@@ -18,7 +20,7 @@ export async function createStaticWebApp(context: IActionContext & Partial<ICrea
 
     const swaNode: StaticWebAppTreeItem = await node.createChild(context);
 
-    const createdSs: string = localize('createdSs', 'Successfully created new static web app "{0}".', swaNode.name);
+    const createdSs: string = localize('createdSs', 'Successfully created new static web app "{0}".  The GitHub Action must complete before your Static Web App will receive the deployed content.', swaNode.name);
```
fiveisprime, line 23: "GitHub Actions is buildig and deploying your app, it will be available once the deployment completes."

nturinski:
Once [this](https://github.com/microsoft/vscode-azurestaticwebapps/pull/256) is merged, I can have logic that tells the user exactly when that build has completed.

src/commands/createStaticWebApp/createStaticWebApp.ts
```diff
@@ -31,6 +33,8 @@ export async function createStaticWebApp(context: IActionContext & Partial<ICrea
             ext.outputChannel.show();
         }
     });
+    const children: EnvironmentTreeItem = <EnvironmentTreeItem>(await swaNode.loadAllChildren(context))[0]; // since it's newly created, should only have the environment tree node
```
EricJizbaMSFT, line 36: No need to assume and risk a bug when you could do a quick `find` to make sure

src/commands/createStaticWebApp/createStaticWebApp.ts
```diff
@@ -31,6 +33,8 @@ export async function createStaticWebApp(context: IActionContext & Partial<ICrea
...
+    await revealTreeItem(children.actionsTreeItem.fullId);
```
EricJizbaMSFT, line 37: I like the message change a lot - not sure about `revealTreeItem`, though. I think we generally try to avoid changing the focus in VS Code because that can be annoying. For example:
- If I started this from the command palette, I probably don't want to see the tree
- If I started working on my local app while the remote app was creating, I probably would get annoyed if you interrupted me to show the tree

nturinski: I think I'm leaning to getting rid of the `revealTreeItem` change.  Since I should be able to detect when the build is done, the line between the two products is clearer to draw for the user.

Before when I added this, I really wanted users to see that their build was in-progress so they could assume that this needed to finish before they could proceed.

EricJizbaMSFT (approved):

Issue #172, opened 2020-07-30
Change the icon of the workflow and add the completed log after finishing "Rerun Action..."


v-ruizh:
**OS:** All
**Build Version:** 20200730.10

**Repro steps:**
1. Expand "Actions" node --> Right click one completed workflow --> Select "Rerun Action...".
2. Check the workflow state and output logs.

**Expect:**
The icon of this workflow is changed to completed state and there is a completed log.

**Actual:**
The icon of this workflow is still running state and there is only one log "Rerun for... has started".
![image](https://user-images.githubusercontent.com/45614674/88891296-81dd6c00-d275-11ea-89a8-b11785d3642d.png)

**More Info:**
1. This issue also reproduces for "Cancel Action...".
2. The icon of the workflow will be changed after refreshing.

PR #256, opened 2020-11-14
Use retry logic to check status and refresh when workflow completes

nturinski:
Fixes https://github.com/microsoft/vscode-azurestaticwebapps/issues/172
Fixes https://github.com/microsoft/vscode-azurestaticwebapps/issues/262

I think that this solution mitigates one of our concerns about polling before which was refreshing the tree or tree node was a waste of resources and also distracting.  Also because the frequency of polling gets incrementally longer, if it's a long-running build, it shouldn't impact the user too badly.

Plus I don't see people rerunning too many actions at once anyway.

### DIFF
--- a/src/commands/github/actionCommands.ts
+++ b/src/commands/github/actionCommands.ts
@@ -5,16 +5,14 @@
 
 import { Octokit } from "@octokit/rest";
 import { ActionsGetWorkflowRunResponseData } from "@octokit/types";
-import * as retry from 'p-retry';
-import { CancellationToken, CancellationTokenSource, window } from "vscode";
-import { IActionContext, parseError, UserCancelledError } from "vscode-azureextensionui";
+import { window } from "vscode";
+import { IActionContext } from "vscode-azureextensionui";
 import { ext } from "../../extensionVariables";
 import { ActionTreeItem } from "../../tree/ActionTreeItem";
 import { ensureStatus } from "../../utils/actionUtils";
+import { pollAsyncOperation } from "../../utils/azureUtils";
 import { localize } from "../../utils/localize";
 import { createOctokitClient } from "./createOctokitClient";
-
-const activeActionTokens: { [key: string]: CancellationTokenSource | undefined } = {};
 
 export async function rerunAction(context: IActionContext, node?: ActionTreeItem): Promise<void> {
     if (!node) {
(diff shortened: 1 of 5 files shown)


src/commands/github/actionCommands.ts
```diff
@@ -34,4 +40,62 @@ export async function cancelAction(context: IActionContext, node?: ActionTreeIte
...
+            {
+                onFailedAttempt: error => {
+                    if (parseError(error).message !== notCompleteError) {
+                        throw error;
+                    }
+                },
+                retries, minTimeout: 5 * 1000,
+            });
+    } catch (error) {
+        if (parseError(error).message === cancelled) {
```
nturinski, line 91: For some reason, the abort error doesn't keep its typing in the catch block, so I have to check the message instead (which is never displayed to the user)

src/commands/github/actionCommands.ts
```diff
@@ -34,4 +41,62 @@ export async function cancelAction(context: IActionContext, node?: ActionTreeIte
...
+    // realistically, we have no idea how long someone's build can take, but might as well keep retrying while VS Code is opened
+    const tokenSource: CancellationTokenSource = new CancellationTokenSource();
+    const token: CancellationToken = tokenSource.token;
+    if (activeActionTokens[node.data.id]) {
+        activeActionTokens[node.data.id]?.cancel();
+    }
+
+    activeActionTokens[node.data.id] = tokenSource;
+
+    const retries: number = 10;
```
EricJizbaMSFT, line 57: I don't think I would use exponential backoff for something like this. I mostly use it to retry an action itself, not to get the status of that action. If you're retrying the action itself, you need the backoff otherwise you might exacerbate the problem on the server. We've talked about polling in the past and how I don't like auto-refresh in the tree, but this is the exact scenario where I think polling is fine (as long as you're careful). The polling is directly related to an action the user just took, and it's limited to a pretty small scope. This is very similar to other scenarios like creating an azure resource or deploying (except you have to manually do the polling instead of letting the sdk do it for us).

You might want to see if you can combine this with [pollAzureAsyncOperation](https://github.com/microsoft/vscode-azurestaticwebapps/blob/main/src/utils/azureUtils.ts#L35) already in this repo.

nturinski: Hmm, I think the only caveat here though is that a build could potentially take hours.  While polling isn't that expensive of an operation, I feel like doing it every 10 seconds for 4 hours is kind of excessive.  Maybe I can build my own back-off where if it's past 10 minutes, it goes to polling every minute or something?  My expectation is most people will just leave it running and miss the notification, but expect it to be updated in the tree or they'll just close VS Code and see it's complete later.  What do you think?

EricJizbaMSFT: Yeah part of "as long as you're careful" is you have to pick a reasonable interval and timeout. Plan around what you actually expect to happen in the majority of cases, not the 4-hour edge-case. Ideally this involves running a bunch of static web app builds and seeing how long they take to pick your number. I would guess around 10 minutes would be plenty, but maaaybe up to 20? After that, I would just give up and say something like "The build is still running, check &lt;link to github&gt; for status".

nturinski: I figured I'd start at 20 and look at telemetry to see if we need to adjust the time (either up or down).  Based off telemetry, not that many people really use re-run action that anyway.

src/commands/github/actionCommands.ts
```diff
@@ -34,4 +39,30 @@ export async function cancelAction(context: IActionContext, node?: ActionTreeIte
...
+            await node.refresh();
+            context.telemetry.properties.secToReport = String((Date.now() - startTime) / 1000);
+            context.telemetry.properties.conclusion = workflowRun.conclusion;
+            return true;
+        }
+
+        return false;
+    };
+
+    if (!await pollAsyncOperation(pollingOperation, 15, 20 * 60, String(node.data.id))) {
```
EricJizbaMSFT, line 63: Use `nonNullProp` instead of `String()`

Or just use `node.fullId` which can't be undefined

EricJizbaMSFT (approved):

src/commands/github/actionCommands.ts
nturinski: The issue is that node.data.id is a number, not a string.

But since it's just for keeping track of cancellation tokens, I think node.fullId will be just fine 👍

Issue #173, opened 2020-07-30
[Suggestion] Remove "Cancel Action..." context menu from the completed workflow


v-ruizh:
**OS:** All
**Build Version:** 20200730.10

**Repro steps:**
1. Expand "Actions" node --> Right click one completed workflow --> Select "Cancel Action...".
2. Check the result.

**Actual:**
There is an error "Cannot cancel a workflow run that is completed".
![image](https://user-images.githubusercontent.com/45614674/88892240-109eb880-d277-11ea-9297-96baf16eb46d.png)

**Suggestion:**
Remove "Cancel Action..." context menu from the completed workflow.

PR #257, opened 2020-11-14
Remove rerun/cancel from actions that are inProgress/completed

nturinski:
Fixes https://github.com/microsoft/vscode-azurestaticwebapps/issues/173

I don't think there's a great way to fix this through the command palette since `pickTreeItemImpl` can't return an array of children (which I could filter).

### DIFF
--- a/src/tree/ActionTreeItem.ts
+++ b/src/tree/ActionTreeItem.ts
@@ -76,4 +76,8 @@ export class ActionTreeItem extends AzureParentTreeItem implements IAzureResourc
         // sort by the jobs that started first
         return ti1.startedDate.getTime() - ti2.startedDate.getTime();
     }
+
+    public isAncestorOfImpl(contextValue: string | RegExp): boolean {
+        return contextValue !== ActionTreeItem.contextValueCompleted && contextValue !== ActionTreeItem.contextValueInProgress;
+    }
 }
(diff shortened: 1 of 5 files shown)


src/tree/ActionTreeItem.ts
```diff
@@ -7,15 +7,15 @@ import { Octokit } from '@octokit/rest';
...
 import { getRepoFullname } from '../utils/gitHubUtils';
 import { ActionsTreeItem } from "./ActionsTreeItem";
 import { IAzureResourceTreeItem } from './IAzureResourceTreeItem';
 import { JobTreeItem } from './JobTreeItem';
 
 export class ActionTreeItem extends AzureParentTreeItem implements IAzureResourceTreeItem {
-    public static contextValue: string = 'azureStaticAction';
-    public readonly contextValue: string = ActionTreeItem.contextValue;
+    public static contextValueCompleted: string = 'azureStaticActionCompleted';
+    public static contextValueInProgress: string = 'azureStaticActionInProgress';
```
EricJizbaMSFT, line 18: > I don't think there's a great way to fix this through the command palette since pickTreeItemImpl can't return an array of children (which I could filter).

I think you need `isAncestorOfImpl` instead of `pickTreeItemImpl`. Something as simple as this:
```typescript
return contextValue !== ActionTreeItem.contextValueCompleted && contextValue !== ActionTreeItem.contextValueInProgress;
```

That basically just tells the tree item picker that this is the last level of the tree you can find those contextValues.

EricJizbaMSFT: While you're fixing the command palette experience, might as well add a better "noItemFoundErrorMessage", too

EricJizbaMSFT (approved):

PR #259, opened 2020-11-17
Changing references to App Artifact to Output Location

nturinski:
As per Mitch's offline email, the yml is now constructed with output_location rather than app_artifact_location so just changing it within the extension to be consistent.  The SDK hasn't updated to reflect this change and they'll keep supporting appArtifactLocation, so we should be good to keep that as-is for now.

### DIFF
--- a/package.json
+++ b/package.json
@@ -416,7 +416,8 @@
                     "staticWebApps.appArtifactSubpath": {
                         "scope": "resource",
                         "type": "string",
-                        "description": "%staticWebApps.appArtifactSubpath%"
+                        "description": "%staticWebApps.appArtifactSubpath%",
+                        "deprecationMessage": "%staticWebApps.appArtifactSubpathDeprecated%"
                     },
                     "staticWebApps.outputLocationSubpath": {
                         "scope": "resource",
@@ -466,7 +467,7 @@
         "webpack-cli": "^3.3.12"
     },
     "dependencies": {
-        "@azure/arm-appservice": "^6.0.0",
+        "@azure/arm-appservice": "^6.1.0",
         "@azure/ms-rest-js": "^2.0.8",
         "@octokit/rest": "^18.0.3",
         "fs-extra": "^8.1.0",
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -416,7 +416,13 @@
                     "staticWebApps.appArtifactSubpath": {
                         "scope": "resource",
                         "type": "string",
-                        "description": "%staticWebApps.appArtifactSubpath%"
+                        "description": "%staticWebApps.appArtifactSubpath%",
+                        "deprecationMessage": "%staticWebApps.appArtifactSubpathDeprecated%"
+                    },
+                    "staticWebApps.outputLocationSubpath": {
```
EricJizbaMSFT, line 422: Seems like your pattern is _not_ to include "Location" in the setting name (even though "app", "api", and "appArtifact" all append "Location" elsewhere)

src/commands/createStaticWebApp/StaticWebAppCreateStep.ts
```diff
@@ -23,7 +23,7 @@ export class StaticWebAppCreateStep extends AzureWizardExecuteStep<IStaticWebApp
             buildProperties: {
                 appLocation: wizardContext.appLocation,
                 apiLocation: wizardContext.apiLocation,
-                appArtifactLocation: wizardContext.appArtifactLocation
+                appArtifactLocation: wizardContext.outputLocation
```
EricJizbaMSFT, line 26: You should add a comment in the code similar to what you put in the PR description

package.json
nturinski: Well, the name of the property is outputLocation.  So it went from app artifact to output location.  I'm curious what the SDK will do because they use the properties appLocation, apiLocation, and appArtifactLocation.... Will it be outputLocationLocation? 🤣

EricJizbaMSFT: The name of the property where? According to what?

> So it went from app artifact to output location.

I thought it went from "app artifact location" to "output location"

EricJizbaMSFT (approved):

Issue #241, opened 2020-11-04
[Suggestion] Add a "how to deploy" notification after creating a SWA


wwlorey:
I wasn't certain that committing & pushing would trigger a new deployment after I created my SWA. Maybe a notification with that info would be helpful

PR #264, opened 2020-11-19
Detect when the build is complete to alert the user

nturinski:
Fixes https://github.com/microsoft/vscode-azurestaticwebapps/issues/241

I'm not sure if the failure logic really works.  While there is a failed state, I don't think I've actually seen it change to that status.

I could add some logic to verify with the latest action-- it's likely if it failed, that the action running should have failed as well but I think there's some trickiness that could introduce.

I debated including the "Commit and push changes the GitHub repository to create a new deployment." message even if it times out. Thoughts?


src/commands/createStaticWebApp/postCreateStaticWebApp.ts
```diff
@@ -0,0 +1,45 @@
...
+            await productionEnv.refresh();
+            return productionEnv.data.status === 'Ready' || productionEnv.data.status === 'Failed';
+        };
+
+        // only output a message if it completed or failed
+        if (await pollAsyncOperation(pollingOperation, 10, 5 * 60, swaNode.fullId)) {
+            const ready: boolean = productionEnv.data.status === 'Ready';
+            const deploymentMsg: string = ready ?
+                localize('deploymentCompleted', 'Successfully built and deployed "{0}". Commit and push changes the GitHub repository to create a new deployment.', swaNode.name) :
+                localize('deploymentFailed', 'Deployment for "{0}" has failed. Commit and push changes the GitHub repository to create a new deployment.', swaNode.name);
```
wwlorey, line 30: Is what happened during the create process a deployment? I thought you had to deploy (by committing & pushing) after creating

wwlorey (commented):
> I debated including the "Commit and push changes the GitHub repository to create a new deployment." message even if it times out.

I like only showing it when we know the creation was successful. I could see it being confusing if the app failed to create but we give users instructions for how to deploy to it

nturinski:
Well, one reason we would want to tell them how to deploy if it failed is so that they know the next steps to fix their broken app.

src/commands/createStaticWebApp/postCreateStaticWebApp.ts
nturinski: I'm not entirely sure what you're asking, but I think the answer is when you create a web app, in the backend, it's waiting on the GitHub actions to see if the initial build deploys successfully or not.  There are several states that we can read, including "WaitingForDeployment", "Ready", and "Failed".  Once the action completes successfully, it changes to Ready.  If it fails or never finishes, it stays on "WaitingForDeployment" for some reason.

We could theoretically check the GitHub action ourselves, however because the action isn't instant, we'd have a hard time locating exactly which action we should be monitoring.  I might look more into this though since it's probably the best indicator of success/failure.

wwlorey:
> Well, one reason we would want to tell them how to deploy if it failed is so that they know the next steps to fix their broken app.

That makes sense. I didn't realize creating a SWA involved a deployment 👍

wwlorey:
> I debated including the "Commit and push changes the GitHub repository to create a new deployment." message even if it times out.

Since this message would be helpful in the case of a failure or a success, I'd vote for showing it even if the deployment fails. Especially if that would allow us to get rid of the failure checks which seem difficult to do

src/commands/createStaticWebApp/postCreateStaticWebApp.ts
```diff
@@ -0,0 +1,72 @@
...
+        const { owner, name } = getRepoFullname(productionEnv.repositoryUrl);
+        let deployActionNode: ActionTreeItem | undefined;
+        const maxTime: number = Date.now() + 30 * 1000; // it can take a second for the action to queue, wait 30 seconds
+
+        while (!deployActionNode) {
+            await productionEnv.actionsTreeItem.refresh();
+            const actions: ActionTreeItem[] = <ActionTreeItem[]>(await productionEnv.actionsTreeItem.loadAllChildren(context));
+            deployActionNode = actions.find(async (ti) => {
+                const workflow: ActionsGetWorkflowResponseData = (await octokitClient.actions.getWorkflow({ owner, repo: name, workflow_id: ti.data.workflow_id })).data;
+                return workflow.path.includes(swaNode.defaultHostname) && ensureStatus(ti.data) === Status.InProgress;
```
wwlorey, line 38: If the deploy fails or succeeds very quickly would that mess up the `Status.InProgress` check? Or is it unlikely that the deploy would show anything but "in progress" for a while?

wwlorey (approved):

nturinski:
I'm about push a new change up that addresses your question.

But the main things are
1. looking at every action is a waste of bandwidth (because some are completed so it's obviously not that one)
2. find does not play with async at all, and since a Promise is an object, it'll return the first element (because when it returns an object, that is considered truthy

So please give that another review! 🙏

src/commands/createStaticWebApp/postCreateStaticWebApp.ts
```diff
@@ -6,67 +6,75 @@
...
+                const promises: Promise<void>[] = filteredTreeItems.map(async ti => {
+                    const workflow: ActionsGetWorkflowResponseData = (await octokitClient.actions.getWorkflow({ owner, repo: name, workflow_id: ti.data.workflow_id })).data;
+                    // example of defaultHostname: 'black-bay-07228711e.azurestaticapps.net'
+                    // example of workflow path: '.github/workflows/azure-static-web-apps-black-bay-07228711e.yml'
+                    // to verify the workflow matches the current default host, we need to remove the 'azurestaticapps.net' portions of the defaultHostname
+                    if (workflow.path.includes(swaNode.defaultHostname.split('.')[0])) {
+                        deployActionNode = ti;
+                    }
+                });
+                await Promise.all(promises);
```
wwlorey, line 48: Could you explain why you await the promises here? I'm not for sure but it looks like `filteredTreeItems.map` will generate an array of all void values since nothing is returned from within the map call

nturinski: This is so the API call `getWorkflow` gets called in parallel.  If I iterate through it in a for loop, it'll be done synchronously (since I'm awaiting each one, one at a time).

The promise isn't returning anything, but it's setting the `deployActionNode` which is what signals the end of the while loop.  What I'm doing here is getting the workflow file associated to the action because it tells me what .yml file it's using to configure, which should match the defaultHostname of the static web app.

wwlorey: That's pretty nifty. I think it might be worth adding a short comment like "Run `getWorkflow` in parallel" to help less advanced typescript peeps like myself understand 😄

wwlorey (approved):

PR #273, opened 2020-12-02
Release changes for v0.5.0

nturinski:


CHANGELOG.md
```diff
@@ -1,5 +1,24 @@
 # Change Log
 
+## 0.5.0 - 2020-12-01
+
+### Added
+
```
wwlorey, line 6: nit: say bye to this blank line 👋

CHANGELOG.md
```diff
@@ -1,5 +1,24 @@
...
+- Publish local projects directly to GitHub
+- Advanced creation to have more control over creation
```
wwlorey, line 8: What about something like this?
```suggestion
- Advanced creation provides more control when creating static web apps
```

CHANGELOG.md
```diff
@@ -1,5 +1,24 @@
...
+- Download remote app settings
+- Notification for when long running GitHub actions complete
+
+### Changed
+- Functions/App Setting tree items no longer appear if build doesn't contain functions
+- `Create New Static Web App` prompts now asks for deployment method
```
wwlorey, line 14: Should "prompts" be singular here?
```suggestion
- `Create New Static Web App` prompt now asks for deployment method
```

Or should "asks" be singular?
```suggestion
- `Create New Static Web App` prompts now ask for deployment method
```

CHANGELOG.md
```diff
@@ -1,5 +1,24 @@
...
+- Completed GitHub actions cannot be cancelled, on-going actions cannot be re-run
```
wwlorey, line 15: Should this go in the "Fixed" section? It seems like its more of a bug 🐛

CHANGELOG.md
nturinski: Prompts should probably be plural in this case, but I think I should probably just use the word "wizard" since that encompasses the entirety of the prompts?

CHANGELOG.md
nturinski: Hmm, I think that this was more of a by-design choice originally.  We wanted to increase discoverability by having the node there, even if it led to an error.  It's more of a change now that we include the learn more node.

CHANGELOG.md
nturinski: Oh I see why you were asking now heh

CHANGELOG.md
nturinski: 👋

CHANGELOG.md
nturinski: I always read the lines with added prefixed in front of it i.e. Added advanced creation provides more control... 

That being said, I think "to provide more control" sounds good to me

wwlorey (approved):

Issue #129, opened 2020-06-18
TSLint has been deprecated in favour of eslint-typescript


indcoder:
On the contribution section in the readme in both the repository and the marketplace, there's a requirement to install TSLint which has has been deprecated by its maintainers at the end of 2019

![image](https://user-images.githubusercontent.com/2099300/85023249-17bfaa80-b192-11ea-82de-3545e1938630.png)

https://github.com/palantir/tslint/issues/4904

https://medium.com/palantir/tslint-in-2019-1a144c2317a9

Perhaps the text need to read eslint-typescript.

PR #291, opened 2021-02-08
Switch to ESLint

wwlorey:
Fixes https://github.com/microsoft/vscode-azurestaticwebapps/issues/129


package.json
```diff
@@ -434,7 +434,7 @@
         "cleanReadme": "gulp cleanReadme",
         "compile": "tsc -watch",
         "package": "vsce package --githubBranch main",
-        "lint": "tslint --project tsconfig.json -e src/*.d.ts -t verbose",
```
nturinski, line 437: Would you mind adding the lint-fix script as well?

src/commands/createStaticWebApp/AppLocationStep.ts
```diff
@@ -24,6 +24,7 @@ export class AppLocationStep extends AzureWizardPromptStep<IStaticWebAppWizardCo
         return !wizardContext.appLocation;
     }
 
+    // eslint-disable-next-line @typescript-eslint/require-await
```
nturinski, line 27: Rather than disabling this rule, can you make this not async anymore?  Or will the ts error still get in the way?

src/commands/createStaticWebApp/addWorkspaceTelemetry.ts
```diff
@@ -14,7 +14,6 @@ export function addWorkspaceTelemetry(wizardContext: IStaticWebAppWizardContext)
     wizardContext.telemetry.properties.gotRemote = String(!!wizardContext.repoHtmlUrl);
 
     const localGit: git.SimpleGit = git(wizardContext.fsPath);
-    // tslint:disable-next-line:no-floating-promises
     localGit.status().then(() => {
```
nturinski, line 17: You don't need to put `void` here?

nturinski (approved):

src/commands/createStaticWebApp/AppLocationStep.ts
wwlorey: We get a typescript error when the function isn't async:

Screen Shot 2021-02-09 at 10 00 49 AM

src/commands/createStaticWebApp/addWorkspaceTelemetry.ts
wwlorey: ESLint doesn't complain about this - I think because we're making it synchronous with `.then()`

PR #294, opened 2021-02-17
Use "Report Issue" command

ejizba:
instead of button


nturinski (commented):
nit: https://github.com/microsoft/vscode-azurestaticwebapps/search?q=suppressReportIssue 😏

ejizba:
I left those in because they don't do any harm and I feel like there's still a risk we switch back to the "Report Issue" button. And when I say "risk", I mostly mean because in the past people on our team have had strong opinions about the button

nturinski (approved):
Oops, that was a nit.  I totally thought I still approved, sorry

PR #295, opened 2021-02-17
Add local project and configuration tree items

wwlorey:
I removed the click to open config functionality until I can figure out a better way to know which YAML file is the one we want. 

Also it's possible for `staticwebapp.config.json` to be somewhere other than the root: "The recommended location for the staticwebapp.config.json is your application root. However, the file may be placed in any location within your application source code folder." ([source](https://review.docs.microsoft.com/en-us/azure/static-web-apps/configuration?branch=pr-en-us-139784)) For now this just checks the workspace root.

image


src/tree/AzureAccountTreeItem.ts
```diff
@@ -3,15 +3,77 @@
...
+        this.disposables.push(workspace.onDidChangeConfiguration(async event => {
+            await callWithTelemetryAndErrorHandling('AzureAccountTreeItemWithProject.onDidChangeConfiguration', async (context: IActionContext) => {
+                context.errorHandling.suppressDisplay = true;
+                context.telemetry.suppressIfSuccessful = true;
+                const settings: string[] = [appSubpathSetting, apiSubpathSetting, appArtifactSubpathSetting, outputSubpathSetting];
+                if (settings.some(s => event.affectsConfiguration(`${ext.prefix}.${s}`))) {
+                    await this.refresh(context);
+                }
+            });
+        }));
```
wwlorey, line 37: I'm not sure if this is necessary - will any tree items need to update when these settings change?

src/tree/localProject/ConfigTreeItem.ts
```diff
@@ -0,0 +1,24 @@
...
+    public readonly label: string;
+    public parent: ConfigGroupTreeItem;
+
+    public constructor(parent: ConfigGroupTreeItem, label: string) {
+        super(parent);
+        this.label = label;
+    }
+
+    public get iconPath(): TreeItemIconPath {
+        return new ThemeIcon('gear');
```
wwlorey, line 22: I'm not super sold on the gear icon here but as usual pickins are slim 🤷‍♂️

nturinski: Hmm, maybe we use a more simple icon for this and the gear for its parent?

Something like this might work: 
image

src/tree/localProject/ConfigGroupTreeItem.ts
```diff
@@ -0,0 +1,40 @@
...
+
+    public get iconPath(): TreeItemIconPath {
+        return treeUtils.getThemedIconPath('settings');
+    }
+
+    public hasMoreChildrenImpl(): boolean {
+        return false;
+    }
+
+    // eslint-disable-next-line @typescript-eslint/require-await
```
nturinski, line 30: You have an await in this function so I assume that you wouldn't need this eslint-disable line.

wwlorey:
Here's the updated version with new icons:

Screen Shot 2021-02-18 at 11 11 06 AM

Now it matches the icons in app service's "Application Settings" node

src/tree/AzureAccountTreeItem.ts
```diff
@@ -3,15 +3,46 @@
...
+
+    public async loadMoreChildrenImpl(clearCache: boolean, context: IActionContext): Promise<AzExtTreeItem[]> {
+        const children: AzExtTreeItem[] = await super.loadMoreChildrenImpl(clearCache, context);
+
+        for (const workspaceFolder of workspace.workspaceFolders || []) {
+            const workspaceFolderPath: string = workspaceFolder.uri.fsPath;
+            const configFilePathExists: boolean = await pathExists(join(workspaceFolderPath, configFileName));
+
+            if (configFilePathExists) {
+                const localProjectTreeItem: LocalProjectTreeItem = new LocalProjectTreeItem(this, workspaceFolderPath);
```
ejizba, line 31: Can we put all of this UI behind a feature flag? I have concerns, but I feel like they don't need to be ironed out in the first PR. I think this is a big feature that will have a ton of PRs and changes between now and when it's final. Using a feature flag would give me peace of mind to have this merged without trying to block on things 🙂

ejizba:
Feature flag looks good, thanks 👌 I'll let @nturinski finish off the review of this PR

src/tree/localProject/LocalProjectTreeItem.ts
```diff
@@ -0,0 +1,46 @@
...
+    private readonly _projectName: string;
+
+    public constructor(parent: AzExtParentTreeItem, projectPath: string) {
+        super(parent);
+        this.projectPath = projectPath;
+        this._projectName = basename(projectPath);
+    }
+
+    public get iconPath(): TreeItemIconPath {
+        return treeUtils.getThemedIconPath('azure-staticwebapps');
```
nturinski, line 27: We should probably look for a different icon to differentiate it from the remote static web apps.  Not this PR, but something to keep in mind.

nturinski (approved):

Issue #288, opened 2021-01-28
Some suggestions to improve the SWA creation flow


anthonychu:
Would it make sense to default to the name of the current folder?
![image](https://user-images.githubusercontent.com/3982077/106177540-7f3cd000-614d-11eb-9bef-af068999b411.png)

Can this default to everything selected? Also it's unclear what this does. Appears to create the .gitignore file?
![image](https://user-images.githubusercontent.com/3982077/106177895-f4100a00-614d-11eb-8c0a-c96e02da5a89.png)

Not sure if `./` is a valid option. Our docs use `/` for root.
![image](https://user-images.githubusercontent.com/3982077/106178123-3afdff80-614e-11eb-8683-67afafea30e3.png)

The text talks about `/api`, but dialog defaults to `api`. Perhaps change the text to match.
![image](https://user-images.githubusercontent.com/3982077/106178354-83b5b880-614e-11eb-8f60-5c8bfae46cac.png)

Maybe there should be some text about leaving this blank.
![image](https://user-images.githubusercontent.com/3982077/106178509-b5c71a80-614e-11eb-9c3b-2701382722f5.png)

Some steps say (press enter to confirm or esc to cancel). Not sure if it's just me, but when I read it while going through the steps, I almost read it as "esc to skip this step".

Edit: one more. If you just start typing a folder name here and press enter, it just exits out of the flow. It's not clear that you have to press enter at `Manually enter location` before entering the location.

![image](https://user-images.githubusercontent.com/3982077/107427099-e2c3e780-6ad5-11eb-8d47-082142245af3.png)

PR #300, opened 2021-02-17
Add SWA creation flow improvements

wwlorey:
Fixes https://github.com/microsoft/vscode-azurestaticwebapps/issues/288


src/commands/createRepo/GitignoreListStep.ts
```diff
@@ -19,11 +19,9 @@ export class GitignoreListStep extends AzureWizardPromptStep<IStaticWebAppWizard
         const gitignorePath: string = join(fsPath, '.gitignore');
         const files: string[] = (await fse.readdir(fsPath))
             .filter(name => name !== '.git');
-        const placeHolder: string = localize('ignore', 'Select which files should be included in the repository.');
+        const placeHolder: string = localize('ignore', 'Select which files should be ignored in the repository.');
```
nturinski, line 22: So the only reason I'm hesitant on this change is because the publish experience from the SCM client in VS Code asks for which files to include rather than ignore.  So if users have used that experience before, this might be confusing.

Anthony's request was to select everything by default, but this seems like a good alternative.  I think I'd like a second opinion: @ejizba @fiveisprime

ejizba: Regardless of what's better ("ignored" or "included"), I don't think you can flip it after it's already been released. If we were convinced one was markedly better than the other (I'm not), we could maybe come up with a transition plan, but I think it would be messy.

That being said, I think something _does_ need to change. Users should be able to click "enter" for every prompt in our wizard and it should "just work".

Here's a few options I like better:
1. Get rid of the prompt and just use some sort of default `.gitignore`. Nathan modeled this experience after the SCM client, but maybe this prompt only makes sense in the context of the SCM wizard and not the SWA wizard
2. Go with Anthony's request and select stuff by default. Doesn't have to be everything - you could be smart and uncheck some common folders people would ignore. One benefit of checking things by default is I think it makes it more obvious that we _did_ have a default and they can just click "enter"

src/commands/createRepo/RepoNameStep.ts
```diff
@@ -15,7 +16,8 @@ export class RepoNameStep extends AzureWizardPromptStep<IStaticWebAppWizardConte
     public async prompt(wizardContext: IStaticWebAppWizardContext): Promise<void> {
         wizardContext.newRepoName = (await ext.ui.showInputBox({
             prompt: localize('AppServicePlanPrompt', 'Enter the name of the new GitHub repository. Special characters will be replaced with "-" upon creation.'),
-            validateInput: async (value: string): Promise<string | undefined> => await this.validateRepoName(wizardContext, value)
+            validateInput: async (value: string): Promise<string | undefined> => await this.validateRepoName(wizardContext, value),
+            value: workspace.workspaceFolders ? workspace.workspaceFolders[0].name : undefined
```
ejizba, line 20: It bugs me when the default value used is immediately invalid, which is possible here. Can we either:
1. Convert invalid names to valid ones (shorten it, replace special characters, etc.)
2. Only use a default value if we detect it _is_ valid

src/commands/createStaticWebApp/OutputLocationStep.ts
```diff
@@ -16,7 +16,7 @@ export class OutputLocationStep extends AzureWizardPromptStep<IStaticWebAppWizar
         const defaultLocation: string = 'build';
         wizardContext.outputLocation = (await ext.ui.showInputBox({
             value: getWorkspaceSetting(outputSubpathSetting, wizardContext.fsPath) || getWorkspaceSetting(appArtifactSubpathSetting, wizardContext.fsPath) || defaultLocation,
-            prompt: localize('publishLocation', "Enter the path of your build output relative to your app's location. For example, setting a value of 'build' when your app location is set to 'app' will cause the content at 'app/build' to be served.")
+            prompt: localize('publishLocation', "Enter the path of your build output relative to your app's location. For example, setting a value of 'build' when your app location is set to 'app' will cause the content at 'app/build' to be served. Leave blank to use root.")
```
ejizba, line 19: > Leave blank to use root.

I don't think this is super accurate. Sounds like it'll use the app's location, not the root of the workspace. Also, I think Anthony's comment was more for projects that don't have a build (aka they're not using a framework), so this step doesn't matter.

What about this:

> Enter the path of your build output relative to your app's location, or leave blank if it has no build. For example, setting a value of 'build' when your app location is set to 'app' will cause the content at 'app/build' to be served.

src/commands/createRepo/GitignoreListStep.ts
wwlorey: I like option 2. It seems better to give people the option to ignore specific files if they want

src/commands/createRepo/RepoNameStep.ts
wwlorey: I went with option 2. Part of the validation is that the repo name is available and it seemed simpler to just not show a default value than try to generate a unique repo name

src/commands/createStaticWebApp/OutputLocationStep.ts
nturinski: > Also, I think Anthony's comment was more for projects that don't have a build (aka they're not using a framework), so this step doesn't matter.

https://docs.microsoft.com/en-us/azure/static-web-apps/front-end-frameworks

I don't think that this is accurate either.  I'm pretty sure all Static Web Apps need to have a build output location otherwise we would let users just the skip the step.  Of course there are cases where the build script outputs all the code in the app folder itself, in which case you'd just leave it blank to refer to the app folder.

nturinski: Actually, I take that back, it _is_ an optional parameter.

ejizba: You're welcome to suggest a better wording 😊 I was kind of going off the "No Framework" tab in this tutorial


https://docs.microsoft.com/en-us/azure/static-web-apps/getting-started?tabs=vanilla-javascript
Screen Shot 2021-02-18 at 2 33 58 PM

nturinski: Nuts. Eric wins again.  This is why you don't challenge the boss.

src/commands/createRepo/GitignoreListStep.ts
```diff
@@ -34,3 +35,8 @@ export class GitignoreListStep extends AzureWizardPromptStep<IStaticWebAppWizard
         return wizardContext.gitignoreExists === false;
     }
 }
+
+
+function shouldIncludeInRepo(pick: QuickPickItem): boolean {
+    return !/^(\.vscode|.*\.log|\.DS_Store)$/.test(pick.label);
```
nturinski, line 41: How did you decide on this list of folders/files to ignore?

I think we should use this standard gitignore https://github.com/github/gitignore/blob/master/Node.gitignore to help us decide which files to ignore.

ejizba: Honestly I think I'm leaning to get rid of the prompt and just use that gitignore. There's a lot of things in there that I might not have in my repo _yet_ (so they won't show up in the quick pick), but I probably still want to ignore. SWA is pretty Node focused at the moment anyways, right?

wwlorey: I grabbed what looked applicable from this [gist](https://gist.github.com/octocat/9257657).

I see that standard gitignore is for node - should we account for other types of project like Django?

ejizba: Also

shame

That's for falling into the trap of ignoring the `.vscode` folder, which a lot of VS Code noobs do and causes us problems

wwlorey: > I see that standard gitignore is for node - should we account for other types of project like Django?

I didn't see Eric's comment before I posted. I agree it's probs simpler to use the default node gitignore as long as that's the predominant type

nturinski: It's the predominant type, but SWA technically supports Python and C# as well...

nturinski: > Honestly I think I'm leaning to get rid of the prompt and just use that gitignore. There's a lot of things in there that I might not have in my repo _yet_ (so they won't show up in the quick pick), but I probably still want to ignore. SWA is pretty Node focused at the moment anyways, right?

I mean ideally they already have a `.gitignore` in their project so that this is a non-issue.  There's no real way for us to know what their SWA is.  We could try to take a guess based on their app/output locations, but they don't ever select a runtime on creation.

I dunno, I feel like just having everything selected by default isn't the worst experience.  I would imagine most demo repos would already have a `.gitignore`.  A more developed project would probably have a `.gitignore` . And a quick test project doesn't really need the best `.gitignore`.  I'm not totally against having the default node `.gitignore` as a solution but I think it will age poorly.

ejizba: > It's the predominant type, but SWA technically supports Python and C# as well...

I think JS, Python, and C# will be our primary languages for a really long time. If we're not satisfied with always using Node, I think we could come up with a quick pick for the user to pick between those 3. That being said, I think the node gitignore will always be better than a 2-line or empty gitignore, which is what users in the current path would end up with. So I think this change is still an improvement

ejizba (approved):

PR #304, opened 2021-02-22
Add 'Create SWA from Local Project' node

wwlorey:
If there are yaml files in `.github/workflows`, then the "Configuration" tree item is shown:

Screen Shot 2021-02-22 at 3 31 14 PM

If not, it shows a 'create SWA' tree item instead:

Screen Shot 2021-02-22 at 3 28 39 PM


src/commands/createStaticWebApp/createStaticWebApp.ts
```diff
@@ -41,3 +42,7 @@ export async function createStaticWebApp(context: IActionContext & Partial<ICrea
 export async function createStaticWebAppAdvanced(context: IActionContext, node?: SubscriptionTreeItem): Promise<StaticWebAppTreeItem> {
     return await createStaticWebApp({ ...context, advancedCreation: true }, node);
 }
+
+export async function createStaticWebAppFromLocalProject(context: IActionContext, projectPath: string): Promise<StaticWebAppTreeItem> {
```
nturinski, line 46: I don't think that we can assume that the user wants to publish to a new repo when creating a new static web app.

For instance, I can see a scenario where I have a GitHub project, clone it down, work on it a bit, push it up, and then create a new SWA.

I think if we detect a remote, we could assume that the user wants to use that repository.  If there isn't an origin, then we can assume publish to a new repo.

src/tree/localProject/LocalProjectTreeItem.ts
```diff
@@ -39,8 +40,30 @@ export class LocalProjectTreeItem extends AzExtParentTreeItem {
...
 
-    // eslint-disable-next-line @typescript-eslint/require-await
     public async loadMoreChildrenImpl(): Promise<AzExtTreeItem[]> {
-        return [new ConfigGroupTreeItem(this)];
+        const workflows: string[] = await this.getWorkflows(this.projectPath);
+        if (workflows.length) {
+            return [new ConfigGroupTreeItem(this)];
+        } else {
+            const options: IGenericTreeItemOptions = {
+                label: 'Create Static Web App from Local Project...',
```
nturinski, line 49: nit: localize please

neelip:
I am seeing "Create Static Web App..." option even after successfully creating and deploying one with local repo. The Local Project node is not loading the Configure node. 
image

wwlorey:
> I am seeing "Create Static Web App..." option even after successfully creating and deploying one with local repo. The Local Project node is not loading the Configurations node.
> image

You have to `git pull` to get the yaml files created by the Azure bot when you create the static web app. This will load the configuration node instead of the "Create SWA" node. But that isn't obvious. 

Would it be better to show a prompt that you must pull to update the build config? We could show this after creating the SWA

neelip:
> > I am seeing "Create Static Web App..." option even after successfully creating and deploying one with local repo. The Local Project node is not loading the Configurations node.
> > image
> 
> You have to `git pull` to get the yaml files created by the Azure bot when you create the static web app. This will load the configuration node instead of the "Create SWA" node. But that isn't obvious.
> 
> Would it be better to show a prompt that you must pull to update the build config? We could show this after creating the SWA

That makes sense, but I like the idea of notifying the user. Especially the new users will benefit from it, and it can have "don't show again" option if needed.

wwlorey:
> That makes sense, but I like the idea of notifying the user. Especially the new users will benefit from it, and it can have "don't show again" option if needed.

I agree - I think that would be v helpful for this scenario and for the standard case when you've created a SWA and are just unsure of next steps

@nturinski what do you think?

nturinski:
We might be better off checking to see if the user has a remote for their local project, and if so, using this [command](https://github.com/microsoft/vscode-azurestaticwebapps/blob/main/src/utils/gitHubUtils.ts#L167-L178) to check for the existence of the .yml files.

Then if the user tries to click on configuration nodes and they don't have the .yml file locally, we prompt them to pull down their files.  What do you think?

I agree with Neelima that it isn't immediately obvious why the configuration node doesn't fill out after creating a static web app.  I think the proposed solution would solve that problem without the use of a notification _until_ the user absolutely needs to do something.

ejizba:
If the tree is the main source of confusion, you could just have a tree node that says "Hey do a 'git pull'" rather than a notification. Not sure if that would be in-addition-to or in-place-of "Create Static Web App from Local Project..."

neelip:
I agree with not prompting a notification until necessary, it could be easily ignored and end up not solving the problem.  
I like the idea of replacing the tree node with a new node with the "pull" prompt.

wwlorey:
Screen Shot 2021-03-08 at 5 28 27 PM

I can demo this tomorrow in standup but I added this 'git pull' node that shows up if there are remote workflows but no local workflows.

wwlorey:
After digging into the `git pull` vs. `git fetch` question a bit, I think a standard `git pull` is the way to go.

If you have *uncommitted* local changes that conflict with files on the remote, git will let you know and abort the pull:
Screen Shot 2021-03-09 at 4 16 30 PM

If you have *committed* local changes that conflict with files on the remote, git will attempt to merge and let you fix conflicts: 
Screen Shot 2021-03-09 at 4 17 58 PM

So running `git pull` automatically won't work cleanly in these cases but at least the error messages are actionable and there isn't data loss.

neelip:
I am unable to see any context menus for Local Project or Configuration nodes when I am trying to test it. I think the recent code changes affected this, as I was able to see the menus 2 commits before.

wwlorey:
I actually hadn't added context menus yet but thanks for the reminder. I started that process here: https://github.com/microsoft/vscode-azurestaticwebapps/pull/310

nturinski (commented):
Sorry for the extra work, but I've been thinking that we should have the create static web app node be a separate PR from the gitPull PR.  These could be considered two separate features (although they're related) but mostly I'm not entirely satisfied with our current experience and I would like to think about/discuss it some more, but I don't want it to block this PR.

wwlorey:
No problem - just removed the 'git pull' stuff so it only includes the SWA creation node

src/commands/createStaticWebApp/CreateScenarioListStep.ts
```diff
@@ -57,9 +57,12 @@ export class CreateScenarioListStep extends AzureWizardPromptStep<IStaticWebAppW
         if (wizardContext.createScenario === 'publishToNewRepo') {
             // calling to verify the user has git enabled so they don't go through the whole process and then it fails
             await getGitApi();
-            if (workspace.workspaceFolders && workspace.workspaceFolders.length === 1 && !wizardContext.advancedCreation) {
+            if (!wizardContext.fsPath && workspace.workspaceFolders && workspace.workspaceFolders.length === 1 && !wizardContext.advancedCreation) {
                 // only one workspace folder, in basic mode, set that as local cod
                 wizardContext.fsPath = workspace.workspaceFolders[0].uri.fsPath;
+            }
+
+            if (wizardContext.fsPath) {
```
nturinski, line 65: We don't want to set these `workspaceContexts` when the user creates via `advancedCreation` which I think it'll be doing now.

nturinski (approved):
Besides that one change in behavior, I think this should be good.

PR #307, opened 2021-03-10
Update local project icon

wwlorey:
I grabbed function's local project icon. Not the best but it's better than what we had before.

| Old | New |
|-|-|
| Screen Shot 2021-03-09 at 4 50 50 PM| Screen Shot 2021-03-09 at 4 48 52 PM|


nturinski (approved):
Where are those sick photoshop skillz?

PR #309, opened 2021-03-10
Add connected SWA node to local project tree

wwlorey:
This touches some of the same code as https://github.com/microsoft/vscode-azurestaticwebapps/pull/304 but I wanted to get some early feedback on this.

swag-reveal

I opted to add a link to the SWA tree item (not the SWA tree item itself) because I could see problems keeping duplicate tree items in sync.

`findStaticWebAppTreeItem` uses `WebSiteManagementClient` to find the static site and then uses that to create the `StaticWebAppTreeItem` we're looking for. This should be faster than finding the tree item directly (by using `SubscriptionTreeItem.loadAllChildren`) because we're just making API calls instead of creating *all* the tree items.


neelip:
I like this idea. It makes sense to not have a duplicate tree item when the original tree item accessible in the same window.

src/tree/localProject/LocalProjectTreeItem.ts
```diff
@@ -40,7 +45,25 @@ export class LocalProjectTreeItem extends AzExtParentTreeItem {
...
-        return [new ConfigGroupTreeItem(this)];
+    public async loadMoreChildrenImpl(_clearCache: boolean, context: IActionContext): Promise<AzExtTreeItem[]> {
+        const treeItems: AzExtTreeItem[] = [new ConfigGroupTreeItem(this)];
+
+        const repositoryUrl: string | undefined = (await tryGetRemote(context, this.projectPath))?.html_url;
+        if (repositoryUrl) {
+            const staticWebAppTreeItem: StaticWebAppTreeItem | undefined = await this.parent.findStaticWebAppTreeItem(context, repositoryUrl);
+            if (staticWebAppTreeItem) {
+                const options: IGenericTreeItemOptions = {
+                    label: localize('reveal', 'Reveal connected Static Web App'),
```
ejizba, line 56: IMO, the root of the problem is that so far the "Local Project" tree item only has stuff related to the remote resource (hence why you're trying to link it up), not the local project. In the Functions case, we have files like host.json/functions.json/local.settings.json that are all used to run/develop the project locally. The only files you use so far are config files related to CI/CD. Functions has no concept of the CI/CD pattern, so I don't think you should be using the same "Local Project" naming/design.

I'm not necessarily opposed to your fix and I agree there's a problem here, but I just think there needs to be a bigger re-design to solve the root problem. For example, maybe you 
1. Rename "Local Project" to "Connected App"
2. Add a `(Connected)` description to the SWA node (kind of like we have in Databases) and move "Configuration" out from under "Local Project" and into the SWA node (maybe named "Local Configuration" or something)

wwlorey: Yeah right now there isn't much local project-specific content under this tree item. That will change eventually though when the emulator is supported and we'll also show routes that are defined in the local staticwebapp.config.json file.

I agree that "Local Project" still may not be the best naming for this though. It's probs worth discussing this more in the next standup

ejizba: Yeah fwiw this is one of my biggest concerns I may have blocked the "Local Project" feature on except for the fact that it's under a feature flag. I agree this could change a lot once the emulator is supported, so I'm willing to wait and see how that turns out.

I still haven't decided if I would approve this PR as-is, but I will wait to decide until if/when it goes out of draft 😝

wwlorey:
Closing due to local project pivot

PR #311, opened 2021-03-11
Turn off require-await rule

wwlorey:
Following Eric's lead in [functions](https://github.com/microsoft/vscode-azurefunctions/blob/6f8215cd997d910538a74fd022873bd742e2edc9/.eslintrc.js#L36) and turning off this rule.


src/commands/createStaticWebApp/ApiLocationStep.ts
```diff
@@ -24,7 +24,6 @@ export class ApiLocationStep extends AzureWizardPromptStep<IStaticWebAppWizardCo
         return !wizardContext.apiLocation;
     }
 
-    // eslint-disable-next-line @typescript-eslint/require-await
```
neelip, line 27: I don't see this line of code removed from Eric's commit. Do you think it might be because we might use it again later?

wwlorey: Eric could have deleted all instances of this line since the rule was turned off but I'm assuming it was just a time saver for him not to. It doesn't hurt anything to keep these exceptions around but they're also no longer required

ejizba: Yeah I was annoyed with the rule when authoring new code, so I turned it off. I wanted to save time and I didn't want to clutter the diff of that PR. If I was doing a PR to specifically remove the rule, I would do exactly what Will did.

ejizba (approved):

PR #312, opened 2021-03-12
Add new local project tree item icon round #2

wwlorey:
| Current | Proposed |
| - | - |
|Screen Shot 2021-03-12 at 2 09 26 PMScreen Shot 2021-03-12 at 2 09 14 PM |Screen Shot 2021-03-16 at 8 39 22 AMScreen Shot 2021-03-16 at 8 39 11 AM|


src/tree/localProject/LocalProjectTreeItem.ts
```diff
@@ -24,7 +24,7 @@ export class LocalProjectTreeItem extends AzExtParentTreeItem {
     }
 
     public get iconPath(): TreeItemIconPath {
-        return treeUtils.getThemedIconPath('CreateNewProject');
+        return treeUtils.getThemedIconPath('folder');
```
ejizba, line 27: Uh we're not usually in the business of creating our own icons. You can use any existing icon you find, but otherwise I think you should talk to an actual icon person if you need a new one (Jing probably knows who to talk to).

ejizba (changes requested):

src/tree/localProject/LocalProjectTreeItem.ts
wwlorey: The consensus in standup on Friday was that if you and Nathan are OK with it we can use this one otherwise we can escalate it to a designer

ejizba: My concern is not the icon itself (that's very subjective), but how we got it. I think it's a slippery slope if we start designing our own icons and I think there should be plenty of existing icons to choose from. For example if you want a folder without the lightning bolt, you could go with this one: https://github.com/microsoft/vscode-azuretools/blob/main/appservice/resources/dark/folder.svg

There's probably some folder icons in the Azure icon set, too

ejizba: Also on a somewhat related note, we want to start using `ThemeIcon` as much as possible so that users can use their own icons and get it from the theme instead of us. We can do that for simple things like a folder, but we'll run into problems for Azure-specific icons.

https://code.visualstudio.com/updates/v1_54#_icon-contribution-point

wwlorey: Going with the folder codicon sounds best. For whatever reason that specific icon doesn't seem to work with `ThemeIcon` right now though lol. So pending on https://github.com/microsoft/vscode/issues/119028

wwlorey:
One problem with using `ThemeIcon` is that some themes hide folder icon. As far as the standard VS Code icon themes go, folder icons only show up in the `Minimal` theme (not `Seti` or `None`).

ejizba (approved):
> One problem with using ThemeIcon is that some themes hide folder icon. As far as the standard VS Code icon themes go, folder icons only show up in the Minimal theme (not Seti or None).

Yeah not sure if I like that or not. In any case - as long as we're getting icons from some existing source, I'm pretty much fine with whatever you pick.

PR #314, opened 2021-03-16
Execute `git pull` and open workflow file after creating SWA

wwlorey:
Fixes https://github.com/microsoft/vscode-azurestaticwebapps/issues/340


src/tree/localProject/LocalProjectTreeItem.ts
```diff
@@ -41,17 +46,29 @@ export class LocalProjectTreeItem extends AzExtParentTreeItem {
...
         } else {
-            const options: IGenericTreeItemOptions = {
-                label: localize('createSWA', 'Create Static Web App from Local Project...'),
-                iconPath: treeUtils.getThemedIconPath('add'),
-                commandId: 'staticWebApps.createStaticWebAppFromLocalProject',
-                contextValue: 'createStaticWebAppFromLocalProject'
-            };
+            let options: IGenericTreeItemOptions;
+            const remoteWorkflows: GitTreeData[] = await this.getRemoteWorkflows(context, this.projectPath);
+            if (remoteWorkflows.length) {
```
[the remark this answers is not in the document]
wwlorey: That's a good point. I'm not sure how to mitigate that. Maybe we should only show the "git pull" node if there aren't any uncommitted local changes?

src/commands/gitPull.ts
```diff
@@ -0,0 +1,40 @@
...
+    const git: API = await getGitApi();
+    const projectUri: Uri = Uri.file(localProjectTreeItem.projectPath);
+    const repo: Repository | null = git.getRepository(projectUri);
+
+    if (repo) {
+        try {
+            await repo.pull();
+        } catch (error) {
+            const parsedError: IParsedError = parseError(error);
+            if (/Failed to execute git/.test(parsedError.message)) {
```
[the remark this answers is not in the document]
wwlorey: For some reason I'm not seeing a `gitErrorCode` when I have uncommitted local changes:

Screen Shot 2021-03-18 at 2 42 53 PM

src/commands/gitPull.ts
```diff
@@ -0,0 +1,40 @@
...
+                /* eslint-disable @typescript-eslint/no-unsafe-member-access */
+                error.stderr && ext.outputChannel.appendLog(error.stderr);
+                error.stdout && ext.outputChannel.appendLog(error.stdout);
```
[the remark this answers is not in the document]
wwlorey: In the case of a merge conflict, there isn't any `stderr` but there is `stdout` even though the `git pull` didn't complete as intended. But you right - they should definitely be swapped.

nturinski:
We can discuss during standup, but I believe our design decision was that we will pull for the user automatically to get the new changes.

We won't let users create under the following conditions:
> No workspace opened
> Workspace has uncommitted local changes
   > If workspace has no remote, create remote for the user
   > If workspace has a remote, confirm user has admin rights to repo

nturinski:
This will be retrofitted to fix https://github.com/microsoft/vscode-azurestaticwebapps/issues/340 once is in https://github.com/microsoft/vscode-azurestaticwebapps/pull/335

src/commands/openYAMLConfigFile.ts
```diff
@@ -23,7 +23,14 @@ export async function openYAMLConfigFile(context: IActionContext, node?: StaticW
     if (node instanceof StaticWebAppTreeItem || node instanceof EnvironmentTreeItem && node.gitHubConfigGroupTreeItems.length === 0) {
         const defaultHostname: string = node instanceof StaticWebAppTreeItem ? node.defaultHostname : node.parent.defaultHostname;
         const ymlFileName: string = `.github/workflows/azure-static-web-apps-${defaultHostname.split('.')[0]}.yml`;
-        return await openUrl(`${node.repositoryUrl}/edit/${node.branch}/${ymlFileName}`);
+        const localYamlFiles: Uri[] = await workspace.findFiles(ymlFileName);
+
+        if (localYamlFiles.length === 1) {
+            await window.showTextDocument(localYamlFiles[0]);
```
nturinski, line 29: I think that you could combine this with the logic below where you prompt the user for a yaml file.  It's kind of weird, but you have two different branches to open a local yaml file now.

src/commands/createStaticWebApp/createStaticWebApp.ts
```diff
@@ -68,6 +71,9 @@ export async function createStaticWebApp(context: IActionContext & Partial<ICrea
         }
     });
 
+    folder && await gitPull(folder.uri);
```
nturinski, line 74: I think you could just change `gitPull` to accept `repo` as a paramemter and pass it `context.repo` instead.

src/utils/gitUtils.ts
```diff
@@ -165,6 +165,20 @@ export async function promptForDefaultBranch(context: IActionContext, repo: Repo
...
 }
 
+export async function gitPull(repo: Repository): Promise<void> {
+    try {
+        const options: ProgressOptions = {
+            location: ProgressLocation.Notification,
+            title: localize('executingGitPull', 'Executing "git pull"...')
+        };
+        await window.withProgress(options, async () => {
+            await repo.pull();
```
nturinski, line 175: I would prefer you only wrap this line in the try/catch.  As pointed out in this PR: https://github.com/microsoft/vscode-azurestaticwebapps/pull/369#discussion_r625468725, we aren't verifying that the error is actually a Git error.  I'll fix that, but I still want to be careful where we use `GitError`.

src/utils/gitUtils.ts
```diff
@@ -165,6 +165,20 @@ export async function promptForDefaultBranch(context: IActionContext, repo: Repo
     }
 }
 
+export async function gitPull(repo: Repository): Promise<void> {
+    try {
+        const options: ProgressOptions = {
+            location: ProgressLocation.Notification,
+            title: localize('executingGitPull', 'Executing "git pull"...')
```
nturinski, line 172: Let's discuss offline, but I think we need a better messaging as to _why_ we're doing a git pull for the user.  Maybe something along the lines of `Performing "git pull" to retrieve xxx_workflow_file.yml...`

nturinski: @fiveisprime Thoughts

nturinski (approved):
Let's get this merged in.  If the string/messaging are off, we can fix that pretty easily.  Hopefully we can get some of the other SWA people to bug bash this and they can give us their opinion on what they think about the strings/flow of the entire creation experience.

PR #321, opened 2021-03-22
Move ConfigGroupTreeItem under EnvironmentTreeItem and make ConfigTreeItems clickable

wwlorey:
Screen Shot 2021-03-22 at 1 44 46 PM

### DIFF
--- a/src/commands/openYAMLConfigFile.ts
+++ b/src/commands/openYAMLConfigFile.ts
@@ -29,7 +29,7 @@ export async function openYAMLConfigFile(context: IActionContext, node?: StaticW
             // if we couldn't find it, then try opening it in GitHub
             if (await fse.pathExists(ymlFsPath)) {
                 const configDocument: TextDocument = await workspace.openTextDocument(ymlFsPath);
-                const selection: Range | undefined = await getSelection(configDocument, buildConfigToSelect);
+                const selection: Range | undefined = buildConfigToSelect ? await getSelection(configDocument, buildConfigToSelect) : undefined;
                 await window.showTextDocument(configDocument, { selection });
                 return;
             }
@@ -39,7 +39,7 @@ export async function openYAMLConfigFile(context: IActionContext, node?: StaticW
     await openUrl(`${node.repositoryUrl}/edit/${node.branch}/${ymlFileName}`);
 }
 
-async function getSelection(configDocument: TextDocument, buildConfigToSelect?: BuildConfig): Promise<Range | undefined> {
+async function getSelection(configDocument: TextDocument, buildConfigToSelect: BuildConfig): Promise<Range | undefined> {
     if (buildConfigToSelect) {
         const configRegex: RegExp = new RegExp(`${buildConfigToSelect}:`);
 
--- a/src/commands/openYAMLConfigFile.ts
+++ b/src/commands/openYAMLConfigFile.ts
@@ -5,15 +5,16 @@
 
 import * as fse from 'fs-extra';
 import * as path from 'path';
-import { window, workspace } from "vscode";
+import { Position, Range, TextDocument, window, workspace } from "vscode";
 import { IActionContext } from "vscode-azureextensionui";
 import { ext } from "../extensionVariables";
 import { EnvironmentTreeItem } from "../tree/EnvironmentTreeItem";
+import { BuildConfig } from '../tree/localProject/ConfigGroupTreeItem';
 import { StaticWebAppTreeItem } from "../tree/StaticWebAppTreeItem";
 import { openUrl } from "../utils/openUrl";
 import { getSingleRootFsPath } from '../utils/workspaceUtils';
 
-export async function openYAMLConfigFile(context: IActionContext, node?: StaticWebAppTreeItem | EnvironmentTreeItem): Promise<void> {
+export async function openYAMLConfigFile(context: IActionContext, node?: StaticWebAppTreeItem | EnvironmentTreeItem, buildConfigToSelect?: BuildConfig): Promise<void> {
     if (!node) {
         node = await ext.tree.showTreeItemPicker<EnvironmentTreeItem>(EnvironmentTreeItem.contextValue, context);
     }
@@ -27,11 +28,29 @@ export async function openYAMLConfigFile(context: IActionContext, node?: StaticW
             const ymlFsPath: string = path.join(fsPath, ymlFileName);
             // if we couldn't find it, then try opening it in GitHub
             if (await fse.pathExists(ymlFsPath)) {
-                await window.showTextDocument(await workspace.openTextDocument(ymlFsPath));
+                const configDocument: TextDocument = await workspace.openTextDocument(ymlFsPath);
+                const selection: Range | undefined = await getSelection(configDocument, buildConfigToSelect);
+                await window.showTextDocument(configDocument, { selection });
                 return;
             }
         }
     }
 
     await openUrl(`${node.repositoryUrl}/edit/${node.branch}/${ymlFileName}`);
+}
+
+async function getSelection(configDocument: TextDocument, buildConfigToSelect?: BuildConfig): Promise<Range | undefined> {
+    if (buildConfigToSelect) {
+        const configRegex: RegExp = new RegExp(`${buildConfigToSelect}:`);
+
+        let offset: number = configDocument.getText().search(configRegex);
+        // Shift the offset to the beginning of the build config's value
+        offset += `${buildConfigToSelect}: `.length;
+
+        const position: Position = configDocument.positionAt(offset);
+        const configValueRegex: RegExp = /['"].*['"]/;
+        return configDocument.getWordRangeAtPosition(position, configValueRegex);
+    }
+
+    return undefined;
 }
--- a/src/commands/openYAMLConfigFile.ts
+++ b/src/commands/openYAMLConfigFile.ts
@@ -3,40 +3,46 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import * as fse from 'fs-extra';
-import * as path from 'path';
+import { basename } from 'path';
 import { Position, Range, TextDocument, window, workspace } from "vscode";
-import { IActionContext } from "vscode-azureextensionui";
+import { IActionContext, IAzureQuickPickItem } from "vscode-azureextensionui";
 import { ext } from "../extensionVariables";
 import { EnvironmentTreeItem } from "../tree/EnvironmentTreeItem";
-import { BuildConfig } from '../tree/localProject/ConfigGroupTreeItem';
+import { BuildConfig, GitHubConfigGroupTreeItem } from '../tree/localProject/ConfigGroupTreeItem';
 import { StaticWebAppTreeItem } from "../tree/StaticWebAppTreeItem";
-import { getYAMLFileName } from '../utils/gitHubUtils';
+import { localize } from '../utils/localize';
 import { openUrl } from "../utils/openUrl";
-import { getSingleRootFsPath } from '../utils/workspaceUtils';
 
-export async function openYAMLConfigFile(context: IActionContext, node?: StaticWebAppTreeItem | EnvironmentTreeItem, buildConfigToSelect?: BuildConfig): Promise<void> {
+export async function openYAMLConfigFile(context: IActionContext, node?: StaticWebAppTreeItem | EnvironmentTreeItem | GitHubConfigGroupTreeItem, buildConfigToSelect?: BuildConfig): Promise<void> {
     if (!node) {
         node = await ext.tree.showTreeItemPicker<EnvironmentTreeItem>(EnvironmentTreeItem.contextValue, context);
     }
 
-    const ymlFileName: string = getYAMLFileName(node);
+    if (node instanceof StaticWebAppTreeItem || node instanceof EnvironmentTreeItem && node.gitHubConfigGroupTreeItems.length === 0) {
+        const defaultHostname: string = node instanceof StaticWebAppTreeItem ? node.defaultHostname : node.parent.defaultHostname;
+        const ymlFileName: string = `.github/workflows/azure-static-web-apps-${defaultHostname.split('.')[0]}.yml`;
+        return await openUrl(`${node.repositoryUrl}/edit/${node.branch}/${ymlFileName}`);
+    }
 
-    if (node instanceof EnvironmentTreeItem && node.inWorkspace) {
-        const fsPath: string | undefined = getSingleRootFsPath();
-        if (fsPath) {
-            const ymlFsPath: string = path.join(fsPath, ymlFileName);
-            // if we couldn't find it, then try opening it in GitHub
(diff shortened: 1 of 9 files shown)


src/tree/localProject/ConfigTreeItem.ts
```diff
@@ -5,17 +5,24 @@
...
-    public constructor(parent: ConfigGroupTreeItem, label: string) {
+    public constructor(parent: ConfigGroupTreeItem, buildConfig: BuildConfig) {
         super(parent);
-        this.label = label;
+        this.buildConfig = buildConfig;
+        this.commandArgs = [this.parent.parent, this.buildConfig];
+    }
+
+    public get label(): string {
+        return this.buildConfig;
```
nturinski, line 25: I think if we use the yaml library, we could probably more easily display the value of the build config as well.

src/tree/localProject/ConfigGroupTreeItem.ts
```diff
@@ -6,16 +6,18 @@
...
+import { EnvironmentTreeItem } from "../EnvironmentTreeItem";
 import { ConfigTreeItem } from "./ConfigTreeItem";
-import { LocalProjectTreeItem } from "./LocalProjectTreeItem";
+
+export type BuildConfig = 'app_location' | 'api_location' | 'output_location';
 
 export class ConfigGroupTreeItem extends AzExtParentTreeItem {
     public static contextValue: string = 'azureStaticConfigGroup';
     public contextValue: string = ConfigGroupTreeItem.contextValue;
     public readonly label: string = localize('config', 'Configuration');
```
nturinski, line 17: I think we should call this GitHub Configurations especially since it's on the remote SWA.

src/tree/localProject/ConfigGroupTreeItem.ts
```diff
@@ -28,12 +30,12 @@ export class ConfigGroupTreeItem extends AzExtParentTreeItem {
     }
 
     public async loadMoreChildrenImpl(): Promise<AzExtTreeItem[]> {
-        const buildSettings: string[] = ['app_location', 'api_location', 'app_artifact_location'];
+        const buildConfigs: BuildConfig[] = ['app_location', 'api_location', 'output_location'];
```
nturinski, line 33: We actually do need to support both output and app_artifact since older yamls use `app_artifact_location`

src/commands/openYAMLConfigFile.ts
```diff
@@ -27,11 +28,29 @@ export async function openYAMLConfigFile(context: IActionContext, node?: StaticW
...
+                await window.showTextDocument(configDocument, { selection });
                 return;
             }
         }
     }
 
     await openUrl(`${node.repositoryUrl}/edit/${node.branch}/${ymlFileName}`);
 }
+
+async function getSelection(configDocument: TextDocument, buildConfigToSelect?: BuildConfig): Promise<Range | undefined> {
```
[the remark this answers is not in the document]
wwlorey: It looks like that package is used for parsing specific values out of yaml files but since we're looking for the buildConfig's location in the document I think regex is the way to go

src/tree/localProject/ConfigTreeItem.ts
wwlorey: I do think the yaml library will be needed for this part. The `parse` function gives us an object with all the yaml config

src/tree/localProject/ConfigGroupTreeItem.ts
```diff
@@ -3,37 +3,101 @@
...
-            buildSettings,
-            'azureStaticConfigInvalid',
-            (setting: string) => new ConfigTreeItem(this, setting),
-            (setting: string) => setting
-        );
+        const treeItems: GitHubConfigTreeItem[] = [];
+        this.buildConfigs.forEach((value: string, config: BuildConfig) => {
+            treeItems.push(new GitHubConfigTreeItem(this, config, value));
+        });
+        return treeItems;
```
nturinski, line 101: This isn't exactly what I had in mind to resolve the problem.  Now if there's an error, the entire environment node doesn't load.

image

This line is causing the problem: `const yamlFiles: string[] = await readdir(workflowsDir);`

You should be ensuring that this file exists before you try to open it.

src/tree/localProject/ConfigTreeItem.ts
```diff
@@ -3,22 +3,31 @@
...
+    public constructor(parent: GitHubConfigGroupTreeItem, buildConfig: BuildConfig, buildConfigValue: string) {
         super(parent);
-        this.label = label;
+        this.buildConfig = buildConfig;
+        this.buildConfigValue = buildConfigValue;
+        this.commandArgs = [this.parent, this.buildConfig];
+    }
+
+    public get label(): string {
+        return `${this.buildConfig}="${this.buildConfigValue}"`;
```
nturinski, line 27: We should probably implement a `refreshImpl` on this tree item.  I can see CTI saying the values don't update on an edit.

src/tree/localProject/ConfigGroupTreeItem.ts
```diff
@@ -3,37 +3,101 @@
...
+
+    public static async createGitHubConfigGroupTreeItems(parent: EnvironmentTreeItem): Promise<GitHubConfigGroupTreeItem[]> {
+        if (parent.localProjectPath && parent.inWorkspace) {
+            const treeItems: GitHubConfigGroupTreeItem[] = [];
+            const workflowsDir: string = join(parent.localProjectPath, '.github/workflows');
+            const yamlFiles: string[] = await readdir(workflowsDir);
+
+            for (const yamlFile of yamlFiles) {
+                if (yamlFile.endsWith('.yml')) {
+                    const yamlFilePath: string = join(workflowsDir, yamlFile);
```
nturinski, line 41: Can we extract all of this into its own yaml-utils file or something of that nature?  I don't want this to clutter up all of the `ConfigGroupTreeItem` code, plus yaml parsing could probably be used in other places in the future.

src/tree/localProject/ConfigGroupTreeItem.ts
```diff
@@ -3,37 +3,101 @@
...
+                    const contents: string = (await readFile(yamlFilePath)).toString();
+
+                    if (/Azure\/static-web-apps-deploy/.test(contents)) {
+                        /* eslint-disable @typescript-eslint/no-unsafe-call, @typescript-eslint/no-explicit-any, @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access, no-prototype-builtins */
+                        const parsedYaml: any = await parse(contents);
+                        const buildConfigs: Map<BuildConfig, string> = new Map();
+
+                        for (const job of <any[]>Object.values(parsedYaml.jobs)) {
+                            for (const step of <any[]>Object.values(job['steps'])) {
+                                if (step.hasOwnProperty('id') && step['id'] === 'builddeploy') {
```
nturinski, line 51: Could you made this a function like, verifyIsSWAYaml?

src/tree/localProject/ConfigGroupTreeItem.ts
```diff
@@ -3,37 +3,101 @@
...
+                                    if (!step.hasOwnProperty('with') || !step['with'].hasOwnProperty('api_location') || !step['with'].hasOwnProperty('app_location')) {
```
nturinski, line 52: And then if it's verified, we can start extracting the values.

Also, could you make this a function that accepts the buildConfig parameter and returns the value of the configuration?  There's a decent amount of repeat code here.

src/tree/localProject/ConfigGroupTreeItem.ts
wwlorey: I moved the majority of this code to `yamlUtils.ts` and split it up into separate functions as some other comments suggested. It did look pretty cluttered hanging out in `ConfigGroupTreeItem` 😅

src/utils/yamlUtils.ts
```diff
@@ -0,0 +1,66 @@
...
+
+function isSwaYaml(yamlFileContents: string): boolean {
+    return /Azure\/static-web-apps-deploy/.test(yamlFileContents);
+}
+
+function isBuildDeployStep(step: any): boolean {
+    return step.hasOwnProperty('id') && step['id'] === 'builddeploy';
+}
+
+function getBuildConfigs(step: any, yamlFileName: string): Map<BuildConfig, string> | undefined {
```
nturinski, line 41: I think we can break this down a bit more.  If we make the BuildConfig a parameter, you can replace some of the repeat code such as

```
!step['with'].hasOwnProperty('api_location') || !step['with'].hasOwnProperty('app_location'))
 buildConfigs.set('api_location', getBuildConfig(step, 'api_location'))
    buildConfigs.set('app_location', getBuildConfig(step, 'app_location'))
void ext.ui.showWarningMessage(localize('mustIncludeLocations', `"{0}" must include "api_location" and "app_location". See the [workflow file guide](https://aka.ms/AAbrcox).`, yamlFileName));
```

And then just have it return the string values and set the map when you call `getBuildConfigs`.  To test for `output/app_artifact`, you can just check if the map has `output` defined (or has 3 settings) before checking for `app_artifact`

src/utils/yamlUtils.ts
```diff
@@ -0,0 +1,72 @@
...
+/* eslint-disable @typescript-eslint/no-unsafe-call, @typescript-eslint/no-explicit-any, @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access, no-prototype-builtins */
+async function getBuildDeployStep(yamlFileContents: string, yamlFileName: string): Promise<BuildDeployStep | undefined> {
+    if (/Azure\/static-web-apps-deploy/.test(yamlFileContents)) {
+        const parsedYaml: any = await parse(yamlFileContents);
+
+        for (const job of <any[]>Object.values(parsedYaml.jobs)) {
+            for (const step of <any[]>Object.values(job['steps'])) {
+                if (step.id === 'builddeploy' && step.with) {
+                    if (!stepIncludesBuildConfig(step, 'app_location')) {
+                        return showYamlWarningMessage(yamlFileName, 'app_location');
```
nturinski, line 48: I think it's pretty unintuitive that `showYamlWarningMessage` returns `undefined`.  I understand why you're doing it though.

I hatched this as a solution:

```
if (stepIncludesBuildConfig(step, 'app_location')) {
                                if (stepIncludesBuildConfig(step, 'api_location')) {
                                    if (stepIncludesBuildConfig(step, 'output_location') || stepIncludesBuildConfig(step, 'app_artifact_location')) {
                                        return <BuildDeployStep>step;
                                    }
                                    void showYamlWarningMessage(yamlFileName, 'output_location');
                                }
                                void showYamlWarningMessage(yamlFileName, 'api_location');
                            }
                            void showYamlWarningMessage(yamlFileName, 'app_location');
                        }
```

But it honestly might be grosser/less intuitive anyway.

src/utils/yamlUtils.ts
```diff
@@ -0,0 +1,72 @@
...
+                    return <BuildDeployStep>step;
+                }
+            }
+        }
+    }
+
+    return undefined;
+}
+
+function stepIncludesBuildConfig(step: { with: any }, buildConfig: BuildConfig): boolean {
```
nturinski, line 64: Use the BuildDeployStep here

src/utils/yamlUtils.ts
```diff
@@ -0,0 +1,72 @@
...
+    return !!step.with.hasOwnProperty(buildConfig);
+}
+/* eslint-enable @typescript-eslint/no-unsafe-call, @typescript-eslint/no-explicit-any, @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access, no-prototype-builtins */
```
nturinski, line 67: You should be able to remove this now.

src/utils/yamlUtils.ts
```diff
@@ -0,0 +1,72 @@
...
+
+function showYamlWarningMessage(yamlFileName: string, buildConfig: BuildConfig): undefined {
+    void ext.ui.showWarningMessage(localize('mustIncludeBuildConfig', `"{0}" must include "{1}". See the [workflow file guide](https://aka.ms/AAbrcox).`, yamlFileName, buildConfig));
```
nturinski, line 70: It'd be more helpful if the aka.ms link took them directly to here: https://docs.microsoft.com/en-us/azure/static-web-apps/github-actions-workflow#build-and-deploy

This section directly deals with the build config settings.

src/commands/openYAMLConfigFile.ts
```diff
@@ -3,35 +3,56 @@
...
+        yamlFilePath = node.yamlFilePath;
+    } else {
+        const picks: IAzureQuickPickItem<string>[] = node.gitHubConfigGroupTreeItems.map(configNode => {
+            return { label: basename(configNode.yamlFilePath), data: configNode.yamlFilePath };
+        });
+
+        if (picks.length === 1) {
+            yamlFilePath = picks[0].data;
+        } else {
+            const placeHolder: string = localize('selectGitHubConfig', 'Select the GitHub configuration file to open.');
```
nturinski, line 38: I think we should call this a GitHub workflow file.  It's what the bot calls it when it commits the file:

image

The configuration portion of it is what we're looking at, but the entire thing is a workflow file and lives in the .github/.workflows folder.

src/tree/localProject/ConfigGroupTreeItem.ts
```diff
@@ -3,37 +3,83 @@
...
+    public constructor(parent: EnvironmentTreeItem, yamlFilePath: string, buildConfigs: BuildConfigs) {
         super(parent);
+        this.yamlFilePath = yamlFilePath;
+        this.buildConfigs = buildConfigs;
+        this.id = `${parent.id}-${this.yamlFilePath}`;
+    }
+
+    public static async createGitHubConfigGroupTreeItems(parent: EnvironmentTreeItem): Promise<GitHubConfigGroupTreeItem[]> {
+        if (parent.localProjectPath && parent.inWorkspace) {
+            const treeItems: GitHubConfigGroupTreeItem[] = [];
```
nturinski, line 41: nit: Define treeItems outside of the if conditional, and just return treeItems outside of the if statement (on line 56)

src/utils/yamlUtils.ts
wwlorey: Unfortunately this shows warnings for buildConfigs that are already defined. For instance if `app_location` is defined but `api_location` isn't, these warnings are shown:

image

It just occurred to me that we could throw elses in there to fix that problem but I just committed changes to make the old version more readable. Let me know which version looks better to ya - I don't have much of a preference

src/utils/yamlUtils.ts
wwlorey: This should be fixed now

nturinski (approved):
👍 Thanks for all the hard work on this one. I think this is a pretty big'un 🐟

src/utils/yamlUtils.ts
nturinski: I like your solution.  The main thing was relying on returning `showYamlWarningMessage` which I thought was kind of odd.

wwlorey:
FISH

PR #328, opened 2021-04-01
Updates for experimental service

### LINKED ISSUE #327: Add the EXP framework
Similar to what was done for App Service.
This was discussed as a patch release.

nturinski:
Fixes #327 

Think it's worth downgrading npm so my package lock doesn't go crazy like this?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Change Log
 
-## 0.5.1 - 2021-04-02
+## 0.5.1 - 2021-04-12
 
 ### Added
 - Experimental framework to test incremental changes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Change Log
 
+## 0.5.1 - 2021-04-02
+
+### Added
+- Experimental framework to test incremental changes
+
 ## 0.5.0 - 2020-12-02
 
 ### Added
--- a/package.json
+++ b/package.json
@@ -474,8 +474,8 @@
         "git-url-parse": "^11.1.2",
         "moment": "^2.27.0",
         "simple-git": "^1.132.0",
-        "vscode-azureappservice": "^0.65.0",
-        "vscode-azureextensionui": "^0.35.1",
+        "vscode-azureappservice": "^0.67.0",
+        "vscode-azureextensionui": "^0.36.1",
         "vscode-nls": "^4.1.1"
     },
     "extensionDependencies": [
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,11 +5,18 @@
     "requires": true,
     "dependencies": {
         "@azure/abort-controller": {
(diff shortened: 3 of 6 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Change Log
 
+## 0.5.1 - 2021-04-01
```
ejizba, line 3: @fiveisprime if you thought this would be an easy/small change, it is not - mostly because the last release of SWA was on some older versions of the shared packages. With that in mind, is it still worth a patch release? What's the timing of when you need this? We could also consider doing a full release in the next week or two rather than all this extra work for a patch release. It's hard to remember the state of the shared packages at that point in time and I almost feel more comfortable with what's on main.

Regardless, I think this date should be moved so that we can at least do a full test pass.

package.json
```diff
@@ -474,8 +474,8 @@
         "git-url-parse": "^11.1.2",
         "moment": "^2.27.0",
         "simple-git": "^1.132.0",
-        "vscode-azureappservice": "^0.65.0",
-        "vscode-azureextensionui": "^0.35.1",
+        "vscode-azureappservice": "^0.73.0",
+        "vscode-azureextensionui": "^0.39.4",
```
ejizba, line 478: One way to reduce complexity is to not bump these packages so high. `createExperimentationService` was added in ui [0.36.1](https://github.com/microsoft/vscode-azuretools/wiki/vscode-azureextensionui-changelog#v0361---2020-10-22), which is all you really need.

You could also consider copying `createExperimentationService.ts` from the UI package to this repo so that we don't have to bump anything.

package.json
```diff
@@ -474,8 +474,8 @@
...
         "vscode-nls": "^4.1.1"
     },
```
ejizba, line 480: > Think it's worth downgrading npm so my package lock doesn't go crazy like this?

Yes

ejizba (changes requested):

package.json
nturinski: That seems like an easy enough solution.  Would you be comfortable with shipping that if that were to be the only change?

CHANGELOG.md
fiveisprime: There's no hard required date. Doing a full test pass makes sense to me. No big rush on this

package.json
ejizba: Doing a test pass would be the largest factor in terms of "would you be comfortable", but yes this would certainly help as well.

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Change Log
 
+## 0.5.1 - 2021-04-02
```
ejizba, line 3: As discussed offline, we have plenty of time to run an A/A test before our next actual release, so pick a date where no one (aka CTI) has to rush

ejizba (approved):

Issue #329, opened 2021-04-02
Allow users to fork repo if they try to create with a remote without admin permissions


nturinski:
Use this API call: https://docs.github.com/en/rest/reference/repos#create-a-fork

PR #335, opened 2021-04-07
Enforce having an opened workspace for SWA creation

nturinski:
Fixes https://github.com/microsoft/vscode-azurestaticwebapps/issues/341
Fixes https://github.com/microsoft/vscode-azurestaticwebapps/issues/101

A few things to note:

1. Forking will come later: https://github.com/microsoft/vscode-azurestaticwebapps/issues/329.  The PR was already getting chonky so I will add that next
2. ~~There are three steps that are now unused, namely CreateScenarioStep, RepoPickStep, and BranchPickStep.  However, I'm intentionally keeping them because @fiveisprime wants to do an A/B test with the 2 different creation scenarios, so I will have to add a code path back to them~~  All 3 steps were removed.  Discussed offline: an A/B experiment vs. the old and new experience won't yield metrics that will help drive the product.  It'd be more beneficial to do experiments on the parts of the new creation flow.
3. ~~Currently, the Advanced Creation is broken because it was relying on some of the old process flow (specifically CreateScenarioStep).  What do we want to do with Advanced Create?  The only difference today is that it allows you to choose the resourceGroup and pick the remote name.~~  Fixed.  Check out this comment https://github.com/microsoft/vscode-azurestaticwebapps/pull/335#issuecomment-824425733


src/commands/createStaticWebApp/createStaticWebApp.ts
```diff
@@ -22,6 +22,16 @@ export async function createStaticWebApp(context: IActionContext & Partial<ICrea
         node = await ext.tree.showTreeItemPicker<SubscriptionTreeItem>(SubscriptionTreeItem.contextValue, context);
     }
 
+    const folder: WorkspaceFolder = await getWorkspaceFolder(context);
+    const gitWorkspaceState: GitWorkspaceState = await getGitWorkspaceState(context, folder.uri);
+    // verifyRepoForCreation will set gitWorkspaceState.repo, but typing will yell if we don't set it here
```
wwlorey, line 27: What about changing `verifyRepoForCreation`'s return type to something like this? 
```ts 
type VerifiedGitWorkspaceState = GitWorkspaceState & { repo: Repository };
```
Then you could just set `gitWorkspaceState = await verifyRepoForCreation(...);`

src/commands/createStaticWebApp/createStaticWebApp.ts
```diff
@@ -22,6 +22,16 @@ export async function createStaticWebApp(context: IActionContext & Partial<ICrea
...
+    gitWorkspaceState.repo = await verifyRepoForCreation(context, gitWorkspaceState, folder.uri);
+
+    await promptForDefaultBranch(context, gitWorkspaceState.repo);
+    context.telemetry.properties.cancelStep = undefined;
```
wwlorey, line 31: I think it would be cleaner to put this line at the end of `promptForDefaultBranch` (I believe that would have the same effect?)

src/tree/EnvironmentTreeItem.ts
```diff
@@ -169,8 +168,7 @@ export class EnvironmentTreeItem extends AzureParentTreeItem implements IAzureRe
         const client: WebSiteManagementClient = await createWebSiteClient(this.root);
         this.data = await client.staticSites.getStaticSiteBuild(this.parent.resourceGroup, this.parent.name, this.buildId);
 
-        this.localProjectPath = getSingleRootFsPath();
```
wwlorey, line 172: This has to be set somewhere in `EnvironmentTreeItem` because `GitHubConfigGroupTreeItem` depends on it

src/tree/SubscriptionTreeItem.ts
```diff
@@ -51,13 +57,20 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
...
+            promptSteps.push(new GitHubOrgListStep(), new RepoNameStep(), new RepoPrivacyStep(), new RemoteShortnameStep());
+            executeSteps.push(new RepoCreateStep(), new GitignoreCreateStep());
+        }
 
         promptSteps.push(new StaticWebAppNameStep());
         if (context.advancedCreation) {
             promptSteps.push(new ResourceGroupListStep());
         } else {
-            const remoteRepo: ReposGetResponseData | undefined = await tryGetRemote(context);
+            const remoteRepo: ReposGetResponseData | undefined = await tryGetProjectForCreation(context);
```
wwlorey, line 73: If I have two git projects open in my workspace when creating and I select the second one, this error is thrown:

Screen Shot 2021-04-08 at 1 51 48 PM

src/utils/errors.ts
```diff
@@ -0,0 +1,10 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { localize } from "./localize";
+
+export class NoWorkspaceError extends Error {
```
wwlorey, line 8: Not sure if we should follow this exactly but this file was put directly in `src/` in the functions repo: https://github.com/microsoft/vscode-azurefunctions/blob/main/src/errors.ts

src/utils/gitUtils.ts
```diff
@@ -0,0 +1,121 @@
...
+import { MessageItem, Uri } from "vscode";
+import { DialogResponses, IActionContext } from "vscode-azureextensionui";
+import { IStaticWebAppWizardContext } from "../commands/createStaticWebApp/IStaticWebAppWizardContext";
+import { ext } from "../extensionVariables";
+import { getGitApi } from "../getExtensionApi";
+import { API, CommitOptions, Ref, Repository } from "../git";
+import { hasAdminAccessToRepo, tryGetRemote, tryGetReposGetResponseData } from "./gitHubUtils";
+import { localize } from "./localize";
+
+export type GitWorkspaceState = { repo: Repository | null, dirty?: boolean, remoteUrl?: string; hasAdminAccess?: boolean };
```
wwlorey, line 15: Is it safe to make `dirty` and `hasAdminAccess` required properties? They're always initialized as far as I can tell

src/utils/gitUtils.ts
```diff
@@ -0,0 +1,121 @@
...
+
+    return gitWorkspaceState;
+}
+
+export async function verifyRepoForCreation(context: IActionContext, gitWorkspaceState: GitWorkspaceState, uri: Uri): Promise<Repository> {
+    if (!gitWorkspaceState.repo) {
+        const gitRequired: string = localize('gitRequired', 'A GitHub repository is required to create a Static Web App. Would you like to initialize your project as a git repository and create a GitHub remote?');
+        context.telemetry.properties.cancelStep = 'initRepo';
+
+        await ext.ui.showWarningMessage(gitRequired, { modal: true }, DialogResponses.yes);
```
wwlorey, line 38: It would be slick if the button said "Initialize repo" instead of just "yes"

src/utils/gitUtils.ts
```diff
@@ -0,0 +1,121 @@
...
+        const gitApi: API = await getGitApi();
+        const newRepo: Repository | null = await gitApi.init(uri);
```
wwlorey, line 40: If this fails, is there a way to get more info about the error? It would be nice to direct that to the output channel for more detail

src/utils/gitUtils.ts
```diff
@@ -0,0 +1,121 @@
...
+        gitWorkspaceState.dirty = false;
+    }
+
+    return gitWorkspaceState.repo;
+}
+
+export async function promptForCommit(repo: Repository, value?: string): Promise<void> {
+    const commitPrompt: string = localize('commitPrompt', 'Enter a commit message.');
+    // VS Code doesn't handle nested git repos very well; it'll show changes as working, but is unable to commit
+    // so commit with { empty: true } so that we don't get errors when we making "empty" commits
```
wwlorey, line 69: nit: 
```suggestion
    // so commit with { empty: true } so that we don't get errors when making "empty" commits
```

src/utils/gitUtils.ts
```diff
@@ -0,0 +1,121 @@
...
+    const commitOptions: CommitOptions = { all: true, empty: true };
+
+    const commitMsg: string = await ext.ui.showInputBox({ prompt: commitPrompt, placeHolder: `${commitPrompt}..`, value });
```
wwlorey, line 72: I think it looks kinda redundant to have `commitPrompt` for both the prompt and placeholder. Maybe do something like `value: value || 'Add local changes'` instead of a placeholder?

Screen Shot 2021-04-08 at 3 33 44 PM
Screen Shot 2021-04-08 at 3 37 24 PM

src/utils/gitUtils.ts
wwlorey: Although I see why you just did "yes"... I missed the "and create a GitHub remote" part at the end. A button that says "Initialize repo and create remote" is probs too chonky

src/utils/gitUtils.ts
nturinski: Hmnm, I agree that having the prompt and placeholder being the same looks kind of redundant, but I also prefer just having no placeholder then (and force the user to write something) than just a generic "Add local changes".  If we do want to have a default commit message, I'd prefer something more like "Commit made by Azure Static Web Apps extension" or something that the user will know exactly where the commit came from.

wwlorey: I think "Commit made by Azure Static Web Apps extension" would be good. Just having something as a placeholder is nice. My main concern is it can be hard to come up with a commit message on the spot

src/commands/createStaticWebApp/createStaticWebApp.ts
nturinski: Yeah, though I put it within createStaticWebApp because the cancelStep is related to the `staticWebApps.createStaticWebApp` specifically.

src/utils/gitUtils.ts
nturinski: Haha, yeah true that though maybe just "Create" then?

src/utils/gitUtils.ts
nturinski: I actually haven't seen this fail yet, but I think it _might_ throw its own error if it does.  I have the check there as more of a fail safe/type safety, but I don't expect that we'll actually see many errors here.

src/utils/gitUtils.ts
wwlorey: Tbh I think "yes" is the way to go. Because if we just do "Create" then that doesn't capture that we're also initializing git

src/tree/SubscriptionTreeItem.ts
nturinski: Should be resolved by passing in the fspath

nturinski:
Removed all the unit tests for now, but they are all still here: https://github.com/microsoft/vscode-azurestaticwebapps/tree/nat/unitTests

src/commands/createRepo/RepoCreateStep.ts
```diff
@@ -43,7 +43,7 @@ export class RepoCreateStep extends AzureWizardExecuteStep<IStaticWebAppWizardCo
         const git: API = await getGitApi();
         const fsPath: string = nonNullProp(wizardContext, 'fsPath');
         const uri: Uri = Uri.file(fsPath);
-        let repo: Repository | null = git.getRepository(uri);
+        let repo: Repository | null = await git.openRepository(uri);
```
nturinski, line 46: See why the change here: https://github.com/microsoft/vscode/issues/111210

src/utils/gitUtils.ts
```diff
@@ -0,0 +1,145 @@
...
+    const commitOptions: CommitOptions = { all: true };
+
+    const commitMsg: string = await ext.ui.showInputBox({ prompt: commitPrompt, placeHolder: `${commitPrompt}..`, value });
+    try {
+        await repo.commit(commitMsg, commitOptions);
+    } catch (err) {
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+        if (!/nothing to commit/.test(err.stdout)) {
+            // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call
+            console.debug(parseError(err));
```
wwlorey, line 93: I didn't think that console logs worked from our extensions. Was this just for debugging?

src/utils/gitUtils.ts
```diff
@@ -0,0 +1,145 @@
...
+        const commitChanges: string = localize('commitChanges', 'Commit all working changes to create a Static Web App.');
+        await ext.ui.showWarningMessage(commitChanges, { modal: true }, { title: localize('commit', 'Commit') });
+        await promptForCommit(gitWorkspaceState.repo, localize('commitMade', 'Commit made from VS Code Azure Static Web Apps'));
+        gitWorkspaceState.dirty = false;
+    }
+
+    return <VerifiedGitWorkspaceState>gitWorkspaceState;
+}
+
+export async function promptForCommit(repo: Repository, value?: string): Promise<void> {
```
wwlorey, line 82: nit: No need to export. Actually does it really matter too much if we export something that doesn't need to be exported?

src/utils/gitUtils.ts
```diff
@@ -0,0 +1,145 @@
...
+        if (result === checkoutButton) {
+            await repo.checkout(defaultBranch);
+            context.telemetry.properties.checkoutDefault = 'true';
+        } else {
+            context.telemetry.properties.checkoutDefault = 'false';
+        }
+    }
+}
+
+export async function tryGetDefaultBranch(repo: Repository): Promise<string | undefined> {
```
wwlorey, line 120: Also no need to export this guy

src/utils/gitUtils.ts
```diff
@@ -0,0 +1,145 @@
...
+            defaultBranches.push(await repo.getGlobalConfig('init.defaultBranch'));
+        } catch (err) {
+            // VS Code's git API doesn't fail gracefully if no config is found, so swallow the error
+        }
+    }
+
+    const localBranches: Ref[] = await repo.getBranches({ remote: false });
+    // order matters here because we want the setting, main, then master respectively so use indexing
+    for (let i = 0; i < localBranches.length; i++) {
+        if (localBranches.some(lBranch => lBranch.name === defaultBranches[i])) {
```
wwlorey, line 138: Since `localBranches` and `defaultBranches` aren't guaranteed to be the same length I don't think it's safe to use `i` to index into both

test/global.test.ts
```diff
@@ -17,7 +16,7 @@ export function createTestActionContext(): IActionContext {
 }
 
 // Runs before all tests
-suiteSetup(async function (this: IHookCallbackContext): Promise<void> {
+suiteSetup(async function (this: Mocha.Context): Promise<void> {
```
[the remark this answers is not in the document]
nturinski: The IHookCallbackContext is actually deprecated, so I figured I'd go ahead and just update this now.

src/utils/gitUtils.ts
nturinski: Probably not.  It does make it harder to determine if we have unused code.  If something is exported, no error is thrown if a function isn't being used.  

I guess also if I didn't want it accessible outside of the file, that'd obviously be a problem.

src/utils/gitUtils.ts
nturinski: Good catch.  Meant to use defaultBranches.length as the for loop condition.

src/utils/gitUtils.ts
```diff
@@ -0,0 +1,143 @@
...
+/* Function used to enforce all our opinionated requirements for creating a Static Web App */
+export async function verifyGitWorkspaceForCreation(context: IActionContext, gitWorkspaceState: GitWorkspaceState, uri: Uri): Promise<VerifiedGitWorkspaceState> {
+    const gitFolderName: string = '.git';
+    if ((await fse.readdir(uri.fsPath)).filter(file => file !== gitFolderName).length === 0) {
+        // if the git repo is empty, git push will fail and create an empty GitHub repo so throw an error here
+        throw new Error(localize('emptyWorkspace', 'Cannot create a Static Web App with an empty workspace.'));
+    }
+
+    if (!gitWorkspaceState.repo) {
+        const gitRequired: string = localize('gitRequired', 'A GitHub repository is required to create a Static Web App. Would you like to initialize your project as a git repository and create a GitHub remote?');
```
wwlorey, line 45: I liked Matt's suggestion to switch up the wording here so it isn't in the form of a question since people don't have a choice

src/utils/gitUtils.ts
```diff
@@ -0,0 +1,143 @@
...
+        context.telemetry.properties.cancelStep = 'initRepo';
+
+        await ext.ui.showWarningMessage(gitRequired, { modal: true }, DialogResponses.yes);
+        const gitApi: API = await getGitApi();
+        let newRepo: Repository | null = await gitApi.init(uri);
+        if (!newRepo) {
+            const localGit: git.SimpleGit = git(uri.fsPath);
+            await localGit.init({});
```
wwlorey, line 53: Is `localGit.init` just a fallback for if `gitApi.init` fails? I'm just curious why both versions are needed

src/utils/gitUtils.ts
```diff
@@ -0,0 +1,143 @@
...
+async function tryGetDefaultBranch(repo: Repository): Promise<string | undefined> {
+    // currently git still uses master as the default branch but will be updated to main so handle both cases
+    // https://about.gitlab.com/blog/2021/03/10/new-git-default-branch-name/#:~:text=Every%20Git%20repository%20has%20an,Bitkeeper%2C%20a%20predecessor%20to%20Git.
+    const defaultBranches: string[] = ['main', 'master'];
+    try {
+        defaultBranches.unshift(await repo.getConfig('init.defaultBranch'));
+    } catch (err) {
+        // if no local config setting is found, try global
+        try {
+            defaultBranches.push(await repo.getGlobalConfig('init.defaultBranch'));
```
wwlorey, line 127: I believe this should be `defaultBranches.unshift` as well

src/utils/gitUtils.ts
nturinski: Yeah, I was more experimenting with that in testing.  Honestly, it's probably unnecessary so I'll remove it.

nturinski:
> Currently, the Advanced Creation is broken because it was relying on some of the old process flow (specifically CreateScenarioStep). What do we want to do with Advanced Create? The only difference today is that it allows you to choose the resourceGroup and pick the remote name.

To address this, some changes have been made to the creation flow.
1. If the user must create a new GitHub repo, we assume the authenticated user as their org.  If they need to select a different org, advanced create will allow them to choose.  This is similar to the Git extension's `git clone` command that only populates a list with the authenticated user's repositories.
2. If the user is creating a new repo, default to private.
3. The api step is assumed to be `api` without prompting the user.  We know that less than 20% of users are actually using Functions in SWA anyway, and it's a confusing prompt.  If the user generates a functions project later, it'll default to `api` and all of the build presets default to `api` anyway. Lastly, having a value of `api` doesn't break the user's build, so it should be relatively safe to set for the user.
4. Location is being defaulted to `Central US`.  We may be able to do some smarter logic with this later.

Advanced create also enforces that a workspace be opened and a clean git repo.

Because of all the verification we're doing prior to the first prompt, there was a pretty large delay between hitting "+" and the first prompt appearing.  To help combat that, I've moved a lot of the async code into `createStaticWebApp.ts` to preset as much as we can before prompting the user.  

The subscription node currently spins while we're verifying the user's project which is somewhat obvious _if_ your entry point was the plus or right-click context menu.  F1 may be less obvious-- other alternatives could be a spinning wheel cursor or a long-running notification.

src/commands/createStaticWebApp/setWorkspaceContexts.ts
```diff
@@ -0,0 +1,18 @@
...
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as fse from 'fs-extra';
+import { join } from "path";
+import { IActionContext } from "vscode-azureextensionui";
+import { remoteShortnameExists } from "../../utils/gitHubUtils";
+import { IStaticWebAppWizardContext } from "./IStaticWebAppWizardContext";
+
+export async function setWorkspaceContexts(wizardContext: IActionContext & Partial<IStaticWebAppWizardContext>, fsPath: string): Promise<void> {
```
wwlorey, line 12: This func is only used in `createStaticWebApp.ts`, any reason you didn't put it there?

src/errors.ts
```diff
@@ -0,0 +1,59 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+import { GitErrorCodes } from "./git";
+import { localize } from "./utils/localize";
+
+export class NoWorkspaceError extends Error {
+    public message: string = localize('noWorkspaceError', 'You must have a workspace open to perform this operation.');
+}
+
+// copied from https://github.com/microsoft/vscode/blob/main/extensions/git/src/commands.ts#L2800-L2863
```
wwlorey, line 13: nit: I think it would be good to use a permalink for this URL in case the line numbers change: https://github.com/microsoft/vscode/blob/779434d2d118889e2a5a2113714ad6c8bcb3a6e3/extensions/git/src/commands.ts#L2800-L2863

wwlorey (approved):

src/commands/createStaticWebApp/GitHubOrgListStep.ts
```diff
@@ -27,17 +27,29 @@ export class GitHubOrgListStep extends AzureWizardPromptStep<IStaticWebAppWizard
     }
 
     public shouldPrompt(context: IStaticWebAppWizardContext): boolean {
-        return !context.repoHtmlUrl || context.createScenario === 'publishToNewRepo';
+        // if orgData was set earlier, we should push the value to mask
+        if (context.orgData) {
+            context.valuesToMask.push(context.orgData.login);
```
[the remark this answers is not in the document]
nturinski: Yep, both are needed.  On L26, that's happening if the user selected something during the `prompt` part of the step.  But L32 is happening in the `shouldPrompt` part of the step, which executes prior to `prompt` to see if `prompt` should execute or not.

The reason for this change is because we default the user to their AuthenticatedUser in basic create rather than letting them select an org, so if we auto-default, we should still push the username into `valuesToMask`.

I could have done this at the time of setting orgData, but I thought it was more appropriate here.

src/commands/createStaticWebApp/setWorkspaceContexts.ts
nturinski: Yeah, I prefer just giving commands their own file rather than cluttering files with helper functions.  It also used to be part of `WorkspaceListStep` so I just thought it'd be better as a standalone command and in the same folder (`createStaticWebApp`) as the other create helper files.

src/commands/createStaticWebApp/GitHubOrgListStep.ts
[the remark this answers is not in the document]
wwlorey: I see - I was concerned that it would double-push to `valuesToMask` but I see now that that won't happen

src/errors.ts
nturinski:
```suggestion
// copied from https://github.com/microsoft/vscode/blob/779434d2d118889e2a5a2113714ad6c8bcb3a6e3/extensions/git/src/commands.ts#L2800-L2863
```

Issue #98, opened 2020-05-18
Add snippets for adding fallback route to  routes.json


johnpapa:

nturinski:
Hi @johnpapa

Could you give me some more details about what the feature request is?  I'm not really sure what this entails.

johnpapa:
Sure.  The routes.json file is in our docs at https://aka.ms/swadocs 

We can add a fallback route to this array. The fallback will catch 404 server errors and serve the index.html, commonly. 

I could provide the snippet if it is helpful.

nturinski:
Yes, the snippet would be helpful 😄 

My experience with SWA is still pretty limited so any bigger examples beyond the ReactSite repo would also be greatly appreciated!

johnpapa:
Here you go. Sorry for the delay

```
	"SWA Fallback Routes": {
		"prefix": "swa-fallback",
		"body": [
			"{$0",
			"\t\"routes\": [",
			"\t\t{",
			"\t\t\t\"route\": \"/*\",",
			"\t\t\t\"serve\": \"/index.html\",",
			"\t\t\t\"statusCode\": 200",
			"\t\t}",
			"\t]",
			"}",
		]
	},
```

aartigh:
Hi Team,

My reactjs app is deployed in Azure Static Web App(preview).

My website url : www.example.com shows all images render properly from public folder. 
From home page, if user navigate to https://www.example.com/nextpage/page1 then still it shows all the images properly.

But if user, refresh or navigate to https://www.example.com/nextpage/page1 directly then images are not rendered, gives 404 error. 

my routes.json file in public folder looks like below

{
  "routes": [
    {
      "route": "/*",
      "serve": "/index.html",
      "statusCode": 200
    }
  ]
}

Appreciate your help.

anthonychu:
@aartigh can you please file an issue at https://github.com/Azure/static-web-apps/issues ?

Please include info on how the `img` tag is formed and the error in your browser dev tools. If you have a live site that is public and we can troubleshoot, please include that too. Thanks!

aartigh:
@anthonychu thanks, if you access https://www.skillsandautomation.com/blog/B1 url directly, you can see image at top left not rendered, gives 404 error in dev tool. Whereas if we access the https://www.skillsandautomation.com, image at top left is rendered correctly.

aartigh:
@anthonychu my issue is resolved, i have change way of specifying that image path.

[...]

Issue #97, opened 2020-05-18
Add staticwebapp.config.json command


johnpapa:
Create a command that adds the staticwebapp.config.json file.

nturinski:
Similar to this issue: https://github.com/microsoft/vscode-azurestaticwebapps/issues/94, we feel like creating the routes.json for the user would be pretty unmaintainable since it will be framework specific when they deploy the SWA.

Also similarly, we should probably just direct them to another recipes page on how to set up their routes depending on which framework they are using.

johnpapa:
It would be easy to set up a default rule in a snippet that they can enter the route and the page to return. Those could default to index.html but be a placeholder in a snippet. 

A snippet is far better experience than going to docs. Creating the file is even better. 

Don’t need to create the file automatically but a command to add it would be super friendly.

anthonychu:
@wwlorey FYI we also added the schema that should help with intellisense in the file: https://github.com/SchemaStore/schemastore/pull/1551

> @nturinski: Similar to this issue: #94, we feel like creating the routes.json for the user would be pretty unmaintainable since it will be framework specific when they deploy the SWA.

This is much better now. The recommendation is for staticwebapp.config.json to be at the root fold of their frontend app, but we will search for it in any subfolder.

@johnpapa Is the intention of this to simplify adding a fallback route? For that, it's such a common task that I think we should add an explicit gesture.

I suggest changing this issue to add an "Add default fallback route" command. Something like:
* Ask for the root of the frontend app
* If staticwebapp.config.json doesn't exist in the folder or subfolders, add one at the root with the default fallback route
* If staticwebapp.config.json exists
    * If there's no `navigationFallback` section, add the default
    * If there's already a `navigationFallback` section, error
* If more than one staticwebapp.config.json exists, this is not good in general
* Maybe support a context menu item for this command when right-clicking a folder or existing staticwebapp.config.json

Default fallback route:
```json
  "navigationFallback": {
    "rewrite": "/index.html"
  }
```

johnpapa:
Good question @anthonychu 

I think a fallback is a big reason this would be nice. So yes, your plan could be very helpful. 

Could these both be good thought? user wants to create routes of any kind. They find a command that creates the file (empty) and the schema is applied. User wants to add a 401 page, or a fallback route, or some other common routing logic ... this could be solved with commands or snippets.

anthonychu:
[...]

PR #339, opened 2021-04-07
Add staticwebapp.config.json schema

wwlorey:
Related to https://github.com/microsoft/vscode-azurestaticwebapps/issues/97
Fixes https://github.com/microsoft/vscode-azurestaticwebapps/issues/98

### DIFF
--- a/package.json
+++ b/package.json
@@ -383,6 +383,12 @@
                 }
             ]
         },
+        "jsonValidation": [
+            {
+                "fileMatch": "staticwebapp.config.json",
+                "url": "https://json.schemastore.org/staticwebapp.config.json"
+            }
+        ],
         "configuration": [
             {
                 "title": "Azure Static Web Apps",


nturinski (approved):
This is really awesome/helpful 👍

package.json
```diff
@@ -383,6 +383,12 @@
                 }
             ]
         },
+        "jsonValidation": [
+            {
+                "fileMatch": "staticwebapp.config.json",
```
nturinski, line 388: Change this to 
`/staticwebapp.config.json`

Eric has it this way in Functions, I think it's so it doesn't _have_ to be in the root.

nturinski, line 388: Eric has it like `/function.json` in the Functions repo, but I'm not really sure what the purpose/reason is.  @ejizba Is there any reason that we should fileMatch with a slash prefix?

ejizba: I don't remember. Only thing I can think of is maybe the slash prevents something like "test.function.json" from being affected - but I tested and doesn't seem to make a difference.

wwlorey:
I just tested on my end as well - the file was recognized in a subdirectory or in the root dir

PR #342, opened 2021-04-07
Use VS Code's product icon theme

ejizba:
Some examples:
Screen Shot 2021-04-07 at 4 05 02 PM
Screen Shot 2021-04-07 at 4 05 22 PM
Screen Shot 2021-04-07 at 4 09 52 PM
Screen Shot 2021-04-07 at 4 10 22 PM
Screen Shot 2021-04-07 at 4 10 37 PM


src/utils/actionUtils.ts
```diff
@@ -5,16 +5,43 @@
...
 import { localize } from "./localize";
-import { treeUtils } from "./treeUtils";
 
 export function getActionIconPath(data: ActionWorkflowStepData | ActionsGetJobForWorkflowRunResponseData | ActionsGetWorkflowRunResponseData): TreeItemIconPath {
-    return data.conclusion !== null ?
-        treeUtils.getThemedIconPath(path.join('conclusions', ensureConclusion(data))) :
-        treeUtils.getThemedIconPath(path.join('statuses', ensureStatus(data)));
+    let id: string;
+    if (data.conclusion !== null) {
+        switch (ensureConclusion(data)) {
```
nturinski, line 16: So the only thing is that we were using icons to try to match GitHub's (the coloring, specifically).

image


image

I'm not opposed to changing this, but I was wondering if there was a certain "branding" we wanted to maintain with the GitHub related icon.

nturinski (approved):

src/utils/actionUtils.ts
```diff
@@ -5,16 +5,43 @@
...
+            case Conclusion.Cancelled:
+                id = 'circle-slash';
+                break;
+            case Conclusion.Failure:
+                id = 'error';
+                break;
+            case Conclusion.Skipped:
+                id = 'debug-step-over';
```
wwlorey, line 24: My only issue is this icon Screen Shot 2021-04-08 at 11 39 23 AM

 It reminds me too much of liveshare lol. I do prefer the GitHub icons here but if we have to use the codicons maybe we could match them more closely to what the GitHub icons were?

ejizba: You're welcome to suggest an alternative: https://code.visualstudio.com/api/references/icons-in-labels

I chose that one based on this entry:
```
testing-skipped-icon | debug-step-over | Icon shown for tests that are skipped.
```

src/utils/actionUtils.ts
ejizba: I don't think status icons fit under "branding", so yes I think we need to use VS Code icons from an accessibility/themes perspective. I would say branding is only for things like "Azure Function App" or "Azure Subscription" where it's a concept specific to that brand. In this case, "Failed" or "Skipped" is not a concept specific to GitHub

src/utils/actionUtils.ts
wwlorey: How about `stop` Screen Shot 2021-04-08 at 11 50 47 AM for `Conclusion.Failure` and
`circle-slash` Screen Shot 2021-04-08 at 11 51 05 AM for `Conclusion.Skipped`?

ejizba: Pretty sure "error" is the exact same as "stop" and I'm already using "circle-slash" for "Conclusion.Cancelled"

> I do prefer the GitHub icons here

I don't disagree, but I think we have to use VS Code themes for accessibility reasons and to satisfy all users. Ultimately, you're welcome to make your own product icon theme using the existing GitHub icons to replicate the old behavior :P

wwlorey (approved):

src/utils/actionUtils.ts
nturinski: LOL is that last comment what pushed Will to just approve the PR? 😋

wwlorey: download

Issue #337, opened 2021-04-07
Allow changes to presets


fiveisprime:
This is something we talked about a couple of weeks ago.

* Customer chooses a preset
* Prompt for the prefilled value with the text selected
* Do this for both App Location and API Location

PR #343, opened 2021-04-08
Only pre-generate values for build presets

nturinski:
Fixes https://github.com/microsoft/vscode-azurestaticwebapps/issues/337

`manuallyEnter` effectively was replaced with `Custom` in telemetry.  Because I added telemetry to both `EnterApi/AppLocationSteps`, it was getting overwritten after a user submitted their value. 

I think there is value to see if the our pre-generated values were used, so adding telemetry in the input steps makes sense. The only way to `manuallyEnter` is through `Custom` anyway, so it's a good enough telemetry point to see if users are manually entering their own values.


src/commands/createStaticWebApp/AppLocationStep.ts
```diff
@@ -21,7 +21,7 @@ export class AppLocationStep extends AzureWizardPromptStep<IStaticWebAppWizardCo
     }
 
     public shouldPrompt(wizardContext: IStaticWebAppWizardContext): boolean {
-        return !wizardContext.appLocation;
+        return !wizardContext.appLocation && !wizardContext.presetAppLocation;
```
ejizba, line 24: I'm a bit confused why we even have `AppLocationStep` at this point. When does it actually prompt - and is that functionality worth it? Feels like it might simplify the code and make the step count more accurate if we always use the `EnterAppLocationStep` version instead.

nturinski: We still use it for "Custom".  At that point, the user can "Manually Enter" or click one of their folder directories from the GitHub tree we pull from their repo or local project.

The idea with showing the folder directories was to help make it obvious that we were trying to get the user to point us to their app code in their project.

I'm not opposed to removing `AppLocationStep` though and just relying on the user to input it themselves (with smart defaults either by presets or my detection-logic)

ejizba: Do we have telemetry on how many people are selecting "Custom" and actually clicking a folder (aka _not_ "manually enter")? Unless it's a high percentage, my personal vote is to get rid of the quick pick steps. We expect most users to pick a preset, right? So if the default experience is "manually enter" in the preset case, I feel like it would be more consistent to also use "manually enter" for the "custom" case - plus it simplifies the code.

nturinski: Discussed offline:  We will remove the quickpick steps and just default to an input box for App and Api location steps now.

nturinski:
I'll rename the files in a different PR.  I tried to do it in here and it made the diff all funky since the files I deleted used the names that I wanted to use.

ejizba (approved):

PR #345, opened 2021-04-08
Never show local project tree item

wwlorey:
The local project tree item has been stripped down and it this point it doesn't even render any children. So I don't think it should show up even with the feature flag enabled

### DIFF
--- a/src/tree/AzureAccountTreeItemWithProjects.ts
+++ b/src/tree/AzureAccountTreeItemWithProjects.ts
@@ -3,49 +3,15 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { pathExists } from 'fs-extra';
-import { join } from 'path';
-import { workspace } from 'vscode';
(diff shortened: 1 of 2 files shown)


src/tree/AzureAccountTreeItemWithProjects.ts
```diff
@@ -3,49 +3,15 @@
...
  *--------------------------------------------------------------------------------------------*/
 
-import { pathExists } from 'fs-extra';
-import { join } from 'path';
-import { workspace } from 'vscode';
-import { AzExtTreeItem, AzureAccountTreeItemBase, IActionContext, ISubscriptionContext } from 'vscode-azureextensionui';
-import { configFileName, enableLocalProjectView } from '../constants';
-import { getWorkspaceSetting } from '../utils/settingsUtils';
-import { LocalProjectTreeItem } from './localProject/LocalProjectTreeItem';
+import { AzureAccountTreeItemBase, ISubscriptionContext } from 'vscode-azureextensionui';
```
wwlorey, line 6: Undoing the changes made here: https://github.com/microsoft/vscode-azurestaticwebapps/pull/295/files#diff-924234256676716751c864b426f5c007a88a1797571f3f1f469deb07272a70dc

ejizba (approved):

Issue #349, opened 2021-04-12
The updated config value without a "" is not highlight after clicking the config tree item


v-mengwe:
**OS:** All
**Build Version:** 20210410.1

**Repro Steps:**

1. Add the following to setting.json: `"staticWebApps.enableLocalProjectView": true`.
2. Create a static web app using the local project -> Expand the static web app -> Production -> GitHub Configuration -> Click a "GitHub Configuration" child node -> Edit the config value in the opened .yml file, e.g. "output_location":
   ![image](https://user-images.githubusercontent.com/62053665/114371982-a5520800-9bb3-11eb-80f6-9723d0489384.png)
4. CRTL + S -> Refresh the corresponding "GitHub Configuration" -> The "output_location" value changes to "test" -> Click the "output_location".
5. Check whether the value of the "output_location" is highlight.

**Expect:**
The value of the "output_location" is highlight.

**Actual:**
The value of the "output_location" is not highlight.

PR #351, opened 2021-04-12
Use `yaml` package when selecting build configs

wwlorey:
Yaml values without quotes are supported now as well:
Screen Shot 2021-04-21 at 12 34 22 PM

Fixes https://github.com/microsoft/vscode-azurestaticwebapps/issues/349


src/commands/openYAMLConfigFile.ts
```diff
@@ -47,12 +47,13 @@ export async function openYAMLConfigFile(context: IActionContext, node?: StaticW
...
-    offset += `${buildConfigToSelect}: `.length;
+    if (offset === -1) {
+        return undefined;
+    }
 
     const position: Position = configDocument.positionAt(offset);
-    const configValueRegex: RegExp = /['"].*['"]/;
-    return configDocument.getWordRangeAtPosition(position, configValueRegex);
+    const line: TextLine = configDocument.lineAt(position);
+    return line.range;
```
ejizba, line 58: We might continue to have edge-case bugs with this unless we use a yaml parser. Can you double check to see if `yaml` or `js-yaml` (which seems to be more widely used) gives us the position info in one of their methods? You might have to explore outside the default "parse" method.

src/commands/openYAMLConfigFile.ts
```diff
@@ -45,14 +47,67 @@ export async function openYAMLConfigFile(context: IActionContext, node?: StaticW
...
+            const range = <number[] | undefined>configValue.range;
+
+            if (range) {
+                const buildConfigOffset: number = configDocumentText.search(buildConfigRegex);
+                let newlines: number = 0;
+
+                // The range returned from the `yaml` package doesn't include newlines
+                // So count newlines and include them in the range we return
+                for (const char of configDocumentText.slice(0, buildConfigOffset)) {
+                    newlines += char === '\n' ? 1 : 0;
```
neelip, line 78: There seems to be another way to do this without the for-loop. 
Will, you might have already looked at this, but wondering if we can use this method instead? 
https://github.com/eemeli/yaml/issues/127#issuecomment-628184035

Also, should we track this issue? In case there is an improvement from the yaml side.

src/commands/openYAMLConfigFile.ts
```diff
@@ -45,14 +47,67 @@ export async function openYAMLConfigFile(context: IActionContext, node?: StaticW
...
-    offset += `${buildConfigToSelect}: `.length;
+    if (buildConfigMatches && buildConfigMatches.length > 1) {
+        void ext.ui.showWarningMessage(localize('foundMultipleBuildConfigs', 'Multiple "{0}" build configurations were found in "{1}".', buildConfigToSelect, basename(configDocument.uri.fsPath)));
+        return undefined;
+    }
+
+    const parsedYaml: Document.Parsed = parseDocument(configDocumentText);
+    type YamlNode = YAMLMap | YAMLSeq | Pair | undefined;
+    const yamlNodes: YamlNode[] = [];
+    let yamlNode: YamlNode = parsedYaml.get('jobs');
```
ejizba, line 64: No need to hard-code "jobs" if we just use the root "contents"
```suggestion
    type YamlNode = YAMLMap | YAMLSeq | Pair | Scalar | undefined | null;
    const yamlNodes: YamlNode[] = [];
    let yamlNode: YamlNode = parsedYaml.contents;
```

src/commands/openYAMLConfigFile.ts
```diff
@@ -45,14 +47,67 @@ export async function openYAMLConfigFile(context: IActionContext, node?: StaticW
...
+
+    while (yamlNode) {
+        if (yamlNode.hasOwnProperty('key') && yamlNode['key'].value === buildConfigToSelect && yamlNode.hasOwnProperty('value')) {
+            const configValue = <Scalar>yamlNode['value'];
+            const range = <number[] | undefined>configValue.range;
```
ejizba, line 69: Fixing lint errors

```suggestion
        if ('key' in yamlNode && (<Scalar>yamlNode.key).value === buildConfigToSelect && 'value' in yamlNode) {
            const configValue = <Scalar>yamlNode.value;
            const range = configValue.range;
```

src/commands/openYAMLConfigFile.ts
```diff
@@ -45,14 +47,67 @@ export async function openYAMLConfigFile(context: IActionContext, node?: StaticW
...
+
+                const startPosition = configDocument.positionAt(startOffset);
+                const endPosition = configDocument.positionAt(endOffset);
+                return new Range(startPosition, endPosition);
+            }
+        } else if (yamlNode.hasOwnProperty('items')) {
+            yamlNodes.push(...yamlNode['items'])
+        } else if (yamlNode.hasOwnProperty('value') && yamlNode['value'].hasOwnProperty('items')) {
+            yamlNodes.push(...yamlNode['value']['items'])
+        }
```
ejizba, line 107: Fixing lint errors and seems easier to pop `yamlNode.value` onto the queue rather than dig into `items` in multiple places

```suggestion
        } else if ('items' in yamlNode) {
            yamlNodes.push(...yamlNode.items);
        } else if ('value' in yamlNode && typeof yamlNode.value === 'object') {
            yamlNodes.push(yamlNode.value)
        }
```

ejizba (commented):
I'm generally only okay with disabling lint on a per-line basis. In this case, we can just fix the lint problems instead of disabling them - see suggestions

src/commands/openYAMLConfigFile.ts
```diff
@@ -45,14 +47,67 @@ export async function openYAMLConfigFile(context: IActionContext, node?: StaticW
     await window.showTextDocument(configDocument, { selection });
 }
 
-async function getSelection(configDocument: TextDocument, buildConfigToSelect: BuildConfig): Promise<Range | undefined> {
-    const configRegex: RegExp = new RegExp(`${buildConfigToSelect}:`);
+/* eslint-disable @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-member-access, no-prototype-builtins, @typescript-eslint/no-unsafe-assignment */
+export async function getSelection(configDocument: TextDocument, buildConfigToSelect: BuildConfig): Promise<Range | undefined> {
```
ejizba, line 51: Since the code is a bit complex and has some sketchy "any" types, I'd probably rename this to `tryGetSelection` and add a big try/catch that just returns `undefined` if an error happens. This function already allows `| undefined` in the return type anyways

test/selectBuildConfigs.test.ts
```diff
@@ -0,0 +1,63 @@
...
+        { workflowFileName: 'workflow-duplicates.yml', buildConfig: 'output_location', expectedSelection: undefined },
+        { workflowFileName: 'workflow-duplicates.yml', buildConfig: 'app_artifact_location', expectedSelection: undefined },
+    ];
+
+    for (const testCase of testCases) {
+        const title: string = `${testCase.workflowFileName}: ${testCase.buildConfig}`;
+        const workspacePath: string = getWorkspacePath('testWorkspace');
+
+        test(title, async () => {
+            const configDocument: TextDocument = await workspace.openTextDocument(join(workspacePath, 'testWorkflows', testCase.workflowFileName));
```
ejizba, line 50: Rather than use separate yml files in the "testWorkflows" directory, I'd prefer you just put the test data in this file (like, as a `const`), for a few reasons:
1. Using the file system and/or VS Code APIs can slow tests down. It's worth it for end-to-end tests, but overkill for unit tests IMO
2. You can easily manipulate the test data with template literals and whatnot
3. Idk just seems easier to read when the test data is in the same file.

I feel like the only downside is this file can get pretty long, but it just doesn't bother me 🤷‍♂️. One way to help with that is to scope your test data down to smaller cases. For example, you might have one test running against an "actual" yml file seen in the wild, but the other tests are usually just testing specific edge cases without the full "yml" file necessary.

src/commands/openYAMLConfigFile.ts
wwlorey: Thanks for the suggestion - now this is using only the `yaml` package to get ranges. I don't think we need to track the issue though. From the looks of it they've already rejected the feature request to include newline counts in ranges

test/selectBuildConfigs.test.ts
```diff
@@ -0,0 +1,157 @@
...
+            const selection: Range | undefined = await tryGetSelection(configDocument, testCase.buildConfig);
+            let expectedSelection: Range | undefined;
+
+            if (testCase.expectedSelection) {
+                const expectedStart: Position = new Position(testCase.expectedSelection.line, testCase.expectedSelection.startChar);
+                const expectedEnd: Position = new Position(testCase.expectedSelection.line, testCase.expectedSelection.endChar);
+                expectedSelection = new Range(expectedStart, expectedEnd);
+            }
+
+            assert.ok(expectedSelection && selection?.isEqual(expectedSelection) || selection === expectedSelection, 'Actual and expected selections do not match');
```
nturinski, line 66: You should be able to use `assert.strictEquals()` here.  Using `assert.ok` seems kind of roundabout here.

test/selectBuildConfigs.test.ts
```diff
@@ -0,0 +1,157 @@
...
+    name: Close Pull Request Job
+    steps:
+      - name: Close Pull Request
+        id: closepullrequest
+        uses: Azure/static-web-apps-deploy@v0.0.1-preview
+        with:
+          azure_static_web_apps_api_token: $\{{ secrets.AZURE_STATIC_WEB_APPS_API_TOKEN_AMBITIOUS_ROCK_0D992521E }}
+          action: "close"
+
+`];
```
nturinski, line 157: Could we try location paths with special characters and maybe some foreign characters? (Chinese or Korean or something seems plausible)

test/selectBuildConfigs.test.ts
wwlorey: I went with `assert.ok` because comparing two ranges uses their `isEqual()` function. I updated the assert to hopefully make it look more purposeful

nturinski: Ahh, okay, makse sense.

nturinski (approved):

Issue #323, opened 2021-03-26
Fail to create a HTTP trigger with .NET 5


v-mengwe:
**OS:** Win10
**Build Version:** 20210325.1
**Function Extension:** 20210325.4

**Repro Steps:**

1. Open an empty folder in VS Code -> Click "Create HTTP Function..." button -> Select C# -> Select .NET 5.
2. Check whether create a HTTP trigger with .NET 5 successfully.

**Expect:**
Succeed to create a HTTP trigger with .NET 5.

**Actual:**
Fail to create a HTTP trigger with .NET 5.
![image](https://user-images.githubusercontent.com/62053665/112571693-8fa6b980-8e23-11eb-82ab-040b83807c9b.png)

nturinski:
@anthonychu Are there any plans for SWA to support .NET 5 triggers (or does it already)?  Trying to figure out if we need to hide this option or figure out why this creation is failing.

anthonychu:
Good point. No immediate plans. The supported Functions runtime versions will remain:

- .NET Core 3.1
- Node.js 12
- Python 3.8

anthonychu:
@fiveisprime Can we bump this up higher in the milestone to ensure it can make it into this release? We don't want folks creating .NET 5 apps.

nturinski:
Will be fixed by using the new API in this PR: https://github.com/microsoft/vscode-azurefunctions/pull/2773/files/4908212f6e918ac7e207a016d5aff1107830ff88..bcde4acbd2131dddde8d42b99e238754cad7376b

PR #359, opened 2021-04-23
Use new functions api to select .NET3

nturinski:
Fixes https://github.com/microsoft/vscode-azurestaticwebapps/issues/323

Relies on Functions releasing with https://github.com/microsoft/vscode-azurefunctions/pull/2773/files/4908212f6e918ac7e207a016d5aff1107830ff88..bcde4acbd2131dddde8d42b99e238754cad7376b merged

### DIFF
--- a/src/commands/createHttpFunction.ts
+++ b/src/commands/createHttpFunction.ts
@@ -30,6 +30,6 @@ export async function createHttpFunction(context: IActionContext): Promise<void>
         templateId: 'HttpTrigger',
         languageFilter: /Python|C\#|^(Java|Type)Script$/i,
         functionSettings: { authLevel: 'anonymous' },
-        targetFramework: 'netcoreapp3.1'
+        targetFramework: 'netcoreapp3.1' // Will only work on functions api v1.4.0, but won't hurt on v1.3.0
     });
 }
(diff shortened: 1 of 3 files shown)


nturinski:
Purposely am not checking for v1.4.0 of the api though.  This way we'll just pick up the changes whenever Functions releases, but won't be broken in the meanwhile.

src/commands/createHttpFunction.ts
```diff
@@ -29,6 +29,7 @@ export async function createHttpFunction(context: IActionContext): Promise<void>
         suppressCreateProjectPrompt: true,
         templateId: 'HttpTrigger',
         languageFilter: /Python|C\#|^(Java|Type)Script$/i,
-        functionSettings: { authLevel: 'anonymous' }
+        functionSettings: { authLevel: 'anonymous' },
+        targetFramework: 'netcoreapp3.1'
```
ejizba, line 33: Can you add a comment in the code related to what you said on GitHub. Something like:

```suggestion
        targetFramework: 'netcoreapp3.1' // Will only work on functions api v1.4.0, but won't hurt on v1.3.0
```

ejizba (approved):

Issue #235, opened 2020-11-03
The setting name doesn't change automatically after renaming it


v-xinda:
**OS:** Win10
**Build Version:** 20201103.1

**Repro Steps:**

1. Right click a setting under the "Application Setting" node -> Click "Rename Setting..." -> Enter "test" as the new name.
2. Check the result.

**Expect:**
The setting name changes automatically without any error.

**Actual:**
The setting name doesn't change automatically and an error pops up after waiting about a minute.
![image](https://user-images.githubusercontent.com/57662131/97962673-a3a83b00-1df0-11eb-891e-e2c9fa53de3a.png)
Error type: 500
Error Message: An error has occurred.

**More Info:**
1. This issue also reproduces when deleting/adding/editing the setting.
2. This setting name changes automatically after refreshing.

nturinski:
Is this consistent?  I can't get this to repro on any of my static web apps.

v-xinda:
Firstly, need to add a new setting and name with special characters. Then the above issue reproduces when editing/deleting the other exist settings, you need to wait several minutes. This issue will be fixed after deleting the setting which named with special characters.

v-mengwe:
This issue also reproduces when uploading local setting with special characters name, e.g. 
```
{
"Values":{
   "test!@#":"123"
  }
}
```
![image](https://user-images.githubusercontent.com/62053665/100319618-e5489200-2ffa-11eb-8783-7f4624b5c3f4.png)

Issue #109, opened 2020-05-29
Fail to add a new setting which key name contains special characters


v-ruizh:
**OS:** Windows
**Build Version:** 20200527.3

**Repro Steps:**
Right click "Application Settings" node and select "Add New Setting..."-> Input key name with special characters like "test@"-> Input value "test"-> Enter.

**Expect:**
The setting "test@" is added successfully.

**Actual:**
An error occurs.
![image](https://user-images.githubusercontent.com/45614674/83237723-f3dafb80-a1c7-11ea-829b-e5673034496a.png)

PR #361, opened 2021-04-26
Don't allow special characters in app setting keys

wwlorey:
Fixes https://github.com/microsoft/vscode-azurestaticwebapps/issues/109
Fixes https://github.com/microsoft/vscode-azurestaticwebapps/issues/235

It looks like keys started adhering to the Linux naming standards.


nturinski (approved):
Tried in the portal.  It appears to work and then throws this descriptive error.

image

I think it's safe to assume that we should be using Linux app settings validation.

wwlorey:
I saw the same error in the extension too. It looks like the API changed since CTI filed the issue
Screen Shot 2021-04-26 at 4 19 40 PM

Issue #319, opened 2021-03-22
Error when Static Web App resource when there's no repositoryUrl


anthonychu:
We've started allowing Static Web Apps to be created without repositoryUrls and other details to support source providers other than GitHub (like Azure DevOps).

![image](https://user-images.githubusercontent.com/3982077/111927060-85c03780-8a6c-11eb-8a0a-3fa68302ddcc.png)

nturinski:
Related to https://github.com/microsoft/vscode-azurestaticwebapps/issues/265

nturinski:
Provide a better error as the internal error doesn't give the customers much information as to why the app is invalid.

PR #362, opened 2021-04-27
Throw better errors if SWA isn't linked to GitHub

wwlorey:
Fixes https://github.com/microsoft/vscode-azurestaticwebapps/issues/319

| Old | New |
|-|-|
|Screen Shot 2021-04-27 at 1 53 35 PM|Screen Shot 2021-04-27 at 1 52 58 PM|

The first SWA in this example has been linked to Azure DevOps so the SWA node loads fine but the environment doesn't.
The second SWA hasn't been linked to any version control.

### DIFF
--- a/src/tree/StaticWebAppTreeItem.ts
+++ b/src/tree/StaticWebAppTreeItem.ts
@@ -5,8 +5,8 @@
 
 import { WebSiteManagementClient, WebSiteManagementModels } from "@azure/arm-appservice";
 import { ProgressLocation, window } from "vscode";
-import { AzExtTreeItem, AzureParentTreeItem, IActionContext, TreeItemIconPath } from "vscode-azureextensionui";
-import { productionEnvironmentName } from '../constants';
+import { AzExtTreeItem, AzureParentTreeItem, IActionContext, IParsedError, parseError, TreeItemIconPath } from "vscode-azureextensionui";
+import { onlyGitHubSupported, productionEnvironmentName } from '../constants';
 import { ext } from "../extensionVariables";
 import { createWebSiteClient } from "../utils/azureClients";
 import { getResourceGroupFromId, pollAzureAsyncOperation } from "../utils/azureUtils";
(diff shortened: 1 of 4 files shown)


src/tree/SubscriptionTreeItem.ts
```diff
@@ -40,7 +40,17 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
         return await this.createTreeItemsWithErrorHandling(
             staticWebApps,
             'invalidStaticWebApp',
-            ss => new StaticWebAppTreeItem(this, ss),
+            ss => {
+                try {
+                    return new StaticWebAppTreeItem(this, ss);
+                } catch (error) {
+                    const parsedError: IParsedError = parseError(error);
+                    if (/(null|undefined).*repositoryUrl/.test(parsedError.message)) {
```
ejizba, line 48: We're the ones throwing this error, so you don't need to catch and parse it. Just change the `nonNullProp` [here](https://github.com/microsoft/vscode-azurestaticwebapps/blob/main/src/tree/StaticWebAppTreeItem.ts#L42) to something like
```
if (this.data.repositoryUrl) {
    this.repositoryUrl = this.data.repositoryUrl;
} else {
    throw new Error(onlyGitHubSupported);
}
```

src/tree/StaticWebAppTreeItem.ts
```diff
@@ -62,7 +62,15 @@ export class StaticWebAppTreeItem extends AzureParentTreeItem implements IAzureR
             envs,
             'invalidStaticEnvironment',
             async (env: WebSiteManagementModels.StaticSiteBuildARMResource) => {
-                return await EnvironmentTreeItem.createEnvironmentTreeItem(context, this, env);
+                try {
+                    return await EnvironmentTreeItem.createEnvironmentTreeItem(context, this, env);
+                } catch (error) {
+                    const parsedError: IParsedError = parseError(error);
+                    if (/(null|undefined).*sourceBranch/.test(parsedError.message)) {
```
nturinski, line 69: `sourceBranch` (which gets assigned to `EnvironmentTreeItem.branch`) isn't actually used that often.  With the first scenario, what functionality do we offer to users without having the branch defined?  

Most of our features are tied to GitHub, but we might still be able to offer them basic commands such as open in portal, browse, view properties.  It might be better to support this minimally.

ejizba: I don't disagree, but can't hurt to have a slightly better error message first. Internal errors are never supposed to be shown to the user

ejizba (approved):

PR #366, opened 2021-04-28
Add 'Create SWA Config File' command

wwlorey:
Fixes https://github.com/microsoft/vscode-azurestaticwebapps/issues/97

### DIFF
--- a/src/commands/createSwaConfigFile.ts
+++ b/src/commands/createSwaConfigFile.ts
@@ -5,32 +5,50 @@
 
 import { writeFile } from 'fs-extra';
 import { join } from 'path';
-import { EntryInfo, promise } from 'readdirp';
-import { TextDocument, Uri, window, workspace } from 'vscode';
+import { MessageItem, QuickPickItem, Uri, window, workspace } from 'vscode';
 import { ext } from 'vscode-azureappservice/out/src/extensionVariables';
-import { IActionContext } from 'vscode-azureextensionui';
 import { configFileName } from '../constants';
 import { localize } from '../utils/localize';
-import { getWorkspaceFolder } from '../utils/workspaceUtils';
+import { selectWorkspaceFolder } from '../utils/workspaceUtils';
 
-export async function createSwaConfigFile(context: IActionContext): Promise<void> {
-    const workspacePath: string = (await getWorkspaceFolder(context)).uri.fsPath;
-    let configFilePath: string | undefined = await getExistingConfigFilePath(workspacePath);
+export async function createSwaConfigFile(): Promise<void> {
+    const configFiles: Uri[] = await workspace.findFiles(`**/${configFileName}`);
 
-    if (configFilePath) {
-        void ext.ui.showWarningMessage(localize('swaConfigFileExists', 'Static Web App configuration file "{0}" already exists.', configFilePath));
-    } else {
-        configFilePath = join(workspacePath, configFileName);
+    const openFile: MessageItem = { title: localize('openFile', 'Open File') };
+    const createNew: MessageItem = { title: localize('createNew', 'Create New') };
+    const chooseFileToOverwrite: MessageItem = { title: localize('chooseFileToOverwrite', 'Choose File to Overwrite') };
+    const overwrite: MessageItem = { title: localize('overwrite', 'Overwrite') };
+
+    let response: MessageItem | undefined;
+    let configFilePath: string | undefined;
+
+    if (configFiles.length > 1) {
+        const multipleConfigFilesExist: string = localize('multipleConfigFilesExist', 'Multiple Static Web App configuration files already exist.');
+        response = await ext.ui.showWarningMessage(multipleConfigFilesExist, { modal: true }, createNew, chooseFileToOverwrite);
+    } else if (configFiles.length === 1) {
+        configFilePath = configFiles[0].fsPath;
+        const configFileExists: string = localize('configFileExists', 'Static Web App configuration file "{0}" already exists.', configFilePath);
+        response = await ext.ui.showWarningMessage(configFileExists, { modal: true }, openFile, createNew, overwrite);
+    }
+
+    if (response === chooseFileToOverwrite) {
+        const configFilePaths: QuickPickItem[] = configFiles.map(uri => { return { label: uri.fsPath } });
+        configFilePath = (await ext.ui.showQuickPick(configFilePaths, { canPickMany: false })).label;
+    } else if (!configFilePath || response === createNew) {
+        configFilePath = await promptForConfigFilePath();
+    }
+
+    if (response !== openFile) {
+        // The only case where we don't overwrite is `openFile`
         await writeFile(configFilePath, exampleConfigFile);
     }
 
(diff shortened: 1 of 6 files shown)


src/commands/createSwaConfigFile.ts
```diff
@@ -0,0 +1,47 @@
...
+        configFilePath = join(workspacePath, configFileName);
+        await writeFile(configFilePath, exampleConfigFile);
+    }
+
+    const textDocument: TextDocument = await workspace.openTextDocument(Uri.file(configFilePath));
+    await window.showTextDocument(textDocument);
+}
+
+async function getExistingConfigFilePath(workspacePath: string): Promise<string | undefined> {
+    const entries: EntryInfo[] = await promise(workspacePath, { fileFilter: configFileName });
```
nturinski, line 32: Why not just use fs-extra ala https://github.com/microsoft/vscode-azurefunctions/blob/85bed3eac6a8a236a5d2d886fafb6269a61c44ec/src/funcConfig/local.settings.ts#L73

wwlorey: Here's what the [docs](https://docs.microsoft.com/en-us/azure/static-web-apps/configuration#file-location) say about the config file's location:

> The recommended location for the staticwebapp.config.json is in the folder set as the app_location in the workflow file. However, the file may be placed in any location within your application source code folder.

I think that means it could live in any arbitrary nested folder so I made the search recursive with `readdirp`

nturinski: Hmm, in general, we probably don't want to do a recursive search through their entire project.  If they have a relatively large project without a swa config file, then it could take a while.  I think it makes sense to look maybe 2 folders deep.

I know Eric decided on one folder deep for Function projects: https://github.com/microsoft/vscode-azurefunctions/blob/85bed3eac6a8a236a5d2d886fafb6269a61c44ec/src/commands/createNewProject/verifyIsProject.ts#L22-L56

We also might be able to use their .yml config file (if they have it) to put the file in the app_location folder.

src/commands/createSwaConfigFile.ts
```diff
@@ -0,0 +1,47 @@
...
+import { configFileName } from '../constants';
+import { localize } from '../utils/localize';
+import { getWorkspaceFolder } from '../utils/workspaceUtils';
+
+export async function createSwaConfigFile(context: IActionContext): Promise<void> {
+    const workspacePath: string = (await getWorkspaceFolder(context)).uri.fsPath;
+    let configFilePath: string | undefined = await getExistingConfigFilePath(workspacePath);
+
+    if (configFilePath) {
+        void ext.ui.showWarningMessage(localize('swaConfigFileExists', 'Static Web App configuration file "{0}" already exists.', configFilePath));
```
nturinski, line 21: Should probably give them the option to overwrite it if they want to start fresh.

src/commands/createSwaConfigFile.ts
```diff
@@ -0,0 +1,47 @@
...
+import { EntryInfo, promise } from 'readdirp';
+import { TextDocument, Uri, window, workspace } from 'vscode';
+import { ext } from 'vscode-azureappservice/out/src/extensionVariables';
+import { IActionContext } from 'vscode-azureextensionui';
+import { configFileName } from '../constants';
+import { localize } from '../utils/localize';
+import { getWorkspaceFolder } from '../utils/workspaceUtils';
+
+export async function createSwaConfigFile(context: IActionContext): Promise<void> {
+    const workspacePath: string = (await getWorkspaceFolder(context)).uri.fsPath;
```
nturinski, line 17: We also might be able to use their .yml config file (if they have it) to put the file in the app_location folder.

src/commands/createSwaConfigFile.ts
ejizba: Well one reason I'm not doing a full recursive search is because that code gets called a lot, including when the extension is activated. The scenario here is a command the user explicitly started, so you may or may not want to do something different. Also the code you linked is pretty old in Functions. At this point, I'd probably recommend the `vscode.workspace.findFiles` API over pure `fs-extra`

src/commands/createSwaConfigFile.ts
```diff
@@ -0,0 +1,47 @@
...
+
+    if (configFilePath) {
+        void ext.ui.showWarningMessage(localize('swaConfigFileExists', 'Static Web App configuration file "{0}" already exists.', configFilePath));
+    } else {
+        configFilePath = join(workspacePath, configFileName);
+        await writeFile(configFilePath, exampleConfigFile);
+    }
+
+    const textDocument: TextDocument = await workspace.openTextDocument(Uri.file(configFilePath));
+    await window.showTextDocument(textDocument);
```
ejizba, line 28: You don't need to call `openTextDocument`. Just pass the uri to `showTextDocument`: https://github.com/microsoft/vscode/blob/main/src/vs/vscode.d.ts#L8434

src/commands/createSwaConfigFile.ts
ejizba: Feels like you should let the user pick where the file goes, maybe a quick pick and/or browse dialog

src/commands/createSwaConfigFile.ts
wwlorey: I updated the search to use `vscode.workspace.findFiles` and removed the `readdirp` dependency

src/commands/createSwaConfigFile.ts
```diff
@@ -0,0 +1,65 @@
...
+import { writeFile } from 'fs-extra';
+import { join } from 'path';
+import { MessageItem, QuickPickItem, Uri, window, workspace } from 'vscode';
+import { ext } from 'vscode-azureappservice/out/src/extensionVariables';
+import { configFileName } from '../constants';
+import { localize } from '../utils/localize';
+import { selectWorkspaceFolder } from '../utils/workspaceUtils';
+
+export async function createSwaConfigFile(): Promise<void> {
+    const configFiles: Uri[] = await workspace.findFiles(`**/${configFileName}`);
```
ejizba, line 15: Sorry uh I feel like we did not explain ourselves well enough in our comments last time around haha. Here is the user experience I expected:

- First, a quick pick with `Select location to create "{0}"` with one option per workspace folder and a "Browse..." pick at the bottom. Similar to "Deploy" or "Create New" dialogs in other extensions
- Once they've selected a location, check if that specific path already exists and prompt them if they want to overwrite it or not
- That's it

As it is now, you're doing a lot of the work up front (finding all config files, separate prompts depending on what you found, etc.). It makes the code more complex and possibly adds a delay between starting the command and showing the first prompt. It's also not a pattern I've seen before - like... when a user tries to create a new storage account, we don't prompt with "You have multiple storage accounts already - do you wanna overwrite one or create a new one?". Similar experience with the basic "New File" or "New Folder" - the validation comes _after_ they enter a location

wwlorey: Ah that makes sense 😄 Just pushed a much simpler version

ejizba (approved):

PR #369, opened 2021-04-29
Add git error handling to git calls

nturinski:


nturinski (commented):
Purposely skipped these lines https://github.com/microsoft/vscode-azurestaticwebapps/pull/369/files#diff-304c8205505364459ab4d0e8c1c6c186836847344ac594564e78734ee1c60f09R45-R58 since they'll go away in this PR: https://github.com/microsoft/vscode-azurestaticwebapps/pull/368

src/utils/gitUtils.ts
```diff
@@ -21,10 +21,25 @@ import { getSingleRootFsPath } from './workspaceUtils';
...
+    try {
+        return await callback();
+    } catch (gitError) {
+        // ignore empty commit errors which will happen if a user initializes a blank folder or have changes in a nested git repo
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+        if (!/nothing to commit/.test(gitError.stdout)) {
+            throw new GitError(gitError);
+        }
+
+        return null;
```
wwlorey, line 35: I don't know how likely this is, but isn't it possible for non-git errors to be thrown by these API calls? I worry that normal errors might be lost here

nturinski: Well, it should always be throwing the error regardless of whether or not it was a non-git error.  The only time it skips the error is if there was "nothing to commit" in the stdout.

Though we should only wrapping gitApi commands with it so we don't have to do [this](https://github.com/microsoft/vscode-azurestaticwebapps/pull/314/files#diff-5bdb19eb1274f186c25c09eaf175c74b188056e2d02e678d7b9ebfd547c4e256R173-R186) all the time.

wwlorey: It looks like this part will rewrite the error message if it isn't git related: https://github.com/microsoft/vscode-azurestaticwebapps/blob/c30303e02f95a3481ea53389c8a0d8d4c38c6133/src/errors.ts#L52-L54

nturinski: Well, it'll still display the error, it'll just say Git: `error string here`.  The `hint` should take `String(err)` and that regexp is just removing [husky](https://www.npmjs.com/package/husky) output.

I'm not totally sure what your concern is here though.  We should only use `callWithGitErrorHandling` specifically when calling gitApi commands.  I wouldn't add any other function calls inside of it so that the handling will only apply to Git related errors.

Are you worried that the API calls have other errors that aren't git related?  I'm pretty sure all the git API calls are `git` cli commands, so I think that we should be good.

src/utils/gitUtils.ts
```diff
@@ -21,10 +21,25 @@ import { getSingleRootFsPath } from './workspaceUtils';
...
 export type VerifiedGitWorkspaceState = GitWorkspaceState & { repo: Repository };
 
+// use only when using the gitApi
+export async function callWithGitErrorHandling<T>(callback: () => Promise<T>): Promise<T | null> {
+    try {
+        return await callback();
+    } catch (gitError) {
+        // ignore empty commit errors which will happen if a user initializes a blank folder or have changes in a nested git repo
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+        if (!/nothing to commit/.test(gitError.stdout)) {
```
ejizba, line 31: I totally thought this was backwards at first, and I feel like it can't just be me. I think it's because I expect your comment to match the first case in the `if` - something like this:
```
// ignore empty commit errors ...
if (/nothing to commit/.test(gitError.stdout)) {
    return null;
} else {
    throw new GitError(gitError);
}
```

src/utils/gitUtils.ts
```diff
@@ -21,10 +21,25 @@ import { getSingleRootFsPath } from './workspaceUtils';
 export type GitWorkspaceState = { repo: Repository | null, dirty: boolean, remoteRepo: ReposGetResponseData | undefined; hasAdminAccess: boolean };
 export type VerifiedGitWorkspaceState = GitWorkspaceState & { repo: Repository };
 
+// use only when using the gitApi
+export async function callWithGitErrorHandling<T>(callback: () => Promise<T>): Promise<T | null> {
+    try {
+        return await callback();
+    } catch (gitError) {
+        // ignore empty commit errors which will happen if a user initializes a blank folder or have changes in a nested git repo
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
```
ejizba, line 30: Can we "fix" this error instead of disabling it? Ideally use `parseError`, but at least make sure it's an object and more gracefully handle the types. Would be embarrassing if our error handling logic had it's own error

src/utils/gitUtils.ts
```diff
@@ -21,10 +21,25 @@ import { getSingleRootFsPath } from './workspaceUtils';
 export type GitWorkspaceState = { repo: Repository | null, dirty: boolean, remoteRepo: ReposGetResponseData | undefined; hasAdminAccess: boolean };
 export type VerifiedGitWorkspaceState = GitWorkspaceState & { repo: Repository };
 
+// use only when using the gitApi
+export async function callWithGitErrorHandling<T>(callback: () => Promise<T>): Promise<T | null> {
```
ejizba, line 25: Why we usin `null` instead of `undefined`?

src/utils/gitUtils.ts
```diff
@@ -119,15 +134,7 @@ async function promptForCommit(repo: Repository, value?: string): Promise<void>
...
-    try {
-        await repo.commit(commitMsg, commitOptions);
-    } catch (err) {
-        // ignore empty commit errors which will happen if a user initializes a blank folder or have changes in a nested git repo
-        // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
-        if (!/nothing to commit/.test(err.stdout)) {
-            throw new GitError(err);
-        }
-    }
+    await callWithGitErrorHandling(async () => await repo.commit(commitMsg, commitOptions));
```
ejizba, line 137: Maybe I missed something, but how come this single try/catch wasn't sufficient?

If you _do_ need all these extra calls to `callWithGitErrorHandling` - is there any way you can auto-magically add it to the git api for _all_ calls instead of manually adding it to each call like you're currently doing?

src/utils/gitUtils.ts
ejizba: I can't quite put my finger on it, but I think I agree with Will. Just seems weird to transform every possible error into a git error. If you just did a little validation (maybe `'gitErrorCode' in error`) before assuming it's a "gitError", I think I'd feel better. Would help to "fix" the eslint, too

src/utils/gitUtils.ts
nturinski: It's because I was seeing issues that had `Failed to execute git`.  I noticed that CTI reported that error even when I had this try/catch in the build.  That made me feel a lack of confidence that all the git commands would just work, so I wanted to have better error handling for all of them.

I was thinking of a way I could automatically have all the commands have the error handling without doing what I'm doing. I'll try to think on it a bit more, might be able to do it if I get creative.

src/utils/gitUtils.ts
nturinski: All the gitApi command (specifically git.init and git.GetRepository) return `Repository | null` so just had to do it for typing reasons.

src/utils/gitUtils.ts
nturinski: I'm not entirely sure what you mean by fixing it.  Do you mean fix the cause of this error in our extension? 

I've tried to mitigate it by preventing users from creating SWAs if their workspace is completely empty.  The nested repository case seems unlikely, but I ran into it and was having issues with VS Code reporting that I had git changes, but when I actually tried to commit, it was throwing an error saying that there were no changes.  I actually haven't been able to reproduce it anymore so maybe they're working on it.  There is an issue following nested repos: https://github.com/microsoft/vscode/issues/37947

`parseError` doesn't work in this case because it just sets the message equal to `Failed to execute git` since that's what all the gitApi commands throw up.  All the git terminal information is in the `stderr` and `stdout`.

src/utils/gitUtils.ts
nturinski: I couldn't find a nice way to automagically wrap all the gitApi and repo commands so I opted to just use an error handler that makes this look a little cleaner.

src/errors.ts
```diff
@@ -10,9 +10,24 @@ export class NoWorkspaceError extends Error {
     public message: string = localize('noWorkspaceError', 'You must have a workspace open to perform this operation.');
 }
 
+// use only when using the gitApi
```
ejizba, line 13: how about `handleGitErrors`

Just feels like functions should start with a verb instead of being a noun

src/errors.ts
```diff
@@ -3,16 +3,39 @@
...
+        // ignore empty commit errors which will happen if a user initializes a blank folder or have changes in a nested git repo
+        if (stdout && /nothing to commit/.test(stdout)) {
+            return;
+        } else if ('gitErrorCode' in err) {
+            throw new GitError(<GitErrorType>err);
+        }
+
+        // if it's not a GitError, just throw up the original error
+        throw err;
+    } else if (err === null || err === undefined) {
```
ejizba, line 27: I'd rather you got rid of these last two cases and just threw up the original error for those, too. Our standard error handling should take care of it

ejizba (approved):

Issue #363, opened 2021-04-28
There is only one setting after creating respectively a different setting in different environments


v-xinda:
**OS:** Windows
**Build Version:** 20210427.2

**Repro Steps:**
1. Create a new static web app.
2. Create an environment for the static web app.
3. Expand the Production -> Application Settings -> Create a new setting `new1=123456`.
4. Expand the Environment -> Application Settings -> Create a new setting `test1=asdfgh`.
5. Refresh the Application Settings of the Production and Environment.
6. Check whether the two settings all display under the "Application Settings" nodes of the Production and Environment.

**Expect:**
The two settings all display under the "Application Settings" node of the Production and Environment.

**Actual:**
Only show one setting under the "Application Settings" nodes of the Production and Environment.
![setting](https://user-images.githubusercontent.com/57662131/116369097-91f19e80-a83b-11eb-8fbe-c55c7974589f.gif)

PR #372, opened 2021-04-29
Use build for non-production builds

nturinski:
Fixes https://github.com/microsoft/vscode-azurestaticwebapps/issues/363

I had to move the treeItem creation further below because it relied on envTreeItem properties that weren't defined yet.

Technically `production` is also a build and it worked fine when I used `listStaticSiteBuildFunctions` (rather than `listStaticSiteFunctions`, but I figured I should change it to use the proper API call in case anything changes in the future.

### DIFF
--- a/src/commands/appSettings/AppSettingsClient.ts
+++ b/src/commands/appSettings/AppSettingsClient.ts
@@ -16,12 +16,17 @@ export class AppSettingsClient implements IAppSettingsClient {
     public fullName: string;
     public resourceGroup: string;
     public root: ISubscriptionContext;
+    public prId: string;
+    public isBuild: boolean;
 
     constructor(node: EnvironmentTreeItem) {
         this.ssId = node.id;
         this.fullName = node.parent.name;
         this.resourceGroup = node.parent.resourceGroup;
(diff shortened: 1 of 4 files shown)


src/commands/appSettings/AppSettingsClient.ts
```diff
@@ -16,25 +16,32 @@ export class AppSettingsClient implements IAppSettingsClient {
...
+    public prId: string;
+    public isBuild: boolean;
 
     constructor(node: EnvironmentTreeItem) {
         this.ssId = node.id;
         this.fullName = node.parent.name;
         this.resourceGroup = node.parent.resourceGroup;
         this.root = node.root;
 
+        this.prId = node.buildId;
```
wwlorey, line 28: Why not call this `buildId` instead?

wwlorey (approved):

src/commands/appSettings/AppSettingsClient.ts
nturinski: I chose `prId` because that's the name of the property in the API, but I'd be fine calling it `buildId` as well.
image

PR #375, opened 2021-05-05
Add 'create fork' option if user doesn't have admin rights

wwlorey:
Fixes https://github.com/microsoft/vscode-azurestaticwebapps/issues/329

### DIFF
--- a/src/utils/gitHubUtils.ts
+++ b/src/utils/gitHubUtils.ts
@@ -4,14 +4,13 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { Octokit, RestEndpointMethodTypes } from '@octokit/rest';
-import { authentication, MessageItem, ProgressLocation, window } from 'vscode';
+import { authentication, ProgressLocation, window } from 'vscode';
 import { IActionContext, IAzureQuickPickItem, parseError, UserCancelledError } from 'vscode-azureextensionui';
 import { createOctokitClient } from '../commands/github/createOctokitClient';
 import { ListOrgsForUserData, OrgForAuthenticatedUserData, ReposGetResponseData } from '../gitHubTypings';
 import { getRepoFullname, tryGetRemote } from './gitUtils';
 import { localize } from './localize';
-import { nonNullProp, nonNullValue } from './nonNull';
-import { openUrl } from './openUrl';
+import { nonNullProp } from './nonNull';
 
 /**
  * @param label Property of JSON that will be used as the QuickPicks label
@@ -84,7 +83,7 @@ export function isUser(orgData: ListOrgsForUserData | OrgForAuthenticatedUserDat
     return !!orgData && 'type' in orgData && orgData.type === 'User';
 }
 
-export async function createFork(context: IActionContext, remoteRepo: ReposGetResponseData): Promise<void> {
+export async function createFork(context: IActionContext, remoteRepo: ReposGetResponseData): Promise<RestEndpointMethodTypes["repos"]["createFork"]["response"]> {
     let createForkResponse: RestEndpointMethodTypes["repos"]["createFork"]["response"] | undefined;
 
     if (remoteRepo.owner?.login) {
@@ -100,16 +99,7 @@ export async function createFork(context: IActionContext, remoteRepo: ReposGetRe
     }
 
     if (createForkResponse?.status === 202) {
-        const openInGitHub: MessageItem = { title: localize('openInGitHub', 'Open in GitHub') };
-        const success: string = localize('forkSuccess', 'Successfully forked repository "{0}"', remoteRepo.name);
-
-        void window.showInformationMessage(success, openInGitHub).then(async (result) => {
-            if (result === openInGitHub) {
-                await openUrl(nonNullValue(createForkResponse).data.html_url);
-            }
-        });
-
-        throw new UserCancelledError();
+        return createForkResponse;
     } else {
         throw new Error(localize('forkFail', 'Could not automatically fork repository. Please fork [{0}]({1}) manually.', remoteRepo.name, remoteRepo.html_url));
     }
--- a/src/utils/gitHubUtils.ts
+++ b/src/utils/gitHubUtils.ts
@@ -3,11 +3,12 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { Octokit, RestEndpointMethodTypes } from '@octokit/rest';
+import { Octokit } from '@octokit/rest';
 import { authentication, ProgressLocation, window } from 'vscode';
 import { IActionContext, IAzureQuickPickItem, parseError, UserCancelledError } from 'vscode-azureextensionui';
 import { createOctokitClient } from '../commands/github/createOctokitClient';
-import { ListOrgsForUserData, OrgForAuthenticatedUserData, ReposGetResponseData } from '../gitHubTypings';
+import { ext } from '../extensionVariables';
+import { ListOrgsForUserData, OrgForAuthenticatedUserData, ReposCreateForkResponse, ReposGetResponseData } from '../gitHubTypings';
 import { getRepoFullname, tryGetRemote } from './gitUtils';
 import { localize } from './localize';
 import { nonNullProp } from './nonNull';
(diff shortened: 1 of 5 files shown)


nturinski (commented):
When we fork the repo for the user, I believe that we need to clone/open the repo for the user as well.  I think you can accomplish this with the `gitClone` command that we have.  We'll need to change it to accept a node or a url. 

It'll prompt the user for a workspace location and then proceed to clone it down.  And then when it's done, prompt the user like so:

image

It might be a little jarring, but I think a user would expect to have to clone after forking.

@fiveisprime Do you have any opinions on this?

wwlorey:
> It might be a little jarring, but I think a user would expect to have to clone after forking.

IMO after seeing it in action, it doesn't seem too jarring. Definitely better than just showing a link to the new repo

src/utils/gitUtils.ts
```diff
@@ -64,9 +64,18 @@ export async function verifyGitWorkspaceForCreation(context: IActionContext, git
...
-        await ext.ui.showWarningMessage(adminAccess, { modal: true });
+        const adminAccess: string = localize('adminAccess', 'Admin access to the GitHub repository "{0}" is required. Would you like to create a fork?', gitWorkspaceState.remoteRepo.name);
+        const createForkItem: MessageItem = { title: localize('createFork', 'Create Fork') };
+        await ext.ui.showWarningMessage(adminAccess, { modal: true }, createForkItem);
 
+        const repoUrl: string = (await createFork(context, gitWorkspaceState.remoteRepo)).data.html_url;
+
+        const forkSuccess: string = localize('forkSuccess', 'Successfully forked "{0}". Would you like to clone your new repository?', gitWorkspaceState.remoteRepo.name);
+        const clone: MessageItem = { title: localize('clone', 'Clone') };
+        await ext.ui.showWarningMessage(forkSuccess, { modal: true }, clone);
```
nturinski, line 75: Could you change the value of `cancelStep` to `cloneFork` here so we can distinguish between people who canceled at the fork step and the clone step?

src/utils/gitUtils.ts
```diff
@@ -64,9 +64,18 @@ export async function verifyGitWorkspaceForCreation(context: IActionContext, git
...
+
+        await commands.executeCommand('git.clone', repoUrl);
+        throw new UserCancelledError();
```
nturinski, line 78: And similar thing here-- Maybe something like `afterCloneFork`.

src/utils/gitHubUtils.ts
```diff
@@ -80,3 +82,25 @@ export function isUser(orgData: ListOrgsForUserData | OrgForAuthenticatedUserDat
...
 }
+
+export async function createFork(context: IActionContext, remoteRepo: ReposGetResponseData): Promise<RestEndpointMethodTypes["repos"]["createFork"]["response"]> {
+    let createForkResponse: RestEndpointMethodTypes["repos"]["createFork"]["response"] | undefined;
+
+    if (remoteRepo.owner?.login) {
+        const client: Octokit = await createOctokitClient(context);
+        const title: string = localize('forking', 'Forking "{0}"...', remoteRepo.name);
+
+        await window.withProgress({ location: ProgressLocation.Notification, title }, async () => {
```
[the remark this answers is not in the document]
wwlorey: I've had it take ~5 seconds sometimes and sometimes it's instantaneous. I'll add an output log too

> I think we should use the output channel to give more description as to why we are git pulling after creation

That could be better than showing another notification. We're flashing a lot of stuff at people when that SWA finishes creating 😂

src/utils/gitHubUtils.ts
```diff
@@ -80,3 +82,25 @@ export function isUser(orgData: ListOrgsForUserData | OrgForAuthenticatedUserDat
...
+                owner: nonNullProp(remoteRepo, 'owner').login,
+                repo: remoteRepo.name
+            });
+        });
+    }
+
+    if (createForkResponse?.status === 202) {
+        return createForkResponse;
+    } else {
+        throw new Error(localize('forkFail', 'Could not automatically fork repository. Please fork [{0}]({1}) manually.', remoteRepo.name, remoteRepo.html_url));
```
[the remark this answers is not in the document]
nturinski: When I say I'd assume it works, I'm referring to the hyperlinking within a localized string

src/utils/gitUtils.ts
```diff
@@ -63,10 +63,20 @@ export async function verifyGitWorkspaceForCreation(context: IActionContext, git
...
-        await ext.ui.showWarningMessage(adminAccess, { modal: true });
+        const repoUrl: string = (await createFork(context, gitWorkspaceState.remoteRepo)).data.html_url;
 
+        context.telemetry.properties.cancelStep = 'cloneFork';
+        const forkSuccess: string = localize('forkSuccess', 'Successfully forked "{0}". Would you like to clone your new repository?', gitWorkspaceState.remoteRepo.name);
+        const clone: MessageItem = { title: localize('clone', 'Clone') };
+        await ext.ui.showWarningMessage(forkSuccess, { modal: true }, clone);
+
+        context.telemetry.properties.cancelStep = 'afterCloneFork';
+        await commands.executeCommand('git.clone', repoUrl);
```
nturinski, line 78: Oh yeah, could use our `cloneRepo` command?  It effectively does just this, but if we add some additional behavior, would be nice to consolidate this logic.

I know I have a spot where I use `await commands.executeCommand('git.clone', repoUrl);` that I should probably refactor as well...

wwlorey: This was the only other `git.clone` I found:

https://github.com/microsoft/vscode-azurestaticwebapps/blob/ee207dde2470020ae4de4973070a71ee46743f6e/src/utils/workspaceUtils.ts#L58

Does the git extension prompt for the repo to clone in that case?

src/utils/gitHubUtils.ts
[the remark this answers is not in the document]
wwlorey: Yeah it works:
Screen Shot 2021-05-06 at 5 10 29 PM

src/utils/gitUtils.ts
```diff
@@ -63,10 +64,24 @@ export async function verifyGitWorkspaceForCreation(context: IActionContext, git
...
+        const createForkItem: MessageItem = { title: localize('createFork', 'Create Fork') };
+        await ext.ui.showWarningMessage(adminAccess, { modal: true }, createForkItem);
+
+        const repoUrl: string = (await createFork(context, gitWorkspaceState.remoteRepo)).data.html_url;
+
+        context.telemetry.properties.cancelStep = 'cloneFork';
+        const forkSuccess: string = localize('forkSuccess', 'Successfully forked "{0}". Would you like to clone your new repository?', gitWorkspaceState.remoteRepo.name);
+        const clone: MessageItem = { title: localize('clone', 'Clone Repo') };
+        ext.outputChannel.appendLog(forkSuccess);
+        void window.showInformationMessage(forkSuccess, clone).then(result => {
```
nturinski, line 77: You should be able to just use await and check the input result.  The problem with using `.then` is we'll miss the telemetry of the `afterCloneFork` because once the error is thrown, the event is sent up to telemetry.  

Normally we don't like waiting on information prompts, but due to the nature of this one (there's an action afterwards that the user has to interact with) it's okay to use.  Normally we're just like, notifying that the action has completed.

src/utils/gitUtils.ts
```diff
@@ -75,10 +76,24 @@ export async function verifyGitWorkspaceForCreation(context: IActionContext, git
...
+        await ext.ui.showWarningMessage(adminAccess, { modal: true }, createForkItem);
 
-        const adminAccess: string = localize('adminAccess', 'Admin access to the GitHub repository is required.  Use a repo with admin access or create a fork.');
-        await ext.ui.showWarningMessage(adminAccess, { modal: true });
+        const repoUrl: string = (await createFork(context, gitWorkspaceState.remoteRepo)).data.html_url;
 
+        context.telemetry.properties.cancelStep = 'cloneFork';
+        const forkSuccess: string = localize('forkSuccess', 'Successfully forked "{0}". Would you like to clone your new repository?', gitWorkspaceState.remoteRepo.name);
+        const clone: MessageItem = { title: localize('clone', 'Clone Repo') };
+        ext.outputChannel.appendLog(forkSuccess);
```
nturinski, line 88: Kind of a nit, but.... 

nit: Could you remove the question "Would you like to clone your new repository?" from the appendLog string?  It doesn't make any sense in the context of the output channel 😬 

I'd understand if you didn't.

If it doesn't bother you, then don't worry about it.

...

👀

nturinski (approved):
👏

src/utils/gitUtils.ts
nturinski: Yes, without a resourceUrl, it prompts up like so:

image

src/utils/gitUtils.ts
wwlorey: 😄 No you're right, that looks goofy. Consider it gone 😤

Issue #371, opened 2021-04-29
Environment TreeItem children aren't sorted


nturinski:
![image](https://user-images.githubusercontent.com/5290572/116603505-d899d400-a8e1-11eb-9da7-d2018d397653.png)

PR #376, opened 2021-05-05
Sort EnvironmentTreeItem children

wwlorey:
Fixes https://github.com/microsoft/vscode-azurestaticwebapps/issues/371

Screen Shot 2021-05-05 at 4 38 08 PM

### DIFF
--- a/src/tree/EnvironmentTreeItem.ts
+++ b/src/tree/EnvironmentTreeItem.ts
@@ -105,10 +105,6 @@ export class EnvironmentTreeItem extends AzureParentTreeItem implements IAzureRe
 
     public async loadMoreChildrenImpl(_clearCache: boolean, context: IActionContext): Promise<AzExtTreeItem[]> {
         const children: AzExtTreeItem[] = [this.actionsTreeItem];
-        if (this.inWorkspace) {
-            children.push(...this.gitHubConfigGroupTreeItems);
-        }
-
         const client: WebSiteManagementClient = await createWebSiteClient(this.root);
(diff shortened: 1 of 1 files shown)


src/tree/EnvironmentTreeItem.ts
```diff
@@ -124,6 +120,10 @@ export class EnvironmentTreeItem extends AzureParentTreeItem implements IAzureRe
             children.push(this.appSettingsTreeItem, this.functionsTreeItem);
         }
 
+        if (this.inWorkspace) {
```
nturinski, line 123: image

😏 

I think we should just fix the `compareChildrenImpl` command.  If we do something like 

```
if (item1 instanceof GenericTreeItem && !(item2 instanceof GenericTreeItem)) {
            return -1;
        } else if (!(item1 instanceof GenericTreeItem) && item2 instanceof GenericTreeItem) {
            return 1;
        } else {
            return super.compareChildrenImpl(item1, item2);
        }
}
```

nturinski (commented):
While this works, we should probably just sort

src/tree/EnvironmentTreeItem.ts
ejizba: Our pattern for app service and functions is to sort this level of the tree manually based on what we think is most important for users, not alphabetically (which is what `super.compareChildrenImpl` would do). I think we should keep doing the same here.

I'm guessing you don't want "Actions" at the bottom, which is possible if you sort alphabetically and if/when we ever localize.

wwlorey:
Here's the new order:
Screen Shot 2021-05-06 at 11 38 28 AM
Screen Shot 2021-05-06 at 11 39 10 AM

nturinski (approved):

PR #378, opened 2021-05-06
Add sku step and support standard sku in adv create

nturinski:


src/commands/createStaticWebApp/SkuListStep.ts
```diff
@@ -0,0 +1,41 @@
...
+export class SkuListStep extends AzureWizardPromptStep<IStaticWebAppWizardContext> {
+    public async prompt(wizardContext: IStaticWebAppWizardContext): Promise<void> {
+        const skus: IAzureQuickPickItem<WebSiteManagementModels.SkuDescription>[] = SkuListStep.getSkus().map(s => {
+            return {
+                label: nonNullProp(s, 'name'),
+                data: s
+            };
+        });
+
+        const placeHolder = localize('selectSku', 'Select a sku');
```
wwlorey, line 21: nit: 
```suggestion
        const placeHolder: string = localize('selectSku', 'Select a sku');
```

src/commands/createStaticWebApp/SkuListStep.ts
```diff
@@ -0,0 +1,41 @@
...
+        wizardContext.sku = (await wizardContext.ui.showQuickPick(skus, { placeHolder, suppressPersistence: true })).data;
+    }
+
+    public shouldPrompt(wizardContext: IStaticWebAppWizardContext): boolean {
+        return !wizardContext.sku;
+    }
+
+    public static getSkus(): WebSiteManagementModels.SkuDescription[] {
+        // seems incomplete, but only name/tier are necessary for the SiteEnvelope to pick the right sku
```
wwlorey, line 30: Why do you say it seems incomplete?

wwlorey (approved):
I did some local testing too and it's looking good 👍

src/commands/createStaticWebApp/SkuListStep.ts
nturinski: Ah, I just meant that the SkuDescription has a lot more properties.  And in azuretools/appservice, the SKUs that we hardcode have a lot more properties filled out.

PR #382, opened 2021-05-07
Reveal environment tree item after creating SWA

wwlorey:


src/commands/createStaticWebApp/createStaticWebApp.ts
```diff
@@ -71,7 +72,15 @@ export async function createStaticWebApp(context: IActionContext & Partial<ICrea
     });
 
     await gitPull(nonNullProp(context, 'repo'));
-    await openYAMLConfigFile(context, swaNode);
+
+    const environmentNode = <EnvironmentTreeItem>(await swaNode.loadAllChildren(context)).find(node => {
+        return node instanceof EnvironmentTreeItem && node.inWorkspace;
+    });
+    await environmentNode.loadAllChildren(context);
+    const nodeToReveal: GitHubConfigGroupTreeItem | EnvironmentTreeItem  = environmentNode.gitHubConfigGroupTreeItems.length === 1 ?
```
nturinski, line 80: I think from the offline discussion, we want to reveal the `EnvironmentTreeItem`, not the config node.  The idea is just to show them that there is a configuration node.

src/commands/createStaticWebApp/createStaticWebApp.ts
```diff
@@ -71,7 +72,15 @@ export async function createStaticWebApp(context: IActionContext & Partial<ICrea
     });
 
     await gitPull(nonNullProp(context, 'repo'));
-    await openYAMLConfigFile(context, swaNode);
+
+    const environmentNode = <EnvironmentTreeItem>(await swaNode.loadAllChildren(context)).find(node => {
```
nturinski, line 76: Since this is after creating a SWA, I think it's safe for us to just return the first child node since there shouldn't be any other tree item.

You're not really verifying the type anyway since you're just casting it so I feel like we should be able to just do

```suggestion
    const environmentNode = <EnvironmentTreeItem>(await swaNode.loadAllChildren(context))[0]
```

nturinski: If you want to be extra careful, this is what I seemed to do many many months ago

`const productionEnv: EnvironmentTreeItem | undefined = <EnvironmentTreeItem | undefined>(await swaNode.loadAllChildren(context)).find((ti) => { return ti instanceof EnvironmentTreeItem && ti.label === productionEnvironmentName; });`

nturinski (approved):
👌

PR #383, opened 2021-05-07
Swap swa name with github repo name steps

nturinski:
I also moved the GitHub repo create steps _after_ resourceList and skuList steps.  It felt like it made more sense to have all the Azure stuff together rather than doing SWA name, GitHub stuff, and then RG and Sku.  

@fiveisprime Let me know if you think this is weird.

### DIFF
--- a/src/commands/createRepo/RepoNameStep.ts
+++ b/src/commands/createRepo/RepoNameStep.ts
@@ -4,7 +4,6 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { Octokit } from '@octokit/rest';
-import { workspace } from 'vscode';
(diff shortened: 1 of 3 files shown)


wwlorey:
This might just be a me thing - but the way it modifies my subscription name as the SWA name looks kinda goofy 😄  It seems like the github username gels better as the prefix for the SWA name
Screen Shot 2021-05-07 at 4 34 53 PM
Is there something else we could use instead? Or do you think this just a me problem?

nturinski:
Is your full sub display name WillLoreywiloreyTest? lol

But honestly, swa names do _not_ have to be globally unique, so I'll just opt to only have the folder name.

We can't use the GitHub username anymore since the steps are swapped.

PR #387, opened 2021-05-08
Add long running notification when verifying workspace

wwlorey:


src/commands/createStaticWebApp/createStaticWebApp.ts
```diff
@@ -26,34 +26,35 @@ export async function createStaticWebApp(context: IActionContext & Partial<ICrea
...
-            } else {
-                if (!context.advancedCreation) {
-                    // default repo to private for basic create
-                    context.newRepoIsPrivate = true;
-                    // set the org to the authenticated user for creation
-                    context.orgData = await GitHubOrgListStep.getAuthenticatedUser(context);
-                }
+    const progressOptions: ProgressOptions = {
+        location: ProgressLocation.Notification,
+        title: localize('verifyingWorkspace', 'Verifying workspace...')
```
wwlorey, line 31: I'm pretty sure someone suggested this verbiage instead of "Create Starting" but lemme know if I'm mistaken..

nturinski: Yah that was Mr. Bossman himself, @ejizba

nturinski (approved):
I honestly do feel like it's kind of hard to see, but I imagine if someone thought their VS Code froze, that they would look around to see if anything is happening somewhere.  I'll let this one bake-- we can always revert this very easily if we feel like it's off.

PR #388, opened 2021-05-08
Fix strings, remove preview

nturinski:
image

Basing the new prompt off of this line from [here](https://docs.microsoft.com/en-us/azure/static-web-apps/overview)

> Azure Static Web Apps is a service that automatically builds and deploys full stack web apps to Azure from a code repository.

Kind of going with the informal approach we tried with image

Very very open to feedback @fiveisprime @ejizba

### DIFF
--- a/src/commands/createRepo/RepoNameStep.ts
+++ b/src/commands/createRepo/RepoNameStep.ts
@@ -19,7 +19,7 @@ export class RepoNameStep extends AzureWizardPromptStep<IStaticWebAppWizardConte
             undefined;
 
         wizardContext.newRepoName = (await ext.ui.showInputBox({
-            prompt: localize('AppServicePlanPrompt', 'Enter the name of the new GitHub repository. Special characters will be replaced with "-" upon creation.'),
+            prompt: localize('AppServicePlanPrompt', 'Enter the name of the new GitHub repository. We automatically build and deploy to Azure from your repo using GitHub Actions!'),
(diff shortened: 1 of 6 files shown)


src/commands/createRepo/RepoNameStep.ts
```diff
@@ -19,7 +19,7 @@ export class RepoNameStep extends AzureWizardPromptStep<IStaticWebAppWizardConte
             undefined;
 
         wizardContext.newRepoName = (await ext.ui.showInputBox({
-            prompt: localize('AppServicePlanPrompt', 'Enter the name of the new GitHub repository. Special characters will be replaced with "-" upon creation.'),
+            prompt: localize('AppServicePlanPrompt', 'Enter the name of the new GitHub repository. We automatically build and deploy to Azure from your repo using GitHub Actions!'),
```
ejizba, line 22: How about:

> Azure Static Web Apps will automatically build and deploy using GitHub Actions.

- exclamation point seems a bit much IMO
- "We" is a lot more informal than the example screenshot you gave. Also who is "We"? Because the VS Code extension doesn't do that...  I like mentioning ASWA because we've talked about how this prompt is confusing in the sense of "How does this relate to static web apps?"
- So many "to \<this\>", "from \<this\>", "using \<this\>" feels a bit awkward. I think "to Azure" and "from your repo" can be inferred whereas "using GitHub Actions" is actually informative

PR #405, opened 2021-05-25
Create ms-azuretools.vscode-azurestaticwebapps

chaos-ui:


nturinski:
Thanks for the contribution.  I'm sorry, but could you give me more context on what a `ms-azuretools.vscode-azurestaticwebapps` file is for?  It appears to be an empty file.

PR #407, opened 2021-05-27
Unit tests for Git workspaces and default branches

nturinski:
The biggest thing that I had to change to get this working is change the testFolderDirectory to be in the homedir rather than tempdir.  The reason being, for some reason, the VS Code Git API could init repos, but it couldn't access/retrieve any of those repos for some reasons.  

My guess is the API doesn't have certain permissions to access the repos, though this did not apply to the Linux OS.  I'm fine with trying a different path though.

Also, I was having a problem where if I tried it in the testWorkspace, it would just hang up and freeze.  No idea what that was about-- I'll probably still mess with it a bit and see if I can resolve that because I'd prefer to test in the opened workspace (since that's what we enforce) but these are unit tests, not E2E so I think this should suffice.


test/global.test.ts
```diff
@@ -26,8 +39,39 @@ suiteSetup(async function (this: Mocha.Context): Promise<void> {
...
+    await fse.emptyDir(testFolderPath);
+    testWorkspacePath = await initTestWorkspacePath();
+
     ext.outputChannel = new TestOutputChannel();
     ext.ui = testUserInput;
 
     longRunningTestsEnabled = !/^(false|0)?$/i.test(process.env.ENABLE_LONG_RUNNING_TESTS || '');
+
+    // set the user name and email since this seems to be required for everything
+    await cpUtils.executeCommand(undefined, undefined, 'git', 'config', '--global', 'user.name', 'Automated Tester');
```
ejizba, line 52: This could definitely mess with someone's git config if they're running tests locally. Also I wonder if the CI machines have their own git config and this could cause problems there, too

test/global.test.ts
```diff
@@ -26,8 +39,39 @@ suiteSetup(async function (this: Mocha.Context): Promise<void> {
...
+    await cpUtils.executeCommand(undefined, undefined, 'git', 'config', '--global', 'user.email', 'automated@testing.com');
 });
+
+suiteTeardown(async function (this: Mocha.Context): Promise<void> {
+    await fse.emptyDir(testFolderPath);
+    await fse.remove(testFolderPath);
+});
+
+export async function cleanTestWorkspace(): Promise<void> {
```
ejizba, line 61: If you're not gonna use `testWorkspacePath`, I don't think you need to add that code. If you are going to use it, you need to call `cleanTestWorkspace` in `suiteTeardown`

test/global.test.ts
```diff
@@ -26,8 +39,39 @@ suiteSetup(async function (this: Mocha.Context): Promise<void> {
...
     longRunningTestsEnabled = !/^(false|0)?$/i.test(process.env.ENABLE_LONG_RUNNING_TESTS || '');
+
+    // set the user name and email since this seems to be required for everything
+    await cpUtils.executeCommand(undefined, undefined, 'git', 'config', '--global', 'user.name', 'Automated Tester');
+    await cpUtils.executeCommand(undefined, undefined, 'git', 'config', '--global', 'user.email', 'automated@testing.com');
 });
+
+suiteTeardown(async function (this: Mocha.Context): Promise<void> {
+    await fse.emptyDir(testFolderPath);
+    await fse.remove(testFolderPath);
```
ejizba, line 58: Nit: you don't need `emptyDir` if you do `remove`

test/getGitWorkspaceState.test.ts
```diff
@@ -0,0 +1,73 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as assert from 'assert';
+import * as fse from 'fs-extra';
+import { join } from 'path';
+import { Uri } from "vscode";
+import { parseError } from 'vscode-azureextensionui';
```
ejizba, line 10: Everything from the ui package should be imported through "extension.bundle.ts"

test/getGitWorkspaceState.test.ts
```diff
@@ -0,0 +1,73 @@
...
+
+suite('Workspace Configurations for SWA Creation', function (this: Mocha.Suite): void {
+    this.timeout(2 * 60 * 1000);
+    const context = createTestActionContext();
+    const testFolderUri: Uri = Uri.file(testFolderPath);
+
+    test('Empty workspace with no git repository', async () => {
+        const gitWorkspaceState: GitWorkspaceState = await getGitWorkspaceState(context, testFolderUri)
+        let message: string = '';
+        try {
```
ejizba, line 22: I would use a util method like this:
https://github.com/microsoft/vscode-azuretools/blob/main/ui/test/assertThrowsAsync.ts

test/getGitWorkspaceState.test.ts
```diff
@@ -0,0 +1,73 @@
...
+        await fse.writeFile(join(testFolderPath, 'test.txt'), 'Test');
+
+        const gitWorkspaceState: GitWorkspaceState = await getGitWorkspaceState(context, testFolderUri);
+        assert.strictEqual(gitWorkspaceState.repo, null, `Workspace contained a repository prior to test "${gitWorkspaceState.repo?.rootUri.fsPath}"`);
+        const testCommitMsg: string = 'Test commit';
+        await testUserInput.runWithInputs(['Create', testCommitMsg], async () => {
+            await verifyGitWorkspaceForCreation(context, gitWorkspaceState, testFolderUri);
+        });
+
+        assert.ok(gitWorkspaceState.repo);
```
ejizba, line 42: It's a bit weird to me that `verifyGitWorkspaceForCreation` modifies the original `gitWorkspaceState` object. I would've expected the test needed to call `getGitWorkspaceState` again

test/getGitWorkspaceState.test.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    });
+
+    test('Workspace on default branch', async () => {
+        const gitWorkspaceState: GitWorkspaceState = await getGitWorkspaceState(context, testFolderUri);
```
ejizba, line 47: I'm really not a fan of unit tests that rely on the state from a previous test. Then they're not really "unit" tests and if one fails in the sequence, they all fail. Can you make each test standalone?

test/getGitWorkspaceState.test.ts
```diff
@@ -0,0 +1,73 @@
...
+
+import * as assert from 'assert';
+import * as fse from 'fs-extra';
+import { join } from 'path';
+import { Uri } from "vscode";
+import { parseError } from 'vscode-azureextensionui';
+import { getGitWorkspaceState, GitWorkspaceState, promptForDefaultBranch, tryGetDefaultBranch, verifyGitWorkspaceForCreation } from "../extension.bundle";
+import { createTestActionContext, testFolderPath, testUserInput } from "./global.test";
+
+suite('Workspace Configurations for SWA Creation', function (this: Mocha.Suite): void {
```
ejizba, line 14: Seems like there's a few cases inside `verifyGitWorkspaceForCreation` that are not covered here
- I don't have admin access to a repo
- I have a dirty repo
- My repo is fine and no prompts/errors should occur

test/getGitWorkspaceState.test.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    test('Workspace on default branch', async () => {
+        const gitWorkspaceState: GitWorkspaceState = await getGitWorkspaceState(context, testFolderUri);
+        if (!gitWorkspaceState.repo) {
+            throw new Error('Could not retrieve git repository.');
+        }
+
+        // shouldn't prompt
+        await promptForDefaultBranch(context, gitWorkspaceState.repo);
+        assert.strictEqual(await tryGetDefaultBranch(gitWorkspaceState.repo), 'master');
```
ejizba, line 54: Can we have one unit test for 'master' and one for 'main'

test/tryGetDefaultBranch.test.ts
```diff
@@ -0,0 +1,47 @@
...
+            throw new Error('Could not retrieve git repository.');
+        }
+
+        await gitWorkspaceState.repo.createBranch(localDefaultBranch, false);
+        assert.strictEqual(await tryGetDefaultBranch(gitWorkspaceState.repo), localDefaultBranch);
+    });
+
+    suiteTeardown(async () => {
+        // reset the configs
+        await cpUtils.executeCommand(undefined, testFolderPath, 'git', 'config', '--local', 'init.defaultBranch', '');
```
ejizba, line 44: If you're going to "reset the configs", it should be back to what they originally had set, which might not be empty. But even then I'm just not sure we want to be messing with people's global git config like this because suiteTeardown doesn't always happen (for example if you stop your debugging session during mid tests). My advice would be to skip the global config and only do local config (because that only applies to the random test folder, right?)

test/tryGetDefaultBranch.test.ts
```diff
@@ -0,0 +1,47 @@
...
+        if (!gitWorkspaceState.repo) {
+            throw new Error('Could not retrieve git repository.');
+        }
+
+        await gitWorkspaceState.repo.createBranch(globalDefaultBranch, false);
+        assert.strictEqual(await tryGetDefaultBranch(gitWorkspaceState.repo), globalDefaultBranch);
+    });
+
+    test('Workspace with default branch set in local config', async () => {
+        await cpUtils.executeCommand(undefined, testFolderPath, 'git', 'config', '--local', 'init.defaultBranch', localDefaultBranch);
```
ejizba, line 31: This is more related to the production code over the test itself, but I would not expect "init.defaultBranch" to be the main determiner of the default branch. I would expect it to be determined from the remote. For example, my default branch might be "master" in my local git config, but my default branch for all of our extension repos would be "main" by this point. I did a bit of googling and I'm not sure the best way to get that info, but here's one example:

https://stackoverflow.com/questions/28666357/git-how-to-get-default-branch

ejizba (changes requested):

test/tryGetDefaultBranch.test.ts
nturinski: I think that's fair.  The `init.defaultBranch` setting is relatively new (https://github.blog/2020-07-27-highlights-from-git-2-28/).  I'll put up a PR that will use the remote defaultBranch if it exists, otherwise fallback to this local setting.

test/global.test.ts
nturinski: Hmm, as far as I know, the CI machines don't have these configs set because that was one of the errors I was running into.  But good point about messing with someone running the tests locally.

Would it make sense to check if these configurations exist and only execute if we find that it doesn't?

ejizba: Yes that might make sense. Can show me the exact error you were hitting?

nturinski: https://stackoverflow.com/questions/9617336/how-to-resolve-git-did-not-exit-cleanly-exit-code-128-error-on-tortoisegit

The second answer refers to the error that I was running into.

ejizba: ...I mean yes I know how you fixed it, but that link doesn't really show the error. I was hoping you'd link me to a dev ops build log with the error

If I can see the error, I can maybe help find an alternate solution. Regardless, one idea is maybe you can set the config using "--local" instead of "--global"? Another idea is I think you can use [environment variables](https://git-scm.com/docs/git-commit#_commit_information) so maybe you can set those just for this test process

nturinski: Oh, you can see it in this build.  Test 2 has the error: https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=34481&view=logs&jobId=0da5d1d9-276d-5173-c4c4-9d4d4ed14fdb&j=0da5d1d9-276d-5173-c4c4-9d4d4ed14fdb&t=6e60e40e-3d3d-5182-ba10-afdf8e87d3dd

Though note that error is actually coming from us. https://github.com/microsoft/vscode-azurestaticwebapps/blob/main/src/errors.ts#L57-L59

nturinski: And I think the environment variables is a great idea-- I'll give that a try.

alexweininger:
Can I close this 😄 ?

nturinski:
😭

nturinski:
Damn, actually IIRC, this wasn't working because VS Code wasn't updating asynchronously with fse changing stuff. I wonder if using AzExtFsUtils would have fixed this issue... 🤔

Issue #287, opened 2021-01-21
Use git.openRepository so we can remove SimpleGit


nturinski:
As per https://github.com/microsoft/vscode/issues/111210#issuecomment-764405252

Issue #404, opened 2021-05-24
SWA extension for new SWA isn't detecting local git repo


diberry:
My .NET and JS testing for [this tutorial](https://docs.microsoft.com/en-us/azure/search/tutorial-csharp-deploy-static-web-app) gives me 1/8 steps when I select "Advanced" to create a SWA. 

@HeidiSteen uses the same tutorial steps but with a [Python version](https://github.com/Azure-Samples/azure-search-python-samples/tree/master/search-website) of the code and she gets either 12 or 13 steps - where she is asked for a repo, create a repo, etc. as if the extension doesn't recognize her local Git repo. What could be causing this? 

We need the steps to be proscriptive (completely repeatable) in the tutorial expecting 8 steps and not 12 or 13. 

@anthonychu @craigshoemaker Any ideas or help?

nturinski:
@diberry Thanks for reporting this.

The number of steps is dependent on whether or not the current local repo has a GitHub remote origin that the user has admin access to.  If @HeidiSteen is being asked to create a repo, it's likely because while she has a local repo, she needs to create a remote in GitHub for Static Web Apps to attach to and run GitHub Actions on.

The tutorial should definitely include either the step of creating/forking a repo from GitHub (8 step process) _or_ starting with a local repo with the expectation that a GitHub repo will be created (12 step process).  Also, the documentation is a bit old since we just changed the experience when the platform left preview.  [Here](https://github.com/microsoft/vscode-azurestaticwebapps/wiki/Creating-a-Static-Web-App) is a detailed description of the new creation process.

diberry:
@nturinski Can you work with @HeidiSteen internally to figure out where she went wrong. She successfully completed the same steps for the JS and .NET versions with a local fork of the repo which she owns - yet I didn't have any trouble and I don't own the origin repo, only my fork of it.

HeidiSteen:
@nturinski FYI that I'm using a remote origin that I own.  The repo I'm specifying in your wizard is https://github.com/HeidiSteen/azure-search-python-samples/tree/master/search-website -- which is my fork of the public repo.

nturinski:
I tried forking your repo https://github.com/nturinski/azure-search-python-samples.  I wasn't able to reproduce this with my workspace.

Our extension will only detect the remote repo if it has `origin` as the remote shortname.  Is it possible that your repo has named the remote something else?

HeidiSteen:
[...]

PR #408, opened 2021-05-27
Try to remove every instance of simple-git/git cpUtil calls

nturinski:
Fixes https://github.com/microsoft/vscode-azurestaticwebapps/issues/287

I think we should rely on the VS Code API for our git calls.  In this issue https://github.com/microsoft/vscode-azurestaticwebapps/issues/404, I'm pretty sure the user is being blocked because we don't have the proper permissions to run `git` commands in the temp terminal.  You can get more details as to why I think that from the issue.

Also, if the user doesn't have Git installed, then the Git API will throw an error and considered it in the "disabled" state.  Some of our checks/telemetry was to check for Git, so I removed some of that code since we're not running `git` manually anymore.


src/tree/SubscriptionTreeItem.ts
```diff
@@ -109,8 +109,10 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
         });
 
         const gotRemote: boolean = !!wizardContext.repoHtmlUrl;
+        wizardContext.telemetry.properties.gotRemote = String(gotRemote);
+
         wizardContext.fsPath = wizardContext.fsPath || getSingleRootFsPath();
-        addWorkspaceTelemetry(wizardContext);
```
nturinski, line 113: Some of the telemetry wasn't relevant anymore, specifically isGitProject and isGitInstalled since we validate that upfront now.  Since we were just keeping track of if we got a remote and the num of workspaces, I just moved that into the `SubscriptionTreeItem`.

src/commands/createRepo/RemoteShortnameStep.ts
```diff
@@ -29,7 +29,11 @@ export class RemoteShortnameStep extends AzureWizardPromptStep<IStaticWebAppWiza
                     try {
                         await cpUtils.executeCommand(undefined, undefined, 'git', 'check-ref-format', '--branch', cpUtils.wrapArgInQuotes(value));
                     } catch (err) {
-                        return localize('notValid', '"{0}" is not a valid remote shortname.', value);
+                        if (/is not a valid branch name/g.test(parseError(err).message)) {
```
nturinski, line 32: I think this is the only command there is no equivalent for.  If the error that comes back doesn't match, I think we should ignore it since it's just name validation anyway and it'll fail later in the pipeline if it's really invalid.

wwlorey (approved):

src/commands/createRepo/RemoteShortnameStep.ts
nturinski: Hmm, I think it would be better to have it for a better user experience.  Getting through all of the prompts and having it fail after we create the GitHub remote would be pretty annoying.  And being able to see invalid names before entering them is a better experience.  I think the only reason it wouldn't work is if there was something stopping git from executing, such as being in a directory where they do not have access rights.  I think I'll change the executeCommand to run in the `cwd` though rather than the temp directory (what it does by default)

Issue #406, opened 2021-05-25
Bug: If "origin" exists, but is not a repo on GitHub (or is a fake uri), extension doesn't create repository


nturinski:
The check to see whether or not the extension should create a repo for the user is based on whether or not we found an "origin" remote.  If it's a valid GitHub url, it should verify whether or not the user has admin access so there shouldn't be any problems.  However, if the repo doesn't return anything from GitHub, we proceed with the assumption that the user doesn't have a GitHub remote, however because we found an "origin", we don't actually create a repo for the user at create time.

PR #409, opened 2021-05-27
Check remote to determine whether or not we should create a repo

nturinski:
Fixes https://github.com/microsoft/vscode-azurestaticwebapps/issues/406

### DIFF
--- a/src/tree/SubscriptionTreeItem.ts
+++ b/src/tree/SubscriptionTreeItem.ts
@@ -67,10 +67,10 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
         }
 
         promptSteps.push(new SkuListStep());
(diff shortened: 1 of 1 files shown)


src/tree/SubscriptionTreeItem.ts
```diff
@@ -67,9 +67,10 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
         }
 
         promptSteps.push(new SkuListStep());
+        const gotRemote: boolean = !!wizardContext.repoHtmlUrl;
 
         // if the local project doesn't have a GitHub remote, we will create it for them
-        if (!wizardContext.originExists) {
+        if (!gotRemote) {
```
neelip, line 73: nit: renaming it to "hasRemote"

neelip (approved):

PR #412, opened 2021-06-03
Use the remote's defaultBranch over local settings

nturinski:


src/commands/createStaticWebApp/createStaticWebApp.ts
```diff
@@ -52,6 +51,7 @@ export async function createStaticWebApp(context: IActionContext & Partial<ICrea
             }
         }
 
+        await promptForDefaultBranch(context, verifiedWorkspace.repo);
```
ejizba, line 54: Instead of moving this line, it would be better if you passed `verifiedWorkspace` into `promptForDefaultBranch` and use `workspace.remoteRepo.default_branch` directly instead of referencing `context.branchData.name`. Then the order doesn't matter and plus `remoteRepo.default_branch` is a more intuitive name IMO

ejizba (approved):

PR #415, opened 2021-06-09
Poll for branch data on repo create

nturinski:
I noticed that in telemetry that we were having some users hitting the following error:
> Internal error: Expected value to be neither null nor undefined: branchData

I could see that this was _only_ happening when we didn't discover a GitHub remote so I figured that there's probably a timing issue when we set `wizardContext.branchData`.  GitHub's API claims that it always returns a branch, so I assumed it would throw an error if it didn't, but it seems like it's returning undefined/null.

### DIFF
--- a/src/commands/createRepo/RepoCreateStep.ts
+++ b/src/commands/createRepo/RepoCreateStep.ts
@@ -56,8 +56,17 @@ export class RepoCreateStep extends AzureWizardExecuteStep<IStaticWebAppWizardCo
 
             // getBranch will return undefined sometimes, most likely a timing issue so try to retrieve it for a minute
             const maxTimeout = Date.now() + 60 * 1000;
-            while (!wizardContext.branchData && Date.now() < maxTimeout) {
+            let numOfTries: number = 0;
+
+            while (true) {
+                numOfTries++;
+
                 wizardContext.branchData = (await client.repos.getBranch({ repo: newRepoName, owner: nonNullProp(wizardContext, 'orgData').login, branch: nonNullProp(branch, 'name') })).data;
+                if (wizardContext.branchData || Date.now() > maxTimeout) {
+                    wizardContext.telemetry.properties.getBranchAttempts = String(numOfTries);
+                    break;
+                }
+
                 await delay(2000);
             }
         } catch (err) {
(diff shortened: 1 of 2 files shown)


src/commands/createRepo/RepoCreateStep.ts
```diff
@@ -53,10 +54,20 @@ export class RepoCreateStep extends AzureWizardExecuteStep<IStaticWebAppWizardCo
             ext.outputChannel.appendLog(pushedBranch);
             progress.report({ message: pushedBranch });
 
-            wizardContext.branchData = (await client.repos.getBranch({ repo: newRepoName, owner: nonNullProp(wizardContext, 'orgData').login, branch: nonNullProp(branch, 'name') })).data;
+            // getBranch will return undefined sometimes, most likely a timing issue so try to retrieve it for a minute
+            const maxTimeout = Date.now() + 60 * 1000;
+            while (!wizardContext.branchData && Date.now() < maxTimeout) {
```
ejizba, line 59: I prefer to follow a pattern like this:
```typescript
while (true) {
    wizardContext.branchData = 
    if (got branchData or timed out) {
        break;
    } else {
        await delay
    }
}
```

Your current logic essentially has an extra `delay(2000)` after we already found the branch data, which is what I'm trying to avoid.

eslint doesn't like `while (true)`, but I think it's fine. Seems like we can disable the rule specifically for loops (as opposed to `if (true)`) which I might do.

src/commands/createRepo/RepoCreateStep.ts
```diff
@@ -53,10 +54,20 @@ export class RepoCreateStep extends AzureWizardExecuteStep<IStaticWebAppWizardCo
             ext.outputChannel.appendLog(pushedBranch);
             progress.report({ message: pushedBranch });
 
-            wizardContext.branchData = (await client.repos.getBranch({ repo: newRepoName, owner: nonNullProp(wizardContext, 'orgData').login, branch: nonNullProp(branch, 'name') })).data;
+            // getBranch will return undefined sometimes, most likely a timing issue so try to retrieve it for a minute
+            const maxTimeout = Date.now() + 60 * 1000;
```
ejizba, line 58: Any time we add "retry" logic I think we should add telemetry to see if the retry is actually helping. Basically, there's three cases to watch out for:
1. It succeeds on the first attempt
2. It succeeds on a retry
3. It times out

If only 1 and 3 are happening in telemetry, we know we can remove the retry logic to save users in case 3 from waiting a minute for no reason.

ejizba (approved):

PR #421, opened 2021-06-22
Try catch the call to Azure

nturinski:
I was seeing the error "Cannot find StaticSite with name ---" commonly enough in telemetry from `postCreateStaticWebApp`.  Users shouldn't even be seeing this error in the first place, but I believe that it's probably due to a timing issue (since this should be created by this point).

### DIFF
--- a/src/commands/createStaticWebApp/postCreateStaticWebApp.ts
+++ b/src/commands/createStaticWebApp/postCreateStaticWebApp.ts
@@ -31,8 +31,8 @@ export async function postCreateStaticWebApp(swaNode: StaticWebAppTreeItem): Pro
             const maxTime: number = Date.now() + 30 * 1000; // it can take a little for the action to queue in GitHub, wait for 30 seconds
 
             while (!deployActionNode) {
-                await productionEnv.actionsTreeItem.refresh(context);
                 try {
+                    await productionEnv.actionsTreeItem.refresh(context);
                     const actionTreeItems: ActionTreeItem[] = <ActionTreeItem[]>(await productionEnv.actionsTreeItem.loadAllChildren(context));
                     const filteredTreeItems: ActionTreeItem[] = actionTreeItems.filter(ti => { return ti.data.status !== Status.Completed; }); // only looking at on-going or queued jobs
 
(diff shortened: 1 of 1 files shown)


src/commands/createStaticWebApp/postCreateStaticWebApp.ts
```diff
@@ -32,25 +32,31 @@ export async function postCreateStaticWebApp(swaNode: StaticWebAppTreeItem): Pro
...
+                        }
+                    });
 
-                // the map will create an array of promises that will get resolved in parallel here
-                await Promise.all(promises);
+                    // the map will create an array of promises that will get resolved in parallel here
+                    await Promise.all(promises);
+                } catch (err) {
+                    // sometimes the web app is not ready and will throw "Cannot find StaticSite with name ---"
+                    // just ignore it, users shouldn't see this
```
ejizba, line 53: You're ignoring all errors, so I feel like it shouldn't reference a specific error in the comment
```suggestion
                    // ignore and retry
```

src/commands/createStaticWebApp/postCreateStaticWebApp.ts
```diff
@@ -32,25 +32,31 @@ export async function postCreateStaticWebApp(swaNode: StaticWebAppTreeItem): Pro
 
             while (!deployActionNode) {
                 await productionEnv.actionsTreeItem.refresh(context);
-                const actionTreeItems: ActionTreeItem[] = <ActionTreeItem[]>(await productionEnv.actionsTreeItem.loadAllChildren(context));
-                const filteredTreeItems: ActionTreeItem[] = actionTreeItems.filter(ti => { return ti.data.status !== Status.Completed; }); // only looking at on-going or queued jobs
+                try {
```
ejizba, line 35: Any particular reason line 34 isn't in the `try`? Seems like you'd want everything from the loop (except the timeout error) in there

nturinski: I didn't think that it'd be the source of an error, but you're right, may as well shove it in there as well

ejizba (approved):

PR #425, opened 2021-06-28
Add telemetry for workspace settings

nturinski:
I wanted to add some telemetry on how often workspace settings get used.  I'm seeing a lot of 'non-default' values, but it's very possible that users are just typing their own values.

I wanted to remove the settings from being added since with the current workspace requirements, it'll add the settings silently every time.  I'm thinking maybe we should have a notification or a setting to disable that behavior.  The main downside to it is that it makes the repo dirty and can potentially screw up the `git pull` we do for the workflow file since they're both existing in the .vscode folder.

I'm moving git pull to before the settings getting added for now to help mitigate that problem, but it is still weird to have your repo be "dirty" immediately after creating your SWA.

### DIFF
--- a/src/commands/createStaticWebApp/addLocationTelemetry.ts
+++ b/src/commands/createStaticWebApp/addLocationTelemetry.ts
@@ -7,16 +7,19 @@ import { IStaticWebAppWizardContext } from "./IStaticWebAppWizardContext";
 
 export function addLocationTelemetry(wizardContext: IStaticWebAppWizardContext, key: 'appLocation' | 'apiLocation' | 'outputLocation', defaultValue: string, valueFromSetting?: string): void {
     const value: string | undefined = wizardContext[key];
+    wizardContext.telemetry.properties[`${key}HasSetting`] = (!!valueFromSetting).toString();
+    wizardContext.telemetry.properties[`${key}MatchesSetting`] = (defaultValue === valueFromSetting).toString();
+
     let telemValue: string;
     if (value === undefined) {
         // no telemetry to add yet
         return;
     } else if (value.length === 0) {
         telemValue = 'empty';
-    } else if (value === valueFromSetting) {
-        telemValue = 'setting';
     } else if (value === defaultValue) {
         telemValue = 'default';
+    } else if (value === valueFromSetting) {
+        telemValue = 'setting';
     } else {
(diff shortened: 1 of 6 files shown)


src/commands/createStaticWebApp/addLocationTelemetry.ts
```diff
@@ -5,14 +5,16 @@
...
+export function addLocationTelemetry(wizardContext: IStaticWebAppWizardContext, key: 'appLocation' | 'apiLocation' | 'outputLocation', defaultValue: string, valueFromSetting?: string): void {
     const value: string | undefined = wizardContext[key];
     let telemValue: string;
     if (value === undefined) {
         // no telemetry to add yet
         return;
     } else if (value.length === 0) {
         telemValue = 'empty';
+    } else if (value === valueFromSetting) {
+        telemValue = 'setting';
```
ejizba, line 17: I think there's more information to be gained here if the setting info was part of it's own properties. I might add properties like 'appLocationHasSetting' and 'appLocationMatchesSetting'

It would help you answer questions like these:
- If the user had a setting, did they use it or ignore it?
- Does the user's setting match the default value?

src/tree/SubscriptionTreeItem.ts
```diff
@@ -123,6 +124,7 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
         await wizard.execute();
         context.showCreatingTreeItem(newStaticWebAppName);
 
+        await gitPull(nonNullProp(wizardContext, 'repo'));
         if (wizardContext.fsPath && hasRemote) {
             await updateWorkspaceSetting(appSubpathSetting, wizardContext.appLocation, wizardContext.fsPath);
```
ejizba, line 129: > I wanted to remove the settings from being added since with the current workspace requirements, it'll add the settings silently every time.

Does this modify settings.json even if the setting matches the default value? Ideally it does not

ejizba (approved):

src/tree/SubscriptionTreeItem.ts
nturinski: Yeah, good call.

nturinski: There are enough changes here that I'll handle this in another PR.

PR #426, opened 2021-06-29
Patch up some telemetry gaps

nturinski:
Specifically, there were a lot of gaps for when users were cancelling out of `Create Static Web App` and what users would do if there was no workspace available.

There are also a lot of gaps regarding default branches. We have a lot of instances where we couldn't detect a default branch at all (probably because we weren't looking at remote default branches at all),  but hopefully this should give us a bit more insight.

### DIFF
--- a/src/commands/createStaticWebApp/createStaticWebApp.ts
+++ b/src/commands/createStaticWebApp/createStaticWebApp.ts
@@ -22,6 +22,7 @@ import { setWorkspaceContexts } from './setWorkspaceContexts';
 
 export async function createStaticWebApp(context: IActionContext & Partial<ICreateChildImplContext> & Partial<IStaticWebAppWizardContext>, node?: SubscriptionTreeItem): Promise<StaticWebAppTreeItem> {
     if (!node) {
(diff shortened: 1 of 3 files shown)


src/commands/createStaticWebApp/createStaticWebApp.ts
```diff
@@ -22,6 +22,7 @@ import { setWorkspaceContexts } from './setWorkspaceContexts';
 
 export async function createStaticWebApp(context: IActionContext & Partial<ICreateChildImplContext> & Partial<IStaticWebAppWizardContext>, node?: SubscriptionTreeItem): Promise<StaticWebAppTreeItem> {
     if (!node) {
+        context.telemetry.properties.cancelStep = 'showTreeItemPicker';
```
ejizba, line 25: As a part of my funnel telemetry work, I'm looking into standardizing and enforcing `cancelStep` a bit more. If you want this reviewed ASAP let me know, but otherwise it might be easier to wait a few days to leverage that

ejizba: Never mind, gonna take me longer than I thought

ejizba (approved):

PR #427, opened 2021-06-29
Only update settings if they differ from the buildPreset settings

nturinski:
Rather than always updating the settings, I figured it would make more sense to only update _if_ it was different from the build preset.  If the current value is different from the current workspace setting, it'll overwrite it.  If it's the same, it'll just write the same value and doesn't appear as a change.


src/commands/createStaticWebApp/updateSwaLocationSettings.ts
```diff
@@ -0,0 +1,23 @@
...
+    if (wizardContext.fsPath) {
+        if (wizardContext.appLocation !== wizardContext.buildPreset?.appLocation ?? defaultAppLocation) {
+            await updateWorkspaceSetting(appSubpathSetting, wizardContext.appLocation, wizardContext.fsPath);
+        }
+
+        if (wizardContext.apiLocation !== wizardContext.buildPreset?.apiLocation ?? defaultApiLocation) {
+            await updateWorkspaceSetting(apiSubpathSetting, wizardContext.apiLocation, wizardContext.fsPath);
+        }
+
+        if (wizardContext.outputLocation !== wizardContext.buildPreset?.outputLocation)
```
ejizba, line 20: So when I said [this](https://github.com/microsoft/vscode-azurestaticwebapps/pull/425#discussion_r660765000), I forgot about the build preset thing. I was thinking there was one default for each. 

Is it still as worth it when the presets are different depending on what they select? I honestly don't entirely remember what effect the settings have

nturinski: So do you think it'd make more sense to just remove the setting all together?

The settings just pre-fill the app/api/output locations when the user creates a swa.  This is really only useful in the scenarios where a repo already exists (seems useful for scenarios) so that we can guide users to have the correct settings.

That being said, might not be that useful to set it automagically then?

ejizba: I don't know what the fix is, but yeah these settings don't sound super useful if I still get prompted. If I commit these to my repo, and we enforce that creating a swa has to use your current repo - wouldn't it make sense to use the settings without prompting? That being said, can I even create multiple swa's from the same repo?

nturinski: Yeah, I think the fear is if they made a SWA once and it failed, these settings would be saved and then the user would keep getting defaulted into using the same settings without knowing why the locations aren't showing up anymore.  I guess we _could_ update the settings only when we report that the SWA has successfully been built.

>  That being said, can I even create multiple swa's from the same repo?

That used to be the case, but they changed that restriction and now you can create multiple SWAs from the same repository.  I'm not sure what kind of trickle effects that'll have on workflows, but at least it's doable now.

The only real valuable use-case I see if I had a tutorial repo and you forked it, it now has all of those settings saved. 

The end goal for these settings were to set them based off their emulator/local usage so that way once they've configured it correctly, we could easily deploy it for them.

nturinski:
Discussed offline: Maintain support for settings; don't update the settings after creation.

src/commands/createStaticWebApp/BuildPresetListStep.ts
```diff
@@ -28,13 +28,8 @@ export class BuildPresetListStep extends AzureWizardPromptStep<IStaticWebAppWiza
...
 
-        if (pick.data) {
-            // prefill locations with the preset locations, but don't force the users to use them
-            context.presetAppLocation = pick.data.appLocation;
-            context.presetApiLocation = pick.data.apiLocation;
-            context.presetOutputLocation = pick.data.outputLocation;
-        }
-
+        // prefill locations with the preset locations, but don't force the users to use them
+        context.buildPreset = pick.data;
```
nturinski, line 32: I can revert this back to how it was, but this feels cleaner to me.

ejizba (approved):

Issue #370, opened 2021-04-29
RG related name doesn't verify if it is available in advanced create


v-xinda:
**OS:** Windows
**Build Version:** 20210428.17

**Repro Steps:**
1. Create a static web app with name `nicole-test` -> Default resource group name `nicole-test`  .
2. Delete the static web app.
3. Create a static web app with the same name `nicole-test` -> Default resource group name.
4. Check whether prompt an error message.

**Actual:**
Prompt an error message.
![image](https://user-images.githubusercontent.com/57662131/116519803-7ef6d100-a904-11eb-842e-6c2f85fcab7a.png)

**Suggestion:**
It would be better if the default resource group auto adds a suffix to distinguish their different when existing a same one, like below:
![group](https://user-images.githubusercontent.com/57662131/116520216-004e6380-a905-11eb-92e5-7b3c5518382a.gif)

PR #431, opened 2021-07-06
Fix related name task

nturinski:
Fixes https://github.com/microsoft/vscode-azurestaticwebapps/issues/370

Is there a better way to check for `isRelatedName` available?  Maybe this bug fix can wait until after release and I'll just expose `ResourceGroupListStep.getRelatedName`?

### DIFF
--- a/src/commands/createStaticWebApp/StaticWebAppNameStep.ts
+++ b/src/commands/createStaticWebApp/StaticWebAppNameStep.ts
@@ -4,7 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import * as path from 'path';
-import { AzureNameStep, IAzureNamingRules } from "vscode-azureextensionui";
+import { AzureNameStep, IAzureNamingRules, ResourceGroupListStep, resourceGroupNamingRules } from "vscode-azureextensionui";
 import { ext } from "../../extensionVariables";
 import { localize } from "../../utils/localize";
 import { nonNullProp } from '../../utils/nonNull';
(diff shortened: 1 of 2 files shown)


src/tree/SubscriptionTreeItem.ts
```diff
@@ -118,7 +118,7 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
         const newStaticWebAppName: string = nonNullProp(wizardContext, 'newStaticWebAppName');
 
         if (!context.advancedCreation) {
-            wizardContext.newResourceGroupName = newStaticWebAppName;
+            wizardContext.newResourceGroupName = await StaticWebAppNameStep.getRelatedName(wizardContext, newStaticWebAppName);
```
ejizba, line 121: You're supposed to just do `await wizardContext.relatedNameTask;` here

Here's the app service example:
https://github.com/microsoft/vscode-azureappservice/blob/main/src/tree/SubscriptionTreeItem.ts#L113

src/commands/createStaticWebApp/StaticWebAppNameStep.ts
```diff
@@ -38,17 +38,21 @@ export class StaticWebAppNameStep extends AzureNameStep<IStaticWebAppWizardConte
         return !wizardContext.newStaticWebAppName;
     }
 
-    public async isRelatedNameAvailable(wizardContext: IStaticWebAppWizardContext, name: string): Promise<boolean> {
-        return await this.isSwaNameAvailable(wizardContext, name);
+    protected async isRelatedNameAvailable(wizardContext: IStaticWebAppWizardContext, name: string): Promise<boolean> {
+        if (!wizardContext.newStaticWebAppName) {
```
ejizba, line 42: I don't think this will ever be true? I checked SiteNameStep and confirmed it's only checking names for _related_ resources, not the original resource itself

https://github.com/microsoft/vscode-azuretools/blob/main/appservice/src/createAppService/SiteNameStep.ts#L76

nturinski: Oh interesting, I may have totally misunderstood this function's use.  Let me try that real quickly

nturinski: Ah, I may be using it wrong, but this is when this would be true: https://github.com/microsoft/vscode-azurestaticwebapps/pull/431/files#diff-913ec2c5bfe37c6014c9ea09a862e04a5b85eab00457ceafb2cf79a43915ca6bR27

ejizba (approved):

PR #441, opened 2021-07-09
Use action-specific user input and standardize name of context

nturinski:
Same as microsoft/vscode-azurestorage#987

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,8 +15,8 @@
                 "git-url-parse": "^11.1.2",
                 "moment": "^2.27.0",
                 "open": "^8.0.4",
-                "vscode-azureappservice": "^0.81.1",
-                "vscode-azureextensionui": "^0.44.3",
+                "vscode-azureappservice": "^0.83.0",
+                "vscode-azureextensionui": "^0.45.0",
                 "vscode-nls": "^4.1.1",
                 "yaml": "^1.10.2"
             },
@@ -40,7 +40,7 @@
                 "ts-node": "^7.0.1",
                 "typescript": "^3.9.7",
                 "vsce": "^1.87.1",
-                "vscode-azureextensiondev": "^0.9.4",
+                "vscode-azureextensiondev": "^0.9.5",
                 "vscode-test": "^1.5.2",
                 "webpack": "^5.28.0",
                 "webpack-cli": "^4.6.0"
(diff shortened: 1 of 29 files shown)


package-lock.json
```diff
@@ -11043,9 +11043,9 @@
             }
         },
         "node_modules/vscode-azureextensiondev": {
-            "version": "0.9.2",
-            "resolved": "https://registry.npmjs.org/vscode-azureextensiondev/-/vscode-azureextensiondev-0.9.2.tgz",
-            "integrity": "sha512-2TsA5ct2Zzm5ydHv2Vj3Sko6Bm1BRHRc+qOsKDmgL+XQfblY0tyKBSb93HGJVfu4+HKU6MIV7ZUib/RfKOY5Dg==",
+            "version": "0.9.5",
```
neelip, line 11046: Why does this say 0.9.5 instead of 0.9.4?

nturinski: That's a very good question 🤔

neelip (approved):
I think the "vscode-azureextensiondev" package is causing the build errors.

package-lock.json
ejizba: Because the carot in front of "^0.9.4" allows for higher versions as long as it's not a breaking change

src/extension.ts
```diff
@@ -19,9 +19,7 @@ export async function activateInternal(context: vscode.ExtensionContext, perfSta
     ext.ignoreBundle = ignoreBundle;
     ext.outputChannel = createAzExtOutputChannel('Azure Static Web Apps', ext.prefix);
     context.subscriptions.push(ext.outputChannel);
-    ext.ui = new AzureUserInput(context.globalState);
 
-    registerUIExtensionVariables(ext);
```
ejizba, line 24: We still need this line

src/extension.ts
```diff
@@ -19,9 +19,7 @@ export async function activateInternal(context: vscode.ExtensionContext, perfSta
     ext.ignoreBundle = ignoreBundle;
     ext.outputChannel = createAzExtOutputChannel('Azure Static Web Apps', ext.prefix);
     context.subscriptions.push(ext.outputChannel);
-    ext.ui = new AzureUserInput(context.globalState);
```
ejizba, line 22: If you're going to remove the extension-scoped `ui` in this PR, you also need to remove it  from `extensionVariables.ts` and you need to bump the ui/appservice npm packages

I did it in a separate PR in the case of storage (https://github.com/microsoft/vscode-azurestorage/pull/988), but up to you if you want to combine it here

Issue #439, opened 2021-07-08
`isSwaNameAvailable` doesn't accurately report availability


nturinski:
Because the call `staticSites.getStaticSite(resourceGroup, site);`  uses the resourceGroup name, this check isn't all that accurate since it is specific to the resource group.  If we want to keep sites subscription-unique, I should be using `staticSites.list()` instead.  

However, this won't block users as the resourceGroup name will be incremented if it already exists.  Just some weird, inconsistent behavior.

![image](https://user-images.githubusercontent.com/5290572/124975598-089cdc00-dfe3-11eb-82bb-cee03b60c4ff.png)

![image](https://user-images.githubusercontent.com/5290572/124975627-0f2b5380-dfe3-11eb-884f-f95ec40b4fd1.png)

![image](https://user-images.githubusercontent.com/5290572/124975650-15213480-dfe3-11eb-9d43-292ee4159efc.png)

nturinski:
Actually, this can still cause issues if you select a resource group that has an app with the same name.  This is only possible with advanced create though.

PR #447, opened 2021-07-15
Move `ResourceGroupList` step before `StaticWebAppNamePrompt` step to validate unique name

nturinski:
Fixes https://github.com/microsoft/vscode-azurestaticwebapps/issues/439


src/commands/createStaticWebApp/StaticWebAppNameStep.ts
```diff
@@ -62,14 +66,11 @@ export class StaticWebAppNameStep extends AzureNameStep<IStaticWebAppWizardConte
     }
 
     private async isSwaNameAvailable(context: IStaticWebAppWizardContext, resourceName: string): Promise<boolean> {
-        // Static Web app names must be unique to the current resource group.
-        try {
-            await context.client.staticSites.getStaticSite(resourceName, resourceName);
-            return false;
+        // Our implementation is that Static Web app names must be unique to the subscription.
+        listOfSites = listOfSites || await uiUtils.listAll(context.client.staticSites, context.client.staticSites.list());
```
wwlorey, line 70: It seems possible that a recently created SWA's name will show as available if we use the cached `listOfSites` instead of always getting a fresh list

ejizba, line 70: I'm having a hard time deciding on this PR. I'm just not sure listing all static sites is a robust or efficient way to solve the root cause. We've discussed a few alternative solutions - what were those and why is this one better?

ejizba: Also the cached list won't be accurate if they select a different subscription

ejizba (commented):
Lol ftr, I was already in the process of writing my comments when Mr William swooped in just now 🦅💨

src/commands/createStaticWebApp/StaticWebAppNameStep.ts
nturinski: The main alternative was to put the resource group step first so we can be certain whether or not the swa is within that rg.

Since we don't display swa with duplicate names very well, enforcing a unique name helps mitigate that problem.
image

ejizba: > Since we don't display swa with duplicate names very well

Just to be clear, we're talking about "ReactSite", right?  Seems like the description already differentiates them pretty well. I believe the "Invalid" one is an unrelated issue. 

> The main alternative was to put the resource group step first so we can be certain whether or not the swa is within that rg.

Regardless of the final solution, I think we should make sure it only affects advanced create. Basic create is fine as-is since we intelligently pick the rg for them anyways. Now, in terms of advanced create, we could:
1. Enforce swa name uniqueness at the subscription level (aka your current changes, except only for advanced create)
1. Move the rg step
1. Filter out rgs that would have a conflicting name
1. Validate the name/rg immediately after they pick an rg and show a warning/error at that point
1. Don't even try to validate uniqueness and just go with whatever error is thrown by the platorm

I think I lean towards 2 or 4. I'm just trying to balance code complexity and perf impacts for something that is really just an edge case bug. Also, it would be nice if we're consistent with the portal in terms of name uniqueness rules

nturinski: I went with option 2.  Just felt the best to me.

ejizba (approved):

PR #454, opened 2021-07-23
Open GitHub Actions logs from tree

nturinski:
Rather than using the scheme, I used `-gh.log` suffix for the documentProvider.  However in replace of this PR: https://github.com/microsoft/vscode-azuretools/pull/962#event-5057352597, I will need to expose the uri of the ReadOnly documents that we open.  Right now, I'm just hard-coding the private variable.


src/commands/github/GitHubLogFoldingProvider.ts
```diff
@@ -0,0 +1,32 @@
...
+import { CancellationToken, FoldingContext, FoldingRange, FoldingRangeKind, FoldingRangeProvider, TextDocument } from "vscode";
+import { getLogInfo } from "./vscode-github-actions/logInfoProvider";
+
+export default class GitHubLogFoldingProvider implements FoldingRangeProvider {
+
+    public async provideFoldingRanges(
+        document: TextDocument,
+        _: FoldingContext,
+        _token: CancellationToken
+    ): Promise<FoldingRange[]> {
```
alexweininger, line 15: I think this could be on a single line

src/commands/github/GitHubLogFoldingProvider.ts
```diff
@@ -0,0 +1,32 @@
...
+    private async getRegions(document: TextDocument): Promise<FoldingRange[]> {
+        const logInfo = getLogInfo(document.uri);
+        if (!logInfo) {
+            return [];
+        }
+
+        return logInfo.sections.map(
+            s =>
+                new FoldingRange(s.start, s.end, FoldingRangeKind.Region)
+        );
```
alexweininger, line 27: Formatting

alexweininger (changes requested):

src/extension.ts
```diff
@@ -32,6 +33,7 @@ export async function activateInternal(context: vscode.ExtensionContext, perfSta
         ext.tree = new AzExtTreeDataProvider(accountTreeItem, 'staticWebApps.loadMore');
         ext.treeView = vscode.window.createTreeView('staticWebApps', { treeDataProvider: ext.tree, showCollapseAll: true, canSelectMany: true });
         context.subscriptions.push(ext.treeView);
+        context.subscriptions.push(vscode.languages.registerFoldingRangeProvider({ "pattern": '**/*gh.log' }, new GitHubLogFoldingProvider()));
```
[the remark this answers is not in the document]
nturinski: I think it doesn't affect it either way, but might as well be explicit. 👍

nturinski:
Do you think exposing the uri for `ReadOnlyContent` would be okay?

I could use something else as the id, but because we receive the document in `provideFoldingRanges` it's definitely the most convenient.  I could just use the file name, but I would need to make the file names more unique by adding their full id or owner/repo names.

Also, planning to add tests, but I wanted to ask about the uri stuff first.

alexweininger:
> Do you think exposing the uri for `ReadOnlyContent` would be okay?
> 
> I could use something else as the id, but because we receive the document in `provideFoldingRanges` it's definitely the most convenient. I could just use the file name, but I would need to make the file names more unique by adding their full id or owner/repo names.
> 
> Also, planning to add tests, but I wanted to ask about the uri stuff first.

IMO exposing the uri for `ReadonlyContent` is fine. It's simpler than using the filenames and allows more freedom when setting filenames.

package.json
```diff
@@ -294,6 +299,11 @@
                     "when": "view == staticWebApps && viewItem == azureStaticEnvironment",
                     "group": "3@2"
                 },
+                {
+                    "command": "staticWebApps.openGitHubLog",
```
ejizba, line 303: I tried this out and was getting a lot of `ServerError`, I believe a 410 status code. Are you seeing that as well? Some logs hit it and others worked just fine.

src/commands/github/foldingProvider/foldingRangesMap.ts
```diff
@@ -0,0 +1,17 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { FoldingRange, Uri } from "vscode";
+
+const foldingRangeMap: { [key: string]: FoldingRange[] } = {}
+
+export function getFoldingRanges(uri: Uri): FoldingRange[] {
+    return foldingRangeMap[uri.fsPath];
```
ejizba, line 11: > Do you think exposing the uri for ReadOnlyContent would be okay?

I'm okay with it if you want a quick solution. It's not ideal because it's exposing an internal detail of how `openReadonlyContent` works. For example, we currently stick the unique `idHash` string in the path itself, but `idHash` could've easily gone in the query instead. If we ever switched it, your `fsPath` referenced here would no longer be unique.

I'm starting to feel like the ideal solution might be to create your own version of `ReadOnlyContentProvider`. I've changed my mind for a few reasons
1. It's not as much code as I originally remembered
2. The GitHub logic seems unique enough that it might be worth it. Specifically, you can be inspired by `ReadOnlyContentProvider`, but I wouldn't wanting you copying it over as-is. Ideally you create a new `TextDocumentContentProvider ` specifically for GitHub logs 
3. `openReadOnlyContent` is actually used in quite a few extensions, so I like keeping it simple and clean

package.json
nturinski: I think this is due to log files eventually getting deleted on GitHub.  I think because a lot of the builds I tested this with were made recently, I didn't run into this at all. Good catch.

ejizba: Hmm I mean mine was a brand new SWA. Maybe _too_ new? Like, I wonder if there's timing issues viewing logs while the jobs are in progress

nturinski: That sounds likely.  I'll have do some testing around that and handle it appropriately.

src/commands/github/jobLogs/openGitHubLog.ts
```diff
@@ -0,0 +1,40 @@
...
+import { IActionContext, parseError } from 'vscode-azureextensionui';
+import { ext } from '../../../extensionVariables';
+import { JobTreeItem } from '../../../tree/JobTreeItem';
+import { StepTreeItem } from '../../../tree/StepTreeItem';
+import { localize } from '../../../utils/localize';
+import { createFoldingRanges } from './createFoldingRanges';
+import { openGitHubLogContent } from './GitHubLogContentProvider';
+import { LogState, parseGitHubLog } from './parseGitHubLog';
+
+export async function openGitHubLog(context: IActionContext, node?: StepTreeItem): Promise<void> {
```
alexweininger, line 17: I find displaying an empty log file is a bit weird behavior. Users could mistake this for a bug. I noticed this has only happened for me when a job is skipped. However, I assume jobs that do run are capable of not producing any logs as well.

My ideas for possible solutions:
* Show a notification instead, with text similar to "This job was skipped" or "This job did not output any logs"
* Don't show the "Open GitHub Action Log" context menu item for jobs with no logs.

### Example of empty log file:
empty log file 

### This is what GitHub displays:
github job skipped

test/parseGitHubLog.test.ts
```diff
@@ -0,0 +1,1114 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See LICENSE.md in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as assert from 'assert';
+import { FoldingRange } from 'vscode';
+import { createFoldingRanges, LogState, parseGitHubLog } from '../extension.bundle';
+
+suite('Parse response from GitHub downloadJobLogsForWorkflowRun API', () => {
```
alexweininger, line 10: Tests look good. Only extra test I can think of would be ensuring `parseGitHubLog` behaves correctly if a log has invalid pairs of group/endgroup tags. But I'm not sure how far is too far when it comes to handling bad log files.

nturinski: I was considering that too, but I wasn't sure if we should include faulty logs in the dataset.  The tests are verify that we don't break anything with changes rather than if GitHub sends us bad logs. So I don't really know what the call is

src/commands/github/jobLogs/openGitHubLog.ts
nturinski: I think if we find nothing, I'll write "No logs were found" or something to that effect since there are times where nothing is displayed, but they weren't skipped (like if you view the logs before that step has run)

nturinski:
@alexweininger Just a reminder, but @ejizba said feel free to approve and that he would comment later if he catches any issues with the PR.

alexweininger (approved):

PR #456, opened 2021-07-23
Add no-restricted-imports rule to help prevent wrong ext imports

alexweininger:
To prevent wrong ext imports I added the [no-restricted-imports](https://eslint.org/docs/rules/no-restricted-imports) rule.

```json
"no-restricted-imports": ["error", {
    "paths": ["vscode-azureappservice/out/src/extensionVariables", "**/*/extension.bundle"] 
}],
```
example:

Screen Shot 2021-07-23 at 11 47 16 AM


.eslintrc.js
```diff
@@ -22,6 +22,7 @@ module.exports = {
         "no-useless-escape": "off",
         "no-inner-declarations": "off",
         "no-constant-condition": ["error", { "checkLoops": false }],
+        "no-restricted-imports": ["error", { "paths": ["vscode-azureappservice/out/src/extensionVariables", "**/*/extension.bundle"] }],
```
ejizba, line 25: So for `**/*/extension.bundle` I agree we want that for `src`, but we actually want the opposite for `test` (people _should_ use "extension.bundle" and _not_ `src` because we want tests running against the webpacked code on the build machines)

It looks like this repo doesn't have a test-specific "eslintrc.js" yet, so I'm curious why the build passes when we have extension.bundle imports here: https://github.com/microsoft/vscode-azurestaticwebapps/blob/main/test/selectBuildConfigs.test.ts#L9

ejizba (commented):
Neato! @wwlorey was looking into making a shared set of rules (maybe in the dev package), and it'd be nice if we could include this rule

.eslintrc.js
alexweininger: Oops. I need to use "patterns" instead of "paths" to use globs - so the build passed because the `**/*extension.bundle` rule wasn't actually working.

I'll create a test specific .eslintrc to prevent the error appearing in tests.

test/.eslintrc.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
+    "extends": [
+        "../.eslintrc.js"
+    ],
+    "rules": {
+        "no-restricted-imports": "off",
```
ejizba, line 6: Would something like this work to prevent people from importing from src?

```suggestion
        "no-restricted-imports": ["error", { "patterns": ["**/src/"] }],
```

I've accidentally imported from src instead of extension.bundle.ts several times and it's annoying because you don't know there's a problem until you push to the build machine and then the tests fail

ejizba (approved):

alexweininger:
Added rule to restrict importing from src in tests.

Screen Shot 2021-07-23 at 1 01 23 PM

And also found and added the [no-internal-modules](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-internal-modules.md) rule from eslint-plugin-import.

PR #461, opened 2021-07-28
Add "Create from template" option when there are no workspace folders

alexweininger:
When a user has no workspace, a new option for "Create from template" is available that allows the users to select a repo from a list we pull from the templates available on https://github.com/staticwebdev. After entering a repo name we create the repo on GitHub for them and clone it.

Note: I omitted the GitHub org prompt because we currently don't have a way to check if the user has permissions to create a repository on the org. I looked into how to do this and it seemed like more work than I wanted to include in this PR.

Screenshots: 

Screen Shot 2021-07-27 at 4 57 32 PM

Pick template prompt:
Screen Shot 2021-07-27 at 4 58 19 PM

Get repo name prompt:
image

### DIFF
--- a/src/commands/github/createOctokitClient.ts
+++ b/src/commands/github/createOctokitClient.ts
@@ -14,6 +14,12 @@ export async function createOctokitClient(context: IActionContext & Partial<ISta
     return new Octokit(
         {
             userAgent: appendExtensionUserAgent(),
-            auth: token
+            auth: token,
+            /**
+             * 'baptiste' preview needed to work with template repos
+             * see https://docs.github.com/en/rest/overview/api-previews#create-and-use-repository-templates
+             * and https://developer.github.com/changes/2019-07-16-repository-templates-api/
+             */
+            previews: ['baptiste']
         });
 }
--- a/src/utils/templateUtils.ts
+++ b/src/utils/templateUtils.ts
@@ -3,16 +3,28 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import { Octokit } from '@octokit/rest';
 import { IActionContext, IAzureQuickPickItem } from 'vscode-azureextensionui';
 import { RepoNameStep } from '../commands/createRepo/RepoNameStep';
 import { GitHubOrgListStep } from '../commands/createStaticWebApp/GitHubOrgListStep';
 import { IStaticWebAppWizardContext } from '../commands/createStaticWebApp/IStaticWebAppWizardContext';
-import { RepoData } from '../gitHubTypings';
-import { getTemplateRepos } from './gitHubUtils';
+import { createOctokitClient } from '../commands/github/createOctokitClient';
+import { templateReposUsername } from '../constants';
+import { RepoData, RepoResponse } from '../gitHubTypings';
 import { localize } from './localize';
 
+export async function getTemplateReposFromGitHub(context: IActionContext): Promise<RepoData[]> {
+    const client: Octokit = await createOctokitClient(context);
+
+    const allRepositories: RepoResponse = await client.repos.listForUser({
+        username: templateReposUsername
+    });
+
+    return allRepositories.data.filter((repo) => repo.is_template);
+}
+
 export async function pickTemplate(context: IActionContext): Promise<RepoData> {
-    const templateRepos: RepoData[] = await getTemplateRepos(context);
+    const templateRepos: RepoData[] = await getTemplateReposFromGitHub(context);
 
     const placeHolder: string = localize('chooseTemplatePrompt', 'Choose a static web app template for the new repository.');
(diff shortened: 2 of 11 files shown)


src/commands/github/createOctokitClient.ts
```diff
@@ -14,6 +14,12 @@ export async function createOctokitClient(context: IActionContext & Partial<ISta
...
         {
             userAgent: appendExtensionUserAgent(),
-            auth: token
+            auth: token,
+            /**
+             * 'baptiste' preview needed to work with template repos
+             * see https://docs.github.com/en/rest/overview/api-previews#create-and-use-repository-templates
+             * and https://developer.github.com/changes/2019-07-16-repository-templates-api/
+             */
+            previews: ['baptiste']
```
nturinski, line 23: I don't think it's a huge concern, but just wanted to bring it up to give it a proper consideration:
> API previews let you try out new APIs and changes to existing API methods before they become part of the official GitHub API.
>
> During the preview period, we may change some features based on developer feedback. If we do make changes, we'll announce them on the developer blog without advance notice.

I think that the preview tag `mercy-preview` (Repository topics) would be the only one I'd be wary of.  If we are worried about this, we can always have a separate Octokit specifically for creating from template.

Also their preview names are pretty funny.  Both of the ones I looked at were Overwatch specific 🤣

src/utils/templateUtils.ts
```diff
@@ -0,0 +1,53 @@
...
+    const templateRepos: RepoData[] = await getTemplateReposFromGitHub(context);
+
+    const placeHolder: string = localize('chooseTemplatePrompt', 'Choose a static web app template for the new repository.');
+    const picks: IAzureQuickPickItem<RepoData>[] = templateRepos.map((repo) => { return { label: repo.name, data: repo }; });
+    const pick: IAzureQuickPickItem<RepoData> = await context.ui.showQuickPick(picks, { placeHolder, suppressPersistence: true });
+
+    return pick.data;
+}
+
+export async function promptForTemplateRepoName(context: IActionContext, template: RepoData): Promise<string> {
```
nturinski, line 36: I think you should still make these all wizard prompt steps.  Even if we don't end up putting it in the normal basic create flow, it would still be handy to have the other wizard frills along with it (number of steps, ability to go back, etc.)

So you could have `PickTemplateStep.ts` and `TemplateRepoNameStep.ts`

src/utils/templateUtils.ts
```diff
@@ -0,0 +1,53 @@
...
+import { IStaticWebAppWizardContext } from '../commands/createStaticWebApp/IStaticWebAppWizardContext';
+import { createOctokitClient } from '../commands/github/createOctokitClient';
+import { templateReposUsername } from '../constants';
+import { RepoData, RepoResponse } from '../gitHubTypings';
+import { localize } from './localize';
+
+export async function getTemplateReposFromGitHub(context: IActionContext): Promise<RepoData[]> {
+    const client: Octokit = await createOctokitClient(context);
+
+    const allRepositories: RepoResponse = await client.repos.listForUser({
```
nturinski, line 19: Would a user be able to see the Microsoft templates from this list?  Or would we have to pass in the url of the Microsoft templates specifically?

alexweininger: I'm not sure what you mean by

> Would a user be able to see the Microsoft templates from this list?

Line 19 gets all of the repositories on the https://github.com/staticwebdev account. And then I return only the template repositories on line 23. The only place where the templates are displayed for users are in the quick pick where they select a template.

src/utils/workspaceUtils.ts
```diff
@@ -73,6 +79,17 @@ export async function getWorkspaceFolder(context: IActionContext): Promise<Works
             // don't wait
             void commands.executeCommand('vscode.openFolder', uri[0]);
             context.telemetry.properties.noWorkspaceResult = openExistingProject;
+        } else {
+            const wizardContext: Partial<IStaticWebAppWizardContext> & IActionContext = context;
```
nturinski, line 83: It's a little unintuitive, but you should be adding the new steps to the wizard here: https://github.com/microsoft/vscode-azurestaticwebapps/blob/main/src/tree/SubscriptionTreeItem.ts#L53

The way you have it now, it doesn't actually integrate into the wizard properly so you can't leverage the current flow or use the back button.

That way you won't have to change some steps to accept the Partial contexts either.  

Based off of the result of `getWorkspaceFolder`, we can change the creation flow in https://github.com/microsoft/vscode-azurestaticwebapps/blob/main/src/commands/createStaticWebApp/createStaticWebApp.ts to skip all of our validation code, create the SWA based off the template, and then ask to do the clone/reload at the end.

alexweininger:
Closing because we decided to go in a slightly different direction with this feature.

src/commands/github/createOctokitClient.ts
alexweininger: Coming back to this... Just curious why you are concerned about using the mercy-preview?

PR #466, opened 2021-08-06
Refactor getWorkspaceFolder

alexweininger:
Needed to refactor this for the create from template feature in a future PR. You can see how these changes are used with the create from template feature in the `alex/createFromTemplate` branch.

[Preview the PR that depends on this refactor](https://github.com/microsoft/vscode-azurestaticwebapps/compare/alex/refactorGetWorkspaceFolder...alex/createFromTemplate?expand=1)

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -21,11 +21,7 @@ export const configFileName: string = 'staticwebapp.config.json';
 
 export const showActionsMsg: MessageItem = { title: localize('openActions', 'Open Actions in GitHub') };
 
-export const cloneProjectMsg: MessageItem = { title: localize('cloneProject', 'Clone project from GitHub') };
-
 export const openExistingProject = 'openExistingProject';
-
-export const openExistingProjectMsg: MessageItem = { title: localize(openExistingProject, 'Open existing project') };
 
 export const onlyGitHubSupported: string = localize('onlyGitHubSupported', 'Only Static Web Apps linked to GitHub are supported at this time.');
 
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -21,6 +21,12 @@ export const configFileName: string = 'staticwebapp.config.json';
 
 export const showActionsMsg: MessageItem = { title: localize('openActions', 'Open Actions in GitHub') };
 
+export const cloneProjectMsg: MessageItem = { title: localize('cloneProject', 'Clone project from GitHub') };
+
+export const openExistingProject = 'openExistingProject';
+
+export const openExistingProjectMsg: MessageItem = { title: localize(openExistingProject, 'Open existing project') };
+
 export const onlyGitHubSupported: string = localize('onlyGitHubSupported', 'Only Static Web Apps linked to GitHub are supported at this time.');
 
 export const githubAuthProviderId: string = 'github';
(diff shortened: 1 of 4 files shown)


src/utils/workspaceUtils.ts
```diff
@@ -47,49 +46,39 @@ export async function selectWorkspaceItem(context: IActionContext, placeHolder:
...
-            throw new UserCancelledError(selectProjectFolder);
+            throw new UserCancelledError(selectAppFolder);
         }
-
         context.telemetry.properties.noWorkspaceResult = 'multiRootProject';
+        return folder;
     }
+}
+
+export async function showNoWorkspacePrompt(context: IActionContext): Promise<MessageItem> {
```
nturinski, line 68: I think we should handle setting the contexts here for whatever the user chooses.  This command is specific to `createStaticApp` so we shouldn't need to use the generic IActionContext for it (which is why I assumed that you didn't set the contexts here)

alexweininger: Are you referring to the telemetry contexts? If you are, then I can still use IActionContext.

When I add the "Create from template" option, I will change it from the generic so that I can set the context inside this function for if the user picks "Create from template".

nturinski: Oh no,  I was referring to these lines:
https://github.com/microsoft/vscode-azurestaticwebapps/pull/466/files#diff-71e8df3edc0b95484e8ea0cfb8fb79ffc32b5160dd9e62fb6717d0e7ee711b07R36-R45

It's kind of weird to me that this isn't handled in the same function as the prompt.

alexweininger: Ok, makes sense. I went ahead and moved those into the prompt, including `throw new NoWorkspaceError();`. When I implement the create from template option, I'll have to make sure to not throw that error when they select create from template.

src/constants.ts
```diff
@@ -21,6 +21,12 @@ export const configFileName: string = 'staticwebapp.config.json';
 
 export const showActionsMsg: MessageItem = { title: localize('openActions', 'Open Actions in GitHub') };
 
+export const cloneProjectMsg: MessageItem = { title: localize('cloneProject', 'Clone project from GitHub') };
```
nturinski, line 24: nit: No need to export these as constants since they're only used in one file (unless the next PR uses them too?)

src/utils/workspaceUtils.ts
```diff
@@ -47,49 +48,47 @@ export async function selectWorkspaceItem(context: IActionContext, placeHolder:
...
+        return workspace.workspaceFolders[0];
     } else {
-        const selectProjectFolder: string = 'selectProjectFolder';
-        const placeHolder: string = localize(selectProjectFolder, 'Select the folder containing your SWA project');
-        folder = await window.showWorkspaceFolderPick({ placeHolder });
-        if (!folder) {
-            throw new UserCancelledError(selectProjectFolder);
+        const selectAppFolder: string = 'selectAppFolder';
+        const folder = await selectWorkspaceFolder(context, localize(selectAppFolder, 'Select folder with your app'));
+        if (folder === undefined) {
```
nturinski, line 61: Shouldn't need this check anymore because `selectWorkspaceFolder` guarantees a string.  If the user cancels, the `UserCancelledError` is thrown by `context.ui.showQuickPick`

nturinski (approved):

PR #467, opened 2021-08-06
Add details about pricing options

### LINKED ISSUE #458: [Suggestion] Add details about pricing options
When creating a SWA in advanced mode, you're met with this prompt about the pricing options:

image

If you're not familiar with SWA pricing tiers, there's no cue here to help you make a choice. 
What about adding a one-line description some perks included in each options?

It could be something like this (from a different extension):
image

As a side note, I would argue that "Select a sku" might be unclear if you're unfamiliar with Azure namings, "Select pricing option" would be more understandable by everyone and it's also the terms used on the ACOM page: https://azure.microsoft.com/en-us/pricing/details/app-service/static/

alexweininger:
Fixes #458 

image

### DIFF
--- a/src/commands/createStaticWebApp/SkuListStep.ts
+++ b/src/commands/createStaticWebApp/SkuListStep.ts
@@ -5,7 +5,6 @@
 
 import { WebSiteManagementModels } from '@azure/arm-appservice';
 import { AzureWizardPromptStep, IAzureQuickPickItem } from 'vscode-azureextensionui';
-import { freeSkuDescription, freeSkuDetail, standardSkuDescription, standardSkuDetail } from '../../constants';
 import { localize } from '../../utils/localize';
 import { nonNullProp } from '../../utils/nonNull';
 import { IStaticWebAppWizardContext } from "./IStaticWebAppWizardContext";
@@ -35,14 +34,14 @@ export class SkuListStep extends AzureWizardPromptStep<IStaticWebAppWizardContex
             {
                 name: 'Free',
                 tier: 'Free',
-                description: freeSkuDescription,
-                detail: freeSkuDetail
+                description: localize('freeSkuDescription', 'For hobbies/personal projects'),
+                detail: localize('freeSkuDetail', 'Free SSL, 2 Custom Domains')
             },
             {
                 name: 'Standard',
                 tier: 'Standard',
-                description: standardSkuDescription,
-                detail: standardSkuDetail
+                description: localize('standardSkuDescription', 'For general purpose production app'),
+                detail: localize('standardSkuDetail', 'Free SSL, 5 Custom Domains, Custom Authentication, SLA')
             }
         ];
     }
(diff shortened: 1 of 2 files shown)


src/constants.ts
```diff
@@ -26,6 +26,13 @@ export const onlyGitHubSupported: string = localize('onlyGitHubSupported', 'Only
 export const githubAuthProviderId: string = 'github';
 export const githubScopes: string[] = ['repo', 'workflow', 'admin:public_key'];
 
+export const freeSkuDetail: string = localize('freeSkuDetail', 'Free SSL, 2 Custom Domains');
```
nturinski, line 29: Typically we only export strings to constants when they are used in multiple places.

src/constants.ts
```diff
@@ -26,6 +26,13 @@ export const onlyGitHubSupported: string = localize('onlyGitHubSupported', 'Only
...
+export const standardSkuDetail: string = localize('standardSkuDetail', 'Free SSL, 5 Custom Domains, Custom Authentication, SLA');
```
nturinski, line 30: Looking at the https://azure.microsoft.com/en-us/pricing/details/app-service/static/, I think we should include the $9 per month since that's the only thing that actually describes the pricing.

I'm not sure if we need the Free SSL in the details since that applies to both.

alexweininger: > I think we should include the $9 per month

I thought about including the $9/mo pricing, but was worried about the asterisks, since I'm not sure if we are required to display them. And I don't know where to display them. Or are these not an issue?

image

image

> I'm not sure if we need the Free SSL in the details since that applies to both.

I was using this as an example, and it shows Free SSL on each item.

image

ejizba: IMO the only thing worse than no pricing is inaccurate pricing. If you can find an API to dynamically get the pricing that's great, but otherwise I don't think we should hard-code a price that is subject to change. Unfortunately we've never been able to find that API for other extensions, but we have added a link to the pricing page at the bottom of the quick pick

alexweininger: Azure does have a [Retail prices REST API](https://docs.microsoft.com/en-us/rest/api/cost-management/retail-prices/azure-retail-prices#api-property-details). Seems like we can get pretty specific pricing information, however kinda feels overkill when we could just add link to the pricing page.

src/commands/createStaticWebApp/SkuListStep.ts
```diff
@@ -14,28 +14,35 @@ export class SkuListStep extends AzureWizardPromptStep<IStaticWebAppWizardContex
...
                 tier: 'Free',
+                description: localize('standardSkuDetail', 'Free SSL, 5 Custom Domains, Custom Authentication, SLA'),
+                detail: localize('freeSkuDetail', 'Free SSL, 2 Custom Domains')
             },
             {
                 name: 'Standard',
                 tier: 'Standard',
-            }];
+                description: localize('standardSkuDescription', 'For general purpose production app'),
+                detail: localize('freeSkuDescription', 'For hobbies/personal projects')
```
nturinski, line 44: I think you've put these in the wrong spots when you moved them

nturinski (approved):

Issue #356, opened 2021-04-16
[GitHub Codespaces] Fail to create a static web app with an error "Unable to fork respository" when publishing local code to a new GitHub repository 


v-xinda:
**OS:** Win 10
**Build Version:** 20210415.7
**Environment:** GitHub Codespaces (local)

**Repro Steps:**
1. F1 --> Click "Create Static Web App...(Advanced)" --> Publish local code to a new GitHub repository --> Choose organization 
--> Select a project --> Enter the name of the new GitHub repository --> Select "public" --> Enter a unique shortname for the new remote --> Enter a name for the new static web app --> create a new resource group --> Choose "Custom" --> Enter "app" --> Enter "api" --> Enter "build" --> Select a region.
2. Pop up a dialog to create a fork as below --> Click "Create Fork" or "NO".
![image](https://user-images.githubusercontent.com/57662131/114986722-eead9a80-9ec6-11eb-9f6f-7799f16b15cf.png)
3. Check whether the static web app is created successfully or not.

**Expect:**
The static web app is created successfully.

**Actual:**
Fail to create a static web app with an error "Unable to fork respository".
![image](https://user-images.githubusercontent.com/57662131/114986704-e6edf600-9ec6-11eb-8388-34dc142b9e5a.png)

nturinski:
@v-xinda The error is actually being thrown by the Git extension.  You can see the error message here: https://github.com/microsoft/vscode/blob/874d19930e3200b4ceb23520f59c99d653d4dc3d/extensions/github/src/pushErrorHandler.ts#L23

We're using their push command, so that's why we're seeing their error handling.  I'm not really sure why the fork would be failing, but we're seeing this because you don't have access to the GitHub repo that you're using.  This is strange because we check for admin access.  Do you know how you're getting to this state?

v-xinda:
@nturinski Hi Nathan, I don't know how to get the state. Can you give me some guides to get the state.

nturinski:
@v-xinda I honestly have no idea how this would happen because we check the project to make sure that the user has admin access to the GitHub project.  If you happen to run into this again, could you document the exact steps you did to make it appear?  This is a very interesting bug 🤔

v-xinda:
@nturinski Verified this issue on the latest build 20210422.7. 
Fail to create an advanced/basic static web app with an error "Git: Failed to execute git".
[...]

PR #470, opened 2021-08-11
Add "Create from template" option

alexweininger:
* We plan on adding descriptions to the repos on GitHub and then displaying these in the Quickpick. Once these descriptions are added, I'll make a PR for that.
* The angular-basic template no longer has working defaults. This is because with this flow, we cannot pick up the default values from the settings.json in the template since it's not cloned locally.
* @craigshoemaker is a maintainer of the https://github.com/staticwebdev account. I've met with him and they will make sure to let us test any changes/additions they want to make to the templates.
* This is still just an idea, but there might be templates in the future that we don't want to include in our extension. And I think we could use [GitHub repo topics](https://docs.github.com/en/github/administering-a-repository/managing-repository-settings/classifying-your-repository-with-topics) to mark whether to include or exclude a repo from our Quickpick.


src/utils/gitHubUtils.ts
```diff
@@ -105,3 +105,33 @@ export async function createFork(context: IActionContext, remoteRepo: ReposGetRe
...
+        return client.rest.repos.createUsingTemplate({
+            template_owner: nonNullProp(templateRepo, 'owner').login,
+            template_repo: templateRepo.name,
+            name: newRepoName ?? templateRepo.name,
+            private: isPrivate ?? false,
+            description: description ?? undefined
+        });
+    } catch (e) {
+        const generateUrl: string = `${templateRepo.html_url}/generate`;
+        throw new Error(localize('createFromTemplateFail', 'Could not create repository from template. [Click here to create it manually]({0}).', generateUrl));
```
neelip, line 125: Maybe this can be "Create it manually" hyperlink instead

src/utils/workspaceUtils.ts
```diff
@@ -66,15 +67,22 @@ export async function tryGetWorkspaceFolder(context: IActionContext): Promise<Wo
...
+    const result = await context.ui.showWarningMessage(message, { modal: true, stepName: noWorkspaceWarning }, openExistingProjectMsg, cloneProjectMsg, createFromTemplateMsg);
     if (result === cloneProjectMsg) {
         await cloneRepo(context, '');
-        context.telemetry.properties.noWorkspaceResult = 'cloneProject';
+        context.telemetry.properties.noWorkspaceResult = cloneProject;
     } else if (result === openExistingProjectMsg) {
         await openFolder(context)
         context.telemetry.properties.noWorkspaceResult = openExistingProject;
+    } else if (result === createFromTemplateMsg) {
+        await createStaticWebAppFromTemplate(context);
```
nturinski, line 84: Downside of calling `createStaticWebAppFromTemplate` here like this is the verifying workspace notification doesn't go away during the whole create process (if your entry point is the +)

I think just set the `context.isTemplate` to true here and just return undefined and go through the rest of the create process, if we refactor `createStaticWebApp` a bit.  I've left a comment about that.

src/commands/createStaticWebApp/createStaticWebAppFromTemplate.ts
```diff
@@ -0,0 +1,23 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+import { IActionContext, ICreateChildImplContext } from 'vscode-azureextensionui';
+import { ext } from '../../extensionVariables';
+import { StaticWebAppTreeItem } from '../../tree/StaticWebAppTreeItem';
+import { SubscriptionTreeItem } from '../../tree/SubscriptionTreeItem';
+import { showSwaCreated } from '../showSwaCreated';
+import { IStaticWebAppWizardContext } from './IStaticWebAppWizardContext';
+
+export async function createStaticWebAppFromTemplate(context: IActionContext & Partial<ICreateChildImplContext> & Partial<IStaticWebAppWizardContext>, node?: SubscriptionTreeItem): Promise<StaticWebAppTreeItem> {
```
nturinski, line 13: I think we could refactor this and add it to `createStaticWebApp.ts`.  Kind of similar to how `createStaticWebAppAdvanced`, we could call `createStaticWebApp` and pass in the options to set `context.fromTemplate = true`.

To look cleaner, we could refactor this code:
```
const progressOptions: ProgressOptions = {
        location: ProgressLocation.Notification,
        title: localize('verifyingWorkspace', 'Verifying workspace...')
    };
    await window.withProgress(progressOptions, async () => {
        const folder = await tryGetWorkspaceFolder(context);
        if (folder) {
            await setWorkspaceContexts(context, folder);
        } else {
            await showNoWorkspacePrompt(context);
        }
    });
```
into its own function, and if we see context.fromTemplate === true, we just skip it and go straight down to `node.createChild`.

Another benefit to doing this is we don't have to duplicate the create process at all.  I'm not sure if it was omitted intentionally, but this function is missing the postCreate actions (such as verifying the actions built and opening the node up for the user).  So anytime we wanted to change that behavior, we would have to do it in two places.

src/constants.ts
```diff
@@ -31,6 +33,8 @@ export const isEndGroup = (t: string): boolean => /##\[endgroup\]/.test(t)
 export const githubAuthProviderId: string = 'github';
 export const githubScopes: string[] = ['repo', 'workflow', 'admin:public_key'];
 
+export const templateReposUsername: string = 'staticwebdev';
```
nturinski, line 36: I think this is only used in the one spot, so we probably don't need to throw this in the constants file.  We don't want to unnecessarily clutter this file up.

src/utils/gitHubUtils.ts
```diff
@@ -105,3 +105,33 @@ export async function createFork(context: IActionContext, remoteRepo: ReposGetRe
         throw new Error(localize('forkFail', 'Could not automatically fork repository. Please fork [{0}]({1}) manually.', remoteRepo.name, remoteRepo.html_url));
     }
 }
+
+export async function createRepoFromTemplate(context: IActionContext, templateRepo: RepoData, newRepoName?: string, isPrivate?: boolean, description?: string): Promise<ReposCreateUsingTemplateResponse> {
```
nturinski, line 109: I think we can just include this in the `RepoCreateFromTemplateStep`.  Assuming we refactor it how I said in the comment below, it won't have to be called in two separate places.

src/utils/gitHubUtils.ts
```diff
@@ -105,3 +105,33 @@ export async function createFork(context: IActionContext, remoteRepo: ReposGetRe
...
+        return client.rest.repos.createUsingTemplate({
+            template_owner: nonNullProp(templateRepo, 'owner').login,
+            template_repo: templateRepo.name,
+            name: newRepoName ?? templateRepo.name,
+            private: isPrivate ?? false,
+            description: description ?? undefined
+        });
+    } catch (e) {
+        const generateUrl: string = `${templateRepo.html_url}/generate`;
+        throw new Error(localize('createFromTemplateFail', 'Could not create repository from template. [Create it manually]({0}).', generateUrl));
```
nturinski, line 125: Have you seen an error thrown from GitHub yet?  Usually GitHub throws pretty reasonable/understandable errors so I don't know if its necessary for us to catch and throw our own.  Plus, they may have more insight as to why it failed.

src/utils/gitHubUtils.ts
```diff
@@ -105,3 +105,33 @@ export async function createFork(context: IActionContext, remoteRepo: ReposGetRe
...
+    }
+}
+
+export async function getTemplateRepos(context: IActionContext): Promise<RepoData[]> {
```
nturinski, line 129: I think this can also go in the `TemplateListStep` file

package.nls.json
```diff
@@ -1,6 +1,7 @@
 {
     "staticWebApps.createStaticWebApp": "Create Static Web App...",
     "staticWebApps.createStaticWebAppAdvanced": "Create Static Web App... (Advanced)",
+    "staticWebApps.createStaticWebAppFromTemplate": "Create Static Web App from template...",
```
nturinski, line 4: nit: What do you think about `Create Static Web App... (Template)` or `Create Static Web App from Template...` Somehow the lower case t really bugs me aesthetically lol

src/commands/createRepo/RepoCreateFromTemplateStep.ts
```diff
@@ -0,0 +1,43 @@
...
+import { Progress } from 'vscode';
+import { AzureWizardExecuteStep } from "vscode-azureextensionui";
+import { ext } from '../../extensionVariables';
+import { RepoData } from '../../gitHubTypings';
+import { createRepoFromTemplate } from "../../utils/gitHubUtils";
+import { localize } from '../../utils/localize';
+import { nonNullProp } from '../../utils/nonNull';
+import { IStaticWebAppWizardContext } from '../createStaticWebApp/IStaticWebAppWizardContext';
+
+export class RepoCreateFromTemplateStep extends AzureWizardExecuteStep<IStaticWebAppWizardContext> {
```
nturinski, line 15: I'm a little torn here.  On one hand, this step has a few things different happening when compared to RepoCreateStep, but on the otherhand, a lot of it including the set-up and strings are identical.  

I think I'm okay with either solution, but lean a little bit more towards having one step that has a createRepo and createRepoFromTemplate functions that we can call inside the function.

src/commands/createStaticWebApp/TemplateListStep.ts
```diff
@@ -0,0 +1,33 @@
...
+
+import { AzureWizardPromptStep, IAzureQuickPickItem } from 'vscode-azureextensionui';
+import { RepoData } from '../../gitHubTypings';
+import { getTemplateRepos } from '../../utils/gitHubUtils';
+import { localize } from '../../utils/localize';
+import { nonNullProp } from '../../utils/nonNull';
+import { IStaticWebAppWizardContext } from './IStaticWebAppWizardContext';
+
+export class TemplateListStep extends AzureWizardPromptStep<IStaticWebAppWizardContext> {
+    public async prompt(context: IStaticWebAppWizardContext): Promise<void> {
```
nturinski, line 14: To make it cache template entries, you can do something like this:

```
private _templateRepos: RepoData[] | undefined;

        const placeHolder: string = localize('chooseTemplatePrompt', 'Choose a template for the new static web app.');

        const getPicks = async () => {
            this._templateRepos ||= await getTemplateRepos(context);
            return this._templateRepos.map((repo: RepoData) => ({ label: repo.name, data: repo }));
        }
```

That way, you wont' have to add more things to the context.

package.json
```diff
@@ -58,6 +59,12 @@
     "main": "./main.js",
     "contributes": {
         "commands": [
+            {
+                "command": "staticWebApps.createStaticWebAppFromTemplate",
+                "title": "%staticWebApps.createStaticWebAppFromTemplate%",
+                "category": "Azure Static Web Apps",
+                "icon": "$(add)"
```
nturinski, line 66: Does this icon show up anywhere?  I think this is only for the VS Code ribbon.

src/utils/gitHubUtils.ts
alexweininger: I haven't seen an error. I'll remove that. 

I was sorta following how we did it for `createFork`. Was that for a specific error we were seeing?

https://github.com/microsoft/vscode-azurestaticwebapps/blob/0fb1c6358fea1897be94c246f86e9ceff18961c5/src/utils/gitHubUtils.ts#L105-L106

package.nls.json
alexweininger: Changed to `Create Static Web App from Template...`.

src/utils/gitHubUtils.ts
nturinski: Yeah, we were running into some issues where repos couldn't be forked.  I don't remember the exact details as to how it was happening, but it was tracked in this issue: https://github.com/microsoft/vscode-azurestaticwebapps/issues/356

package.nls.json
```diff
@@ -1,6 +1,7 @@
 {
     "staticWebApps.createStaticWebApp": "Create Static Web App...",
     "staticWebApps.createStaticWebAppAdvanced": "Create Static Web App... (Advanced)",
+    "staticWebApps.createStaticWebAppFromTemplate": "Create Static Web App from Template...",
```
alexweininger, line 4: We probably want to change the wording from "template" to "quickstart". But need help deciding.

* "Create Static Web App with a Quickstart"
* "Static Web App Quickstart"
* "Create with Quickstart Template"
* ???

Issue #460, opened 2021-07-27
Prevent "Verifying Workspace" from executing more than once


nturinski:
If a user tries to create a SWA, we immediately verify their workspace.  However, in the case of somebody hitting it multiple times before it is done verifying, the verification process happens multiple times (it's a bit of a slow process, so this seems likely to happen to users).

The solution is probably to prevent verify from happening while another verification is on-going.

PR #472, opened 2021-08-12
Prevent workspace verification if user is currently creating static app

nturinski:
Fixes https://github.com/microsoft/vscode-azurestaticwebapps/issues/460

I decided to go with the event because I thought it was kind of weird to let users keep verifying while they were going through the prompts.  Also, if you clicked fast enough, you would see two "Verifying workspace..." prompts because the time between the verification ending and the prompt coming up was long enough to start another verification.

But if I waited until after the app was created, users wouldn't be able to create while the resources were being created, which I also thought was unpleasant.


src/commands/createStaticWebApp/createStaticWebApp.ts
```diff
@@ -17,7 +18,18 @@ import { IStaticWebAppWizardContext } from './IStaticWebAppWizardContext';
 import { postCreateStaticWebApp } from './postCreateStaticWebApp';
 import { setWorkspaceContexts } from './setWorkspaceContexts';
 
+export const isVerifyingWorkspaceEmitter: EventEmitter<boolean> = new EventEmitter<boolean>()
+let _isVerifyingWorkspace: boolean = false;
+
+isVerifyingWorkspaceEmitter.event(e => _isVerifyingWorkspace = e);
```
alexweininger, line 24: I like the use of EventEmitter here. Howevever, this returns a `Disposable`. Do we have to dispose it?

nturinski: I think we're okay here.  I don't see this in any of our other extensions that use `EventEmitters`.  I think all of this gets disposed when the extension closes and we wouldn't want to dispose it earlier than that anyway.

ejizba: > I don't see this in any of our other extensions that use EventEmitters

Uhh where? We should be disposing... 😬 

Sounds like you want to add it to `ext.context.subscriptions`

>     export interface ExtensionContext {
> 
>         /**
>          * An array to which disposables can be added. When this
>          * extension is deactivated the disposables will be disposed.
>          */
>         readonly subscriptions: { dispose(): any }[];

src/commands/createStaticWebApp/createStaticWebApp.ts
```diff
@@ -26,25 +38,30 @@ export async function createStaticWebApp(context: IActionContext & Partial<ICrea
...
-    const environmentNode: EnvironmentTreeItem | undefined = <EnvironmentTreeItem | undefined>(await swaNode.loadAllChildren(context)).find(ti => {
-        return ti instanceof EnvironmentTreeItem && ti.label === productionEnvironmentName;
-    });
-    environmentNode && await ext.treeView.reveal(environmentNode, { expand: true });
-
-    void postCreateStaticWebApp(swaNode);
-    return swaNode;
+
+    try {
+        await window.withProgress(progressOptions, async () => {
```
ejizba, line 43: I have an idea - you know how quick picks are disabled with a "loading" indicator while loading the picks? I wonder if we can do a fake quick pick just to show progress during wizards. The main benefit is that the progress is shown in the top center where the rest of the wizard happens instead of at the bottom right (which I think is part of the confusion with "Verifying Workspace")

nturinski: Yeah, I like this idea a lot.

src/commands/createStaticWebApp/createStaticWebApp.ts
nturinski: https://github.com/microsoft/vscode-azuretools/blob/ba49322b5120c25f9202f0edd3383097afbab58d/ui/src/openReadOnlyContent.ts#L55

But okay, I was thinking about doing that as well, just didn't really want to expose the _verifyingWorkspace variable outside of the file, but I guess it doesn't really matter that much.

ejizba: Mmm yeah we're not great at disposing the `EventEmitter` itself (we should, tho). We are pretty good at disposing the `Event` which I think what Alex's comment was about

src/commands/createStaticWebApp/createStaticWebApp.ts
alexweininger: Yeah this idea would help us in a lot of places. We need a way to show loading while we do async work in between wizard steps.

For example, we do async work to load the initial value of an InputBox in several places. This results in a long period of time where no UI is shown. 

I have been experimenting with adding an option to pass in `Promise<string | undefined>` to `showInputBox`, however I'm having trouble making the experience nice due to some limits of the InputBox API.

I think the "fake quick pick" idea Eric mentioned is a great idea.

nturinski:
The workspace verification is actually pretty quick now because the initial delay from location is gone, and really you'll only see it if you have a modal come up (to open a workspace or commit your repo).  I was thinking about how busy it looks with the running notification and modal together and thought that this might look better:

image

Thoughts?

alexweininger:
> Thoughts? I was thinking maybe make the text a little more specific since the user may not know why their workspace is being verified.

It's really fast! Maybe we don't need to show anything at all?

https://user-images.githubusercontent.com/12476526/129769893-fe76e1e7-d72f-4d63-b8e5-ca7e68416f88.mov

nturinski:
> It's really fast! Maybe we don't need to show anything at all?

Discussed offline:  We could delay showing this for only verification that takes longer than half a second, but honestly, doesn't really feel like it's worth the effort for such a small UI element.

alexweininger (approved):
🚀

Issue #411, opened 2021-06-03
The "Production" node changes from "default Invalid" to "Production master (WaitingForDeployment)" after restarting VS Code


v-mengwe:
**OS:** Mac
**Build Version:** 20210602.8

**Repro Steps:**

1. Create a static web app with **`"app"`** when entering a location of application code -> Expand the static web app
    -> There is an error in the "Production" node.
   ![image](https://user-images.githubusercontent.com/62053665/120621296-57a99b80-c490-11eb-971a-2831c720492e.png)
2. Restart the VS Code -> Expand the static web app -> Check the "Production" node.
3. Check whether the "Production" node is unchanged.

**Expect:**
The "Production" node is unchanged.
![image](https://user-images.githubusercontent.com/62053665/120621473-7b6ce180-c490-11eb-9831-e1be6675bceb.png)

**Actual:**
The "Production" node changes from "`default Invalid`" to "`Production master (WaitingForDeployment)`".
![Screen Shot 2021-06-03 at 5 15 15 PM](https://user-images.githubusercontent.com/62053665/120621520-86c00d00-c490-11eb-8bde-a6bb8218ed11.png)

**More Info:**

1. The "Production" node is unchanged on Windows after restarting VS Code.
2. This issue also reproduces when adding double quotes to `api` and `build`.

**Suggestion:**
It would be better to add a validation when entering a location of application code/Azure Function code/build output relative.

PR #474, opened 2021-08-13
Validate user input on location steps as YAML

alexweininger:
This prevents the workflow from containing values like `""app""` which are YAML syntax errors.

This is related to #411, but I don't consider it a fix. A real fix is coming soon...


nturinski:
What do you think if we just validate the yaml settings upfront in the prompt step rather than fixing it retroactively?

alexweininger:
> What do you think if we just validate the yaml settings upfront in the prompt step rather than fixing it retroactively?

I like that approach much better. I was also looking for a package that could just validate a YAML string value since that would be the most robust option.

Would we just say any value with double quotes is invalid? Should we allow them to escape the quotes?

nturinski:
Actually, we already have this package installed: https://www.npmjs.com/package/yaml

It has a `stringify` method which we could use to convert the input as well.

alexweininger:
Sweet, using the YAML package I'm validating the input as YAML and it works really well.

image

nturinski (approved):

PR #476, opened 2021-08-13
Handle invalid workflow YAML files gracefully

### LINKED ISSUE #411: The "Production" node changes from "default Invalid" to "Production master (WaitingForDeployment)" after restarting VS Code

alexweininger:
Fixes #411 

Clicking on the invalid tree item opens and selects the invalid YAML block.

image

### DIFF
--- a/src/tree/GitHubConfigGroupTreeItem.ts
+++ b/src/tree/GitHubConfigGroupTreeItem.ts
@@ -5,8 +5,10 @@
 
 import { pathExists, readdir } from "fs-extra";
 import { basename, join } from "path";
-import { ThemeIcon } from "vscode";
-import { AzExtParentTreeItem, AzExtTreeItem, IActionContext, TreeItemIconPath } from "vscode-azureextensionui";
+import { Range, ThemeIcon } from "vscode";
+import { AzExtParentTreeItem, AzExtTreeItem, GenericTreeItem, IActionContext, parseError, TreeItemIconPath } from "vscode-azureextensionui";
+// eslint-disable-next-line import/no-internal-modules
+import { YAMLSyntaxError } from "yaml/util";
 import { localize } from "../utils/localize";
 import { parseYamlFile } from "../utils/yamlUtils";
 import { EnvironmentTreeItem } from "./EnvironmentTreeItem";
@@ -21,19 +23,44 @@ export type BuildConfigs = {
     'app_artifact_location'?: string
 }
 
+function getRangeFromError(error: YAMLSyntaxError): Range {
+    if (error.linePos) {
+        const { start, end } = error.linePos;
+        return new Range(start.line - 1, start.col - 1, end.line - 1, end.col - 1);
+    }
+    return new Range(0, 0, 0, 0);
+}
+
+function getYamlErrorMessage(error: unknown): string {
+    if (error instanceof YAMLSyntaxError) {
+        const range = getRangeFromError(error);
+        return `Error: Invalid YAML between lines ${range.start.line} and ${range.end.line}`;
+    } else {
+        return parseError(error).message;
+    }
+}
+
 export class GitHubConfigGroupTreeItem extends AzExtParentTreeItem {
     public static contextValue: string = 'azureStaticGitHubConfigGroup';
     public contextValue: string = GitHubConfigGroupTreeItem.contextValue;
-    public readonly label: string = localize('gitHubConfig', 'GitHub Configuration');
     public parent: EnvironmentTreeItem;
     public yamlFilePath: string;
     public buildConfigs: BuildConfigs | undefined;
 
-    public constructor(parent: EnvironmentTreeItem, yamlFilePath: string, buildConfigs: BuildConfigs) {
+    public constructor(parent: EnvironmentTreeItem, yamlFilePath: string, buildConfigs: BuildConfigs, public parseYamlError?: unknown) {
         super(parent);
         this.yamlFilePath = yamlFilePath;
         this.buildConfigs = buildConfigs;
         this.id = `${parent.id}-${this.yamlFilePath}`;
+        this.commandArgs = [this];
+    }
+
+    public get label(): string {
+        return this.parseYamlError ? localize('invalidGitHubConfig', 'Invalid GitHub Configuration') : localize('gitHubConfig', 'GitHub Configuration');
+    }
+
+    public get commandId(): string | undefined {
+        return this.parseYamlError ? 'staticWebApps.openYAMLConfigFile' : undefined;
     }
 
     public static async createGitHubConfigGroupTreeItems(context: IActionContext, parent: EnvironmentTreeItem): Promise<GitHubConfigGroupTreeItem[]> {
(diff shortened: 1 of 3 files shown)


src/tree/EnvironmentTreeItem.ts
```diff
@@ -187,6 +188,6 @@ export class EnvironmentTreeItem extends AzExtParentTreeItem implements IAzureRe
         const branch: string | undefined = remote ? await tryGetLocalBranch() : undefined;
         this.inWorkspace = this.parent.repositoryUrl === remote && this.branch === branch;
 
-        this.gitHubConfigGroupTreeItems = await GitHubConfigGroupTreeItem.createGitHubConfigGroupTreeItems(context, this);
+        [this.gitHubConfigGroupTreeItems, this.invalidGitHubConfigGroupTreeItems] = await GitHubConfigGroupTreeItem.createGitHubConfigGroupTreeItems(context, this);
```
alexweininger, line 191: I ended up using a tuple after trying to return InvalidTreeItems and the GithubConfigGroupTreeItems in an `AzExtTreeItem[]`. It would require more refactoring in this file and in `OpenYAMLConfigFile` and I think using a tuple was simpler.

nturinski: What if you had a boolean to indicate if the yaml is invalid?  

I think that could simplify the logic a little bit.  I think it would eliminate the need for the tuple.  You could just add some optional properties that you need for the range, I believe.

And use the boolean to determine which icon to use ofc :)

nturinski: Alternatively, I think if we edit this function a bit https://github.com/microsoft/vscode-azuretools/blob/344200a6fa49c426efd9debd61e86a75cbd2bce3/ui/src/tree/AzExtParentTreeItem.ts#L188 we could make it so you can specify the type of trees that are created on error rather than defaulting to a GenericTreeItem.

src/tree/GitHubConfigGroupTreeItem.ts
```diff
@@ -49,12 +53,18 @@ export class GitHubConfigGroupTreeItem extends AzExtParentTreeItem {
 
             for (const yamlFile of yamlFiles) {
                 const yamlFilePath: string = join(workflowsDir, yamlFile);
-                const buildConfigs: BuildConfigs | undefined = await parseYamlFile(context, yamlFilePath);
-                buildConfigs && treeItems.push(new GitHubConfigGroupTreeItem(parent, yamlFilePath, buildConfigs));
+                try {
+                    const buildConfigs: BuildConfigs | undefined = await parseYamlFile(context, yamlFilePath);
+                    buildConfigs && treeItems.push(new GitHubConfigGroupTreeItem(parent, yamlFilePath, buildConfigs));
+                } catch (e) {
+                    if (e instanceof YAMLSyntaxError) {
```
nturinski, line 60: We should still have tree items for when the error is not `YAMLSyntaxError`.  In those cases, you can just make `parseError(e).message` the label.

src/tree/GitHubConfigGroupTreeItem.ts
```diff
@@ -6,11 +6,14 @@
 import { pathExists, readdir } from "fs-extra";
 import { basename, join } from "path";
 import { ThemeIcon } from "vscode";
-import { AzExtParentTreeItem, AzExtTreeItem, IActionContext, TreeItemIconPath } from "vscode-azureextensionui";
+import { AzExtParentTreeItem, AzExtTreeItem, IActionContext, InvalidTreeItem, TreeItemIconPath } from "vscode-azureextensionui";
+// eslint-disable-next-line import/no-internal-modules
+import { YAMLSyntaxError } from "yaml/util";
```
nturinski, line 11: No other way to import? 😬

nturinski (commented):
Could you add some tests to the existing `selectBuildConfigs.test` file where the workflow files has invalid entries?

src/tree/GitHubConfigGroupTreeItem.ts
alexweininger: None that I'm aware of ☹️.  We added the `import/no-internal-modules` rule recently when we switched to the shared eslint config.

nturinski:
Yeah, that sounds good.  I was thinking that if you had the invalid workflow that it could check that the invalid range is as expected.

src/tree/GitHubConfigGroupTreeItem.ts
```diff
@@ -49,16 +76,20 @@ export class GitHubConfigGroupTreeItem extends AzExtParentTreeItem {
 
             for (const yamlFile of yamlFiles) {
                 const yamlFilePath: string = join(workflowsDir, yamlFile);
-                const buildConfigs: BuildConfigs | undefined = await parseYamlFile(context, yamlFilePath);
-                buildConfigs && treeItems.push(new GitHubConfigGroupTreeItem(parent, yamlFilePath, buildConfigs));
+                try {
```
nturinski, line 79: It's kind of a small distinction, but could you call `refreshImpl` here to initialize the tree item?  Right now we're doing two similar things in two places.

src/tree/GitHubConfigGroupTreeItem.ts
```diff
@@ -69,7 +100,18 @@ export class GitHubConfigGroupTreeItem extends AzExtParentTreeItem {
         return false;
     }
 
+    public get errorRange(): Range | undefined {
+        return this.parseYamlError instanceof YAMLSyntaxError ? getRangeFromError(this.parseYamlError) : undefined;
+    }
+
     public async loadMoreChildrenImpl(): Promise<AzExtTreeItem[]> {
+        if (this.errorRange) {
```
nturinski, line 108: You should use `this.parseYamlError` as the check since that's what you're using everywhere else.

src/tree/GitHubConfigGroupTreeItem.ts
```diff
@@ -21,19 +23,44 @@ export type BuildConfigs = {
...
     public yamlFilePath: string;
     public buildConfigs: BuildConfigs | undefined;
 
-    public constructor(parent: EnvironmentTreeItem, yamlFilePath: string, buildConfigs: BuildConfigs) {
+    public constructor(parent: EnvironmentTreeItem, yamlFilePath: string, buildConfigs: BuildConfigs, public parseYamlError?: unknown) {
         super(parent);
         this.yamlFilePath = yamlFilePath;
         this.buildConfigs = buildConfigs;
         this.id = `${parent.id}-${this.yamlFilePath}`;
+        this.commandArgs = [this];
```
nturinski, line 55: I don't think this is necessary to fill out
> /**
     * The arguments to pass in when executing `commandId`. If not specified, this tree item will be used.
     */

src/tree/GitHubConfigGroupTreeItem.ts
```diff
@@ -21,19 +23,44 @@ export type BuildConfigs = {
...
+    }
+
+    public get label(): string {
+        return this.parseYamlError ? localize('invalidGitHubConfig', 'Invalid GitHub Configuration') : localize('gitHubConfig', 'GitHub Configuration');
+    }
+
+    public get commandId(): string | undefined {
+        return this.parseYamlError ? 'staticWebApps.openYAMLConfigFile' : undefined;
```
nturinski, line 63: Shouldn't this command id exist on the GenericTreeItem that you create (similar to the GitHubConfigTreeItems?)

alexweininger: I put it here because I think more users will click the GitHubConfigGroupTreeItem than the GenericTreeItem child. The issue I had with putting the command on the error message was users might not know that clicking the error takes them to the YAML. What do you think?

nturinski: Yeah, I guess it's fine, it's just a little weird to me that it's inconsistent.  I would say maybe we should put it on both then?

nturinski (approved):

Issue #477, opened 2021-08-16
I get :Please enable git" eevn if git is enabled.


SagarSathe:

SagarSathe:
If you would like to use git features, please enable git in your settings. To learn more about how to use git and source control in VS Code read our docs.
![image](https://user-images.githubusercontent.com/78426910/129552965-162ad862-51c7-416a-afaf-f9053b7bf5fd.png)

nturinski:
There seems to be another error happening here, but we default to displaying `If you would like to use git features, please enable git in your settings. To learn more about how to use git and source control in VS Code read our docs.`.  There's currently no way for me to verify what this error is, but could you try to reinstall git and reload VS Code?  Maybe something happened to your git path?

PR #478, opened 2021-08-16
Throw original git Api error

nturinski:
Related to https://github.com/microsoft/vscode-azurestaticwebapps/issues/477

Originally when testing retrieving the Git API, the only error I encountered was an unhelpful "Unable to get Git model" whenever git wasn't enabled on VS Code, which is what led me to having custom error handling.  However, it seems like users have run into other errors, so I would like to verify git is disabled before throwing our custom error, otherwise throw the error from the API.

The error at the end shouldn't happen, but I assume that if the git API call doesn't error and returns undefined, then they may not have git.  In the above user's case, it's difficult to know if an error was thrown or if undefined was returned and we just fell through to the error.


src/getExtensionApi.ts
```diff
@@ -33,10 +34,15 @@ export async function getGitApi(): Promise<API> {
             return gitExtension.getAPI(1);
         }
```
alexweininger, line 35: nit: To make it more clear what the error on line 45 is for, I think adding an else statement and throwing the error in the else makes it easier to understand. Easier to understand that the error is only for when `gitExtension` is undefined.

```suggestion
        } else {
         	throw new Error(localize('unableGit', 'Unable to retrieve VS Code Git API. Please ensure git is properly installed and reload VS Code.'));
        }
```

nturinski: Haha, that did not work the way I wanted it to 😬

src/getExtensionApi.ts
```diff
@@ -31,12 +32,19 @@ export async function getGitApi(): Promise<API> {
...
+        if (!getWorkspaceSetting<boolean>('enabled', undefined, 'git')) {
+            // the getExtension error is very vague and unactionable so throw our own error if git isn't enabled
+            throw new Error(localize('gitEnabled', 'If you would like to use git features, please enable git in your [settings](command:workbench.action.openSettings?%5B%22git.enabled%22%5D). To learn more about how to use git and source control in VS Code [read our docs](https://aka.ms/vscode-scm).'));
+        } else {
+            throw err;
+        }
     }
 
-    throw new Error(localize('gitEnabled', 'If you would like to use git features, please enable git in your [settings](command:workbench.action.openSettings?%5B%22git.enabled%22%5D). To learn more about how to use git and source control in VS Code [read our docs](https://aka.ms/vscode-scm).'));
+    throw new Error(localize('unableGit', 'Unable to retrieve VS Code Git API.  Please ensure git is properly installed and reload VS Code.'));
```
alexweininger, line 47: This is unreachable now.

alexweininger (approved):

Issue #479, opened 2021-08-16
Wrong tree nodes show up with app settings commands if environment without a Functions API is selected


nturinski:
If a user uses the command palette to execute the upload local app settings command and selects an environment that doesn't an have an API, the tree node picker ends up getting passed an invalid node and starts to have weird behavior.  We can fix this by confirming the existence of a Functions API.

PR #481, opened 2021-08-16
Throw an error if trying to select a build with no API

nturinski:
Fixes https://github.com/microsoft/vscode-azurestaticwebapps/issues/479

I was trying to cook up a more elegant solution than this which involved either two different context values (with or without functions API) or filtering out any build environments that did not have APIs in the picker list., but it ended up being a bit of a headache for something that is not that commonly used.

The bug originated because the tree item that was being returned technically existed, but was not part of the children list because of the loadChildren logic.  I don't think there's anything wrong with the ui code, just something to be aware of in the future.


alexweininger (approved):
Should we disable/hide the app settings related commands if they don't have any SWAs with APIs?

On one hand this might cause more confusion, on the other hand the commands will never work when they don't have APIs right? So I feel like they shouldn't be enabled if we know they won't work.

nturinski:
Well, the commands are hidden in the context menus primarily because you need to right-click the App Settings node in order to access any of those commands.  This user specifically used the command palette to upload local settings.

image


That being said, in the case of command palette, I think I would either prefer a way to filter out invalid nodes with a warning at the bottom that some environments were filtered and why, or throw an error when selected. While it's annoying to do all that selection for it to just error, if it's just hidden, it's hard for the user to discover the command and the reason why it's not appearing.

PR #484, opened 2021-08-18
Change Angular output location default

alexweininger:
This change was suggested last week during standup.

The idea is that the current default `dist/` already breaks most Angular projects if users leave it unchanged. So we're hoping changing it to `dist/<Project name>` helps alert users to change it before deploying.

### DIFF
--- a/src/buildPresets/buildPresets.ts
+++ b/src/buildPresets/buildPresets.ts
@@ -12,7 +12,7 @@ export const buildPresets: IBuildPreset[] = [
         displayName: 'Angular',
         appLocation: '/',
         apiLocation: 'api',
-        outputLocation: 'dist/<Project name>'
+        outputLocation: 'dist/<project-name>'
     },
     {
         id: 'react',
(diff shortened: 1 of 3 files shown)


src/buildPresets/buildPresets.ts
```diff
@@ -12,7 +12,7 @@ export const buildPresets: IBuildPreset[] = [
         displayName: 'Angular',
         appLocation: '/',
         apiLocation: 'api',
-        outputLocation: 'dist'
+        outputLocation: 'dist/<Project name>'
```
nturinski, line 15: nit: Could we do:
```suggestion
        outputLocation: 'dist/<project-name>'
```

to match Angular's website: https://angular.io/guide/deployment
> The output folder is dist/project-name/ by default.

nturinski (approved):

src/buildPresets/buildPresets.ts
alexweininger: Good suggestion 👍

ejizba: Can we modify `validateInput` for the output step to basically check for this placeholder and make sure they change it? Also it would be nifty if just the placeholder is selected when you open the input box, but not sure if that's possible.

alexweininger: Did some research and we're only able to specify what part of the value is selected when using the native `window.showInputBox`, instead of creating the `InputBox` manually. 

I was able to implement the validate input idea :+1:

ejizba (approved):

Issue #429, opened 2021-06-30
Add Location step to the normal create scenario


gagarine:
Creating a new Azure Static Web App from VS doesn't seem to offer a way to set the region.

How the region is selected?

gagarine:
Also it seem Static Web App are only available on Central US, East US 2, West US 2, East Asia, and West Europe Azure regions. This is the only region available when I create the Static Web App from the Azure portal.

nturinski:
In order to select a region, you'll have to use the `Create a Static Web App... (Advanced)` command.  The last prompt will be location:

![image](https://user-images.githubusercontent.com/5290572/124308311-fa038000-db1d-11eb-8c4c-b60a6d440219.png)

gagarine:
Thanks. In my head, region is basic and not advance at all ;) .

I think this advance/not advance things is not super user friendly. Why not removing this difference and propose only one command with default (recommended) option?

Feel free to close the issue.

nturinski:
Hi @gagarine, thanks for the feedback!  We will strongly consider moving location back to the normal create experience.

Yeah, I agree that region is a pretty basic!  But we figured it was something we could automatically select for users without breaking them.  A big part of our design philosophy is to reduce the number of unnecessary prompts so that users can get their static apps up-and-running asap.

We offer the advanced/normal create scenarios because some users like having more flexibility when creating their apps such as selecting the resource group, but we believe that most users probably don't have a strong preference about this.

PR #485, opened 2021-08-18
Add location to basic create

nturinski:
Fixes https://github.com/microsoft/vscode-azurestaticwebapps/issues/429

I also moved location to be before all the build preset steps because it felt weird to me to do the Azure prompt (the SWA name) and then do all the prompts that also have location in their name, and _then_ do the physical location.

### DIFF
--- a/src/tree/SubscriptionTreeItem.ts
+++ b/src/tree/SubscriptionTreeItem.ts
@@ -58,6 +58,7 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
         const executeSteps: AzureWizardExecuteStep<IStaticWebAppWizardContext>[] = [];
 
         if (!context.advancedCreation) {
+            wizardContext.sku = SkuListStep.getSkus()[0];
             executeSteps.push(new ResourceGroupCreateStep());
         } else {
             promptSteps.push(new ResourceGroupListStep());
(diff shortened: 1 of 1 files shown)


src/tree/SubscriptionTreeItem.ts
```diff
@@ -85,17 +84,13 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
         ];
 
         const webProvider: string = 'Microsoft.Web';
-        if (context.advancedCreation) {
-            LocationListStep.setLocationSubset(wizardContext, new Promise((resolve) => {
-                resolve(locations);
-            }), webProvider);
+        LocationListStep.setLocationSubset(wizardContext, new Promise((resolve) => {
+            resolve(locations);
+        }), webProvider);
```
ejizba, line 89:
```suggestion
        LocationListStep.setLocationSubset(wizardContext, Promise.resolve(locations), webProvider);
```

src/tree/SubscriptionTreeItem.ts
```diff
@@ -58,6 +58,7 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
         const executeSteps: AzureWizardExecuteStep<IStaticWebAppWizardContext>[] = [];
 
         if (!context.advancedCreation) {
+            wizardContext.sku = SkuListStep.getSkus()[0];
```
ejizba, line 61: Fwiw, sku is another thing we decided was worth including in basic create for App Service.

ejizba (approved):

Issue #488, opened 2021-08-19
Fix and enable the "Edit configuration" command for virtual workspaces


alexweininger:
We assume the Uri schema is 'file' use `URI.fsPath` to open the GitHub Actions YAML file, and this doesn't work in virtual workspaces. See [Making your code ready for virtual workspaces](https://github.com/microsoft/vscode/wiki/Virtual-Workspaces#review-that-the-code-is-ready-for-virtual-resources) for more details.

PR #489, opened 2021-08-19
Declare limited virtual workspace support

alexweininger:
Background reading: [Extension Author Guide for Supporting Virtual Workspaces](https://github.com/microsoft/vscode/wiki/Virtual-Workspaces#review-that-the-code-is-ready-for-virtual-resources)

We'll declare "limited" support for virtual workspaces since some commands don't work.

Identified as not working, and are disabled in virtual workspaces:
* Create static web app
* Create static web app (advanced)
* Create HTTP function
* Edit configuration #488 

For edit configuration, I'm disabling it for now and filed #488 to fix and enable it.


package.json
```diff
@@ -422,7 +426,11 @@
                     }
                 }
             }
-        ]
+        ],
+        "virtualWorkspaces": {
```
neelip, line 430: The document shows this to be under "capabilities". Is that not required?

alexweininger: You're right. I had trouble getting it to show up in the extension explorer as "limited", that's probably why...  😅

package.json
```diff
@@ -62,7 +62,8 @@
                 "command": "staticWebApps.createStaticWebApp",
                 "title": "%staticWebApps.createStaticWebApp%",
                 "category": "Azure Static Web Apps",
-                "icon": "$(add)"
+                "icon": "$(add)",
+                "enablement": "!virtualWorkspace"
```
neelip, line 66: I think the "when" property needs to be updated as well. 
Here: 
```
"menus": {
            "view/title": [
                {
                    "command": "staticWebApps.createStaticWebApp",
                    "when": "view == staticWebApps",
                    "group": "navigation@1"
                },
```

alexweininger: Using the "enablement" makes it disabled everywhere, (commands, keybindings, menus). Here is what it looks like on the ribbon:

image

My thinking was it might be confusing to users if the buttons they are accustomed to seeing on the ribbon are gone. So keeping them there, and disabling them might be better.

Thoughts?

neelip: That's true, people might think it's a bug it they don't see it. I think it will be nice to have some notification pop-up for why it is disabled.

alexweininger: I was thinking that too. A notification or even a popup or tooltip that appeared when you hover over the disabled button. Let me see if that's possible.

alexweininger: Without creating a whole new command changing the tooltip isn't possible. And to display a notification we'd have to enable the button. 

I think this might be worth asking the VS Code team for guidance and see if they could add something to help out with this issue. I 100% think you're right in that users will probably not know why the buttons are disabled. 🤔

neelip: When does this description appear?
image

alexweininger: I created https://github.com/microsoft/vscode/issues/131205 asking for a feature to help us with this.

alexweininger: > When does this description appear?
image

When in a virtual workspace, it appears here in this hover, and it appears on the extension page. 

image

Issue #486, opened 2021-08-18
Detect Functions project and use that as api location if it exists


nturinski:
Since we default to folder name "api", this will totally break a user's SWA if they already have a functions project named anything other than "api" unless they know to change it in the workflow yaml.

alexweininger:
Suggested in https://github.com/microsoft/vscode-azurestaticwebapps/issues/465#issuecomment-892266091 as well

PR #491, opened 2021-08-20
Detect Functions projects and default to that value instead of "api"

nturinski:
Fixes https://github.com/microsoft/vscode-azurestaticwebapps/issues/486

In the case of advanced create, I thought if we detect them projects, we should still default to the one detected or use the quickpick rather than forcing users the manually input the location.  I imagine that the detection logic is pretty reliable, but should we add a "browse" and/or "skip" choice?

I was thinking about it, but my thought process was if they have a functions project in their workspace that they are deploying, then would they really want to skip the step?  And would we really not detect their functions project (that is required to be in the current workspace) that we would actually want to provide them with a "browse" option?

### DIFF
--- a/src/commands/createStaticWebApp/ApiLocationStep.ts
+++ b/src/commands/createStaticWebApp/ApiLocationStep.ts
@@ -16,17 +16,21 @@ export class ApiLocationStep extends AzureWizardPromptStep<IStaticWebAppWizardCo
         const defaultValue: string = context.buildPreset?.apiLocation ?? defaultApiLocation;
         const workspaceSetting: string | undefined = getWorkspaceSetting(apiSubpathSetting, context.fsPath);
 
-        context.apiLocation = (await context.ui.showInputBox({
-            value: workspaceSetting || defaultValue,
-            validateInput: (value) => validateLocationYaml(value, 'api_location'),
-            prompt: localize('enterApiLocation', "Enter the location of your Azure Functions code or leave blank to skip this step. For example, 'api' represents a folder called 'api'."),
-        })).trim();
+        context.apiLocation = context.detectedApiLocations ?
+            (await context.ui.showQuickPick(context.detectedApiLocations.map((apiPaths) => { return { label: apiPaths } }), { placeHolder: localize('selectApi', 'Select the location of your Azure Functions code') })).label :
+            (await context.ui.showInputBox({
+                value: workspaceSetting || defaultValue,
+                validateInput: (value) => validateLocationYaml(value, 'api_location'),
+                prompt: localize('enterApiLocation', "Enter the location of your Azure Functions code or leave blank to skip this step. For example, 'api' represents a folder called 'api'."),
+            })).trim();
 
         addLocationTelemetry(context, 'apiLocation', defaultValue, workspaceSetting);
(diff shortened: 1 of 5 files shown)


src/commands/createStaticWebApp/ApiLocationStep.ts
```diff
@@ -16,17 +16,21 @@ export class ApiLocationStep extends AzureWizardPromptStep<IStaticWebAppWizardCo
         const defaultValue: string = context.buildPreset?.apiLocation ?? defaultApiLocation;
         const workspaceSetting: string | undefined = getWorkspaceSetting(apiSubpathSetting, context.fsPath);
 
-        context.apiLocation = (await context.ui.showInputBox({
-            value: workspaceSetting || defaultValue,
-            validateInput: (value) => validateLocationYaml(value, 'api_location'),
-            prompt: localize('enterApiLocation', "Enter the location of your Azure Functions code or leave blank to skip this step. For example, 'api' represents a folder called 'api'."),
-        })).trim();
+        context.apiLocation = context.detectedApiLocations ?
+            (await context.ui.showQuickPick(context.detectedApiLocations.map((apiPaths) => { return { label: apiPaths } }), { placeHolder: localize('selectApi', 'Select the location of your Azure Functions code') })).label :
```
alexweininger, line 20:
```suggestion
            (await context.ui.showQuickPick(context.detectedApiLocations.map((apiPaths) => ({ label: apiPaths })), { placeHolder: localize('selectApi', 'Select the location of your Azure Functions code') })).label :
```

src/commands/createStaticWebApp/tryGetApiLocations.ts
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * Checks root folder and one level down first, then all levels of tree
+ * If a single function project is found, returns that path.
+ * If multiple projects are found, will prompt
+ * @param workspaceFolder Per the VS Code docs for `findFiles`: It is recommended to pass in a workspace folder if the pattern should match inside the workspace.
+ */
+export async function tryGetApiLocations(context: IActionContext, workspaceFolder: WorkspaceFolder | string): Promise<string[] | undefined> {
+    return await telemetryUtils.runWithDurationTelemetry(context, 'tryGetProject', async () => {
+        const folderPath = typeof workspaceFolder === 'string' ? workspaceFolder : workspaceFolder.uri.fsPath;
+        if (await fse.pathExists(folderPath)) {
```
alexweininger, line 23: I know the VS Code FileSystem doesn't have an equivalent `pathExists` function, however
if we're planning to support creating a SWA in virtual workspaces we could create our own implementation using the existing VS Code FileSystem api.

nturinski: We actually have this utils file in storage that supports pathExists using VS Code's file system. The plan was to move this into the ui package after we've had some time to test it, so it was my intention to move that now.  I suppose I could do that with this release then

alexweininger: Sounds good. We also just merged the wizard loading prompt feature in the ui package so we can get that in this release also.

Issue #501, opened 2021-08-30
Fail to open the workflow file when clicking the child node of GitHub Configuration


v-ruizh:
**OS:** Win 10
**Build Version:** 20210827.7
**Regression:** Previous Release

**Repro Steps:**
1. Expand one Static Web App -> Production node -> GitHub Configuration -> Click one child node of GitHub Configuration.
2. Check whether the workflow file opens without any error.

**Expect:**
The workflow file opens without any error.

**Actual:**
Fail to open the workflow file with an error.
![image](https://user-images.githubusercontent.com/45614674/131328254-1c2bf7d8-63e2-44ab-b2ca-244688385b0c.png)

**More Info:**
1. The workflow file exists in File Explorer.
2. This issue does not reproduce on Mac and Linux platforms.
3. This issue does not reproduce on the previous released extension v0.7.1.

nturinski:
Tracking external bug here: https://github.com/microsoft/vscode/issues/131925

But we should fix this for the release

PR #504, opened 2021-08-30
Use Uri.file rather than Uri.parse for creating Uri from yaml file path

nturinski:
Fixes https://github.com/microsoft/vscode-azurestaticwebapps/issues/501

### DIFF
--- a/src/commands/openYAMLConfigFile.ts
+++ b/src/commands/openYAMLConfigFile.ts
@@ -23,17 +23,17 @@ export async function openYAMLConfigFile(context: IActionContext, node?: StaticW
     let yamlFileUri: Uri | undefined;
 
     if (node instanceof GitHubConfigGroupTreeItem) {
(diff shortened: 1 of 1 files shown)


src/commands/openYAMLConfigFile.ts
```diff
@@ -47,7 +47,8 @@ export async function openYAMLConfigFile(context: IActionContext, node?: StaticW
         }
     }
 
-    const configDocument: TextDocument = await workspace.openTextDocument(yamlFileUri);
+    // temporariy workaround for https://github.com/microsoft/vscode/issues/131925
+    const configDocument: TextDocument = process.platform === 'win32' ? await workspace.openTextDocument(yamlFileUri.fsPath) : await workspace.openTextDocument(yamlFileUri);
```
ejizba, line 51: Can you try using `Uri.file` above instead of `Uri.parse`? I think that's the root cause of the problem.

alexweininger (approved):

PR #507, opened 2021-09-02
Bump'n'update 🦙🤛

nturinski:


wwlorey (approved):

wwlorey (approved):

src/tree/JobTreeItem.ts
```diff
@@ -80,6 +81,6 @@ export class JobTreeItem extends AzExtParentTreeItem implements IAzureResourceTr
     }
 
     public isAncestorOfImpl(_contextValue: string | RegExp): boolean {
-        return this.data.steps.length > 0;
+        return this.data.steps?.length !== 0;
```
alexweininger, line 84: This will return true if `this.data.steps` is undefined. Even though it's an edge case, I'm not sure we want that.

alexweininger (approved):

PR #513, opened 2021-09-14
Add  install/update and uninstall SWA CLI commands

alexweininger:
Note: This is prep work for local debugging features which will depend on the SWA CLI.

Add commands:
* Install or Update Azure Static Web Apps CLI
* Uninstall Azure Static Web Apps CLI

Also checks if the installed CLI version is out of date, and notifies the user to update similarly to Functions with func core tools. It will be important to help users get on the latest version of the SWA CLI because we will be depending on new features for local debugging.

Currently only handles global installations of the CLI.

### DIFF
--- a/src/commands/cli/getInstalledSwaCliVersion.ts
+++ b/src/commands/cli/getInstalledSwaCliVersion.ts
@@ -1,0 +1,20 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.txt in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import * as semver from 'semver';
+import { cpUtils } from "../../utils/cpUtils";
+
+export async function getInstalledSwaCliVersion(workingDirectory?: string): Promise<string | null> {
+    try {
+        // npx --global --no-install swa --version
+        // --global makes command only check for globally installed modules
+        // without --global, the command checks locally installed modules first, then globally installed modules
+        // --no-install makes command exit with non-zero exit code if the CLI is not found
+        const output: string = await cpUtils.executeCommand(undefined, workingDirectory, 'npx', '--global', '--no-install', 'swa', '--version');
+        return semver.clean(output);
+    } catch (e) {
+        return null;
+    }
+}
--- a/src/commands/cli/validateSwaCliIsLatest.ts
+++ b/src/commands/cli/validateSwaCliIsLatest.ts
@@ -1,0 +1,78 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as semver from 'semver';
+import * as vscode from 'vscode';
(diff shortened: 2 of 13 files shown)


src/commands/cli/validateSwaCliIsLatest.ts
```diff
@@ -0,0 +1,78 @@
...
+            if (semver.major(newestVersion) === semver.major(installedVersion) && semver.gt(newestVersion, installedVersion)) {
+                context.telemetry.properties.outOfDateFunc = 'true';
+                const message: string = localize(
+                    'outdatedSwaCli',
+                    'Update the Azure Static Web Apps CLI ({0}) to the latest ({1}) for the best experience.',
+                    installedVersion,
+                    newestVersion
+                );
+
+                const update: vscode.MessageItem = { title: 'Update' };
```
nturinski, line 52: I know this is copied and pasted, but we should probably localize it 😛

src/commands/cli/validateSwaCliIsLatest.ts
```diff
@@ -0,0 +1,78 @@
...
+                        await updateGlobalSetting(showSwaCliWarningKey, false);
+                    }
+                }
+                while (result === DialogResponses.learnMore);
+            }
+        }
+    });
+}
+
+async function hasNpm(): Promise<boolean> {
```
nturinski, line 71: I think that you should use this when trying to uninstall and install the package as well.  Similar to functions, if they don't have npm, we can throw an error to direct them how to manually install the CLI.

src/commands/cli/getInstalledSwaCliVersion.ts
```diff
@@ -0,0 +1,20 @@
...
+export async function getInstalledSwaCliVersion(workingDirectory?: string): Promise<string | null> {
+    try {
+        // npx --global --no-install swa --version
+        // --global makes command only check for globally installed modules
+        // without --global, the command checks locally installed modules first, then globally installed modules
+        // --no-install makes command exit with non-zero exit code if the CLI is not found
+        const output: string = await cpUtils.executeCommand(undefined, workingDirectory, 'npx', '--global', '--no-install', 'swa', '--version');
+        return semver.clean(output);
+    } catch (e) {
+        return null;
```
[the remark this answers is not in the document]
alexweininger: [`semver.clean` returns `string | null`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/semver/functions/clean.d.ts#L6).

We could change it to undefined if we wanted, but I was following how we did it here:
https://github.com/microsoft/vscode-azurefunctions/blob/27970afcd4de8fa39ce6ab554813828839b18ed1/src/funcCoreTools/getLocalFuncCoreToolsVersion.ts#L11

src/commands/cli/validateSwaCliIsLatest.ts
alexweininger: Hmm... I'll have to see how to manually install the CLI without npm. Not sure if that's possible yet. But I agree, should check for npm before attempting to run install/uninstall.

src/commands/cli/getInstalledSwaCliVersion.ts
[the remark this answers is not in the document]
nturinski: Oh, interesting.  Yeah, it really doesn't matter and I don't have a strong preference since that's how semver returns it

nturinski (approved):

PR #531, opened 2021-09-30
Add $swa-watch problem matcher

alexweininger:
This PR is the first of a few that will be laying the groundwork for the SWA debugging feature.

You can test it out using https://github.com/alexweininger/swa-cli-js-sample/tree/alex/swa-watch and running the "SWA: Run and Debug" debug configuration.

By providing a problem matcher [^1], users won't have to define their own (which is pretty tedious as it's large and complex). And it simplifies the task definition needed to use the SWA CLI for debugging. For reference, here is a [problem matcher from functions](https://github.com/microsoft/vscode-azurefunctions/blob/main/package.json#L762).

Here is an example task taken from https://github.com/anthonychu/swa-cli-js-sample/blob/main/.vscode/tasks.json#L23-L51 used to start the SWA CLI for debugging.

Before:
```json
{
    "type": "shell",
    "label": "swa: start",
    "command": "swa start http://localhost:3000/ --api http://localhost:7071",
    "isBackground": true,
    "problemMatcher": [
        {
            "pattern": [
                {
                    "regexp": ".",
                    "file": 1,
                    "location": 2,
                    "message": 3
                }
            ],
            "background": {
                "activeOnStart": true,
                "beginsPattern": ".",
                "endsPattern": "Azure Static Web Apps emulator started",
            }
        }
    ]
}
```

After: 
```json
{
    "type": "shell",
    "label": "swa: start",
    "command": "swa start http://localhost:3000/ --api http://localhost:7071",
    "problemMatcher": "$swa-watch",
    "isBackground": true
}
```

[^1]: [Problem matcher VS Code docs](https://code.visualstudio.com/docs/editor/tasks#_defining-a-problem-matcher)


neelip (approved):
I have tested debug with both setups (with and without #swa-watch reference) and didn't find any difference. 
I am good with this change, it can always be updated as needed later. 
Couple questions to think about: 
Is the problemMatcher expected to be exactly the way it is presented here for the debug to work or are customizations allowed? 
How would we alert/inform the user about this new feature?

alexweininger:
@neelip

Both good considerations.

> Is the problemMatcher expected to be exactly the way it is presented here for the debug to work or are customizations allowed? 

This problem matcher is intended to work for most cases, and most of all our own case when setting up debugging for the user. While I know there are some cases it won't work for, we can add additional problem matchers to cover those as they come up.

As for customizations, I don't think users can override or customize extension provided problem matchers. However, we could provide this problem matcher definition in our wiki and let users copy it and tweak it from there. This leads into my answer for the next question...

> How would we alert/inform the user about this new feature?

I hope that users won't have to "know" about this feature. What I mean by that is if the debugging feature we're working on now works properly, users shouldn't have to manually set this up and deal with the problem matcher. That said, I know users will run into issues, and want to use it how they see fit. So I think a page in the wiki explaining how it works and how to use it will be good.

I plan on writing plenty of documentation covering the debugging feature in the wiki when I have it all done, I will make sure to include a section going over this problem matcher.

PR #532, opened 2021-10-01
Add debug provider for SWA CLI

alexweininger:
Depends on https://github.com/microsoft/vscode-azurefunctions/pull/2979

The debug provider[^1] runs when a launch configuration that has a preLaunchTask of "swa: start" and does two checks:

1. Ensures the SWA CLI is installed
image

2. Ensures func core tools is installed (if SWA has an api)
image

It also terminates the SWA CLI task when the debug session is terminated.

How to test:

Need to build and install the functions vsix from the [alex/api](https://github.com/microsoft/vscode-azurefunctions/tree/alex/api) branch. And you can checkout this [sample repo](https://github.com/alexweininger/StartupAdventurer/tree/alex/debugProvider) to try out the debug provider.

[^1]: [Debug provider VS Code docs](https://code.visualstudio.com/api/references/vscode-api#DebugConfigurationProvider)
[^2]: [Problem matcher VS Code docs](https://code.visualstudio.com/docs/editor/tasks#_defining-a-problem-matcher)


src/debug/StaticWebAppDebugProvider.ts
```diff
@@ -0,0 +1,52 @@
...
+    public async provideDebugConfigurations(_folder?: WorkspaceFolder, _token?: CancellationToken): Promise<DebugConfiguration[]> {
+        // Don't provide any debug configurations automatically for now
+        return []
+    }
+
+    public async resolveDebugConfiguration(folder: WorkspaceFolder | undefined, debugConfiguration: DebugConfiguration, _token?: CancellationToken): Promise<DebugConfiguration | undefined> {
+        return await callWithTelemetryAndErrorHandling('resolveSwaDebugConfiguration', async (context: IActionContext) => {
+            if (debugConfiguration.preLaunchTask === swaTaskLabel) {
+                if (!folder) {
+                    return;
```
bwateratmsft, line 25:
```suggestion
                    return undefined;
```

Mentioned in review

src/debug/StaticWebAppDebugProvider.ts
```diff
@@ -0,0 +1,52 @@
...
+
+export class StaticWebAppDebugProvider implements DebugConfigurationProvider {
+    public async provideDebugConfigurations(_folder?: WorkspaceFolder, _token?: CancellationToken): Promise<DebugConfiguration[]> {
+        // Don't provide any debug configurations automatically for now
+        return []
+    }
+
+    public async resolveDebugConfiguration(folder: WorkspaceFolder | undefined, debugConfiguration: DebugConfiguration, _token?: CancellationToken): Promise<DebugConfiguration | undefined> {
+        return await callWithTelemetryAndErrorHandling('resolveSwaDebugConfiguration', async (context: IActionContext) => {
+            if (debugConfiguration.preLaunchTask === swaTaskLabel) {
```
bwateratmsft, line 23: Mentioned in review: could make a helper API that takes a task name as label, and returns true/false if it's SWA. For implementation, can use VSCode tasks API to query for the command, if it starts with `'swa '` it's probably a SWA task.

src/debug/StaticWebAppDebugProvider.ts
```diff
@@ -0,0 +1,52 @@
...
+                    }
+                }
+            }
+            return debugConfiguration;
+        });
+    }
+}
+
+export function terminateSwaTasks(): void {
+    tasks.taskExecutions.filter((execution) => execution.task.name === swaTaskLabel).forEach(te => te.terminate());
```
bwateratmsft, line 51: Mentioned in review: can use the aforementioned API to query

src/commands/cli/swaCliTask.ts
```diff
@@ -0,0 +1,20 @@
...
+import { swaTaskLabel } from '../../constants';
+import { terminateSwaTasks } from '../../debug/StaticWebAppDebugProvider';
+
+export function registerSwaCliTaskEvents(): void {
+    registerEvent('azureStaticWebApps.onDidTerminateDebugSession', vscode.debug.onDidTerminateDebugSession, (context: IActionContext, debugSession: vscode.DebugSession) => {
+        context.errorHandling.suppressDisplay = true;
+        context.telemetry.suppressIfSuccessful = true;
+
+        if (debugSession.configuration.preLaunchTask === swaTaskLabel) {
+            terminateSwaTasks();
```
bwateratmsft, line 17: Mentioned in review: can use the below-mentioned API to check if it's a SWA task, and then terminate only the one task if so.

src/commands/cli/validateSwaCliInstalled.ts
```diff
@@ -0,0 +1,59 @@
...
+                installed = true;
+            } else if (input === DialogResponses.learnMore) {
+                await openUrl(installSwaCliUrl);
+            }
+        }
+    });
+
+    // validate that SWA CLI was installed only if user confirmed
+    if (input === install && !installed) {
+        if (await context.ui.showWarningMessage(localize('failedInstallSwaCli', 'The Azure Static Web Apps CLI installation has failed and will have to be installed manually.'), DialogResponses.learnMore) === DialogResponses.learnMore) {
```
bwateratmsft, line 53: `IAzureUserInput.showWarningMessage` takes a `learnMoreLink` parameter in the options, so you can just pass that `installSwaCliUrl` and don't need to check for the `DialogResponse` or open the URL yourself

src/debug/StaticWebAppDebugProvider.ts
```diff
@@ -0,0 +1,52 @@
...
+                }
+                const hasSwaCli = await validateSwaCliInstalled(context, localize('installSwaCli', 'You must have the Azure Static Web Apps CLI installed to debug your static web app.'));
+                if (!hasSwaCli) {
+                    return undefined;
+                }
+
+                const apiLocations = await tryGetApiLocations(context, folder);
+                const hasApi: boolean = !!apiLocations && apiLocations.length > 0;
+
+                if (hasApi) {
```
bwateratmsft, line 35: It's just a style thing so what you've written is perfectly valid, but another way to do it is this (which would eliminate the `hasApi` variable)
```suggestion
                if (apiLocations?.length) {
```

Basically, if `apiLocations` is `null` or `undefined`, then that will be the expression result--both are falsy, so the `if` won't execute. If `apiLocations` is defined but length is 0, then 0 will be the expression result--also falsy.

package.json
```diff
@@ -56,6 +56,7 @@
         "onCommand:staticWebApps.staticWebApps.action.rerun",
         "onCommand:staticWebApps.uninstallStaticWebAppsCli",
         "onCommand:staticWebApps.viewProperties",
+        "onDebug",
```
alexweininger, line 59: Added this to make sure our extension is activated when the user starts debugging. 

See https://code.visualstudio.com/api/references/activation-events#onDebug

bwateratmsft: I think you can do this instead:
```suggestion
        "onDebugResolve:pwa-chrome",
```

bwateratmsft (approved):

PR #557, opened 2021-10-28
Provide dynamic debug configs

alexweininger:
Leaving this as a draft because we are waiting for updates on microsoft/vscode#135899 before merging/releasing, but I'd still like to get eyes on it sooner rather than later.

In summary, this PR removes the work I did that manually scaffolded out tasks/launch.json files for debugging and replaces it with dynamic debug configs using a DebugConfigurationProvider [^1].

We are dynamically providing both tasks and debug configs. Each does two things:
1. Uses the NodeDetector on the root and each subfolder to look for static web apps, if found a task/config is created for that app using the detected framework defaults
2. If a swa-cli.config.json is present, load configs from there and append them to the detected configs

[^1]: [VS Code API - DebugConfigurationProvider](https://code.visualstudio.com/api/references/vscode-api#DebugConfigurationProvider)

Edit: microsoft/vscode#135899 was fixed and released in insiders, so the PR is ready to go 🎉


src/utils/detectorUtils.ts
```diff
@@ -0,0 +1,30 @@
...
+
+    const detector = new NodeDetector();
+    for (const subfolder of subfolders) {
+        const subResult = await detector.detect(subfolder);
+        if (subResult) {
+            results.push({ uri: subfolder, ...subResult });
+        }
+    }
+
+    return results
```
bwateratmsft, line 29: Nit:

```suggestion
    return results;
```

package.json
```diff
@@ -51,19 +51,27 @@
         "onCommand:staticWebApps.refresh",
         "onCommand:staticWebApps.reportIssue",
         "onCommand:staticWebApps.selectSubscriptions",
-        "onCommand:staticWebApps.setupRunningInVSCode",
         "onCommand:staticWebApps.showActions",
         "onCommand:staticWebApps.staticWebApps.action.cancel",
         "onCommand:staticWebApps.staticWebApps.action.rerun",
         "onCommand:staticWebApps.uninstallStaticWebAppsCli",
         "onCommand:staticWebApps.viewProperties",
+        "onDebug",
```
bwateratmsft, line 59: Would `onDebugResolve:swa` work here?

src/debug/StaticWebAppDebugProvider.ts
```diff
@@ -3,46 +3,128 @@
...
+            }
+
+            const appFolders = await detectAppFoldersInWorkspace(context, folder);
+            appFolders.forEach((appFolder) => {
+                const buildPreset = buildPresets.find((preset) => appFolder.frameworks.find((info) => info.framework === preset.displayName));
+
+                if (buildPreset) {
+                    result.push(this.createDebugConfiguration(path.basename(appFolder.uri.fsPath), path.relative(folder.uri.fsPath, appFolder.uri.fsPath)));
+                }
+            })
```
bwateratmsft, line 43: Nit:

```suggestion
            });
```

src/debug/StaticWebAppDebugProvider.ts
```diff
@@ -3,46 +3,128 @@
...
+
+            return result;
+        }) ?? [];
     }
 
-    public async resolveDebugConfiguration(folder: WorkspaceFolder | undefined, debugConfiguration: DebugConfiguration, _token?: CancellationToken): Promise<DebugConfiguration | undefined> {
-        return await callWithTelemetryAndErrorHandling('resolveSwaDebugConfiguration', async (context: IActionContext) => {
+    public async resolveDebugConfiguration(folder: WorkspaceFolder | undefined, debugConfiguration: DebugConfiguration, cancellationToken: CancellationToken): Promise<DebugConfiguration | undefined> {
+        return await callWithTelemetryAndErrorHandling('staticWebApps.resovlveDebugConfiguration', async (context: IActionContext) => {
```
bwateratmsft, line 50:
```suggestion
        return await callWithTelemetryAndErrorHandling('staticWebApps.resolveDebugConfiguration', async (context: IActionContext) => {
```

src/cli/SwaCliTaskProvider.ts
```diff
@@ -0,0 +1,105 @@
...
+
+    private createSwaCliTask(workspaceFolder: WorkspaceFolder, label: string, options: Pick<SWACLIOptions, 'context' | 'apiLocation' | 'run' | 'appLocation'>): Task {
+
+        const addArg = <T>(object: T, property: keyof T, name?: string, quote?: boolean): string => {
+            return object[property] ? quote ? `--${name ?? property.toString()}='${object[property]}'` : `--${name ?? property.toString()}=${object[property]}` : '';
+        };
+
+        const args: string[] = [addArg(options, 'appLocation', 'app-location'), addArg(options, 'apiLocation', 'api-location'), addArg(options, 'run', 'run', true)];
+        const command = `swa start ${options.context} ${args.join(' ')}`;
+        const task = new Task({
```
bwateratmsft, line 92: Indentation is a little odd here, maybe this?

```typescript
const task = new Task(
    {
        type: shell,
        command,
    },
    workspaceFolder,
    label,
    swa,
    new ShellExecution(
        {
            env: {
                BROWSER: 'none'
            }
        }
    ),
    swaWatchProblemMatcher
);
```

bwateratmsft (approved):

package.json
alexweininger: I removed `onDebug` and I don't think we need `onDebugResolve:swa` since we should only ever provide and therefore have to resolve debug configs with the type `pwa-chrome`.

The only reason I'm contributing our own debug type is so we can have our own menu item here:
image

Otherwise it would show up under a generic label like Chrome or Node.js I think.

bwateratmsft: But we have a `resolveDebugConfigurations` implemented, doesn't that mean we need to support `onDebugResolve`? And thus should have the activation event?

alexweininger: Follow up: What I tried to do, and what I still want to do if it's possible is make the `swa` debug type an "alias" for `pwa-chrome`. I ran into some issues and just ended up only using `pwa-chrome`.

alexweininger: > But we have a resolveDebugConfigurations implemented, doesn't that mean we need to support onDebugResolve? And thus should have the activation event?

We have `onDebugResolve:pwa-chrome`, is there a reason this isn't enough?

bwateratmsft: Ah! Now I understand what you mean. Yeah, you got it right, this should work.

nturinski: I know that this is already merged, but we do want to natively support edge/mozilla debugging as well?  We should be able to just see if they have VS Code debugger extensions installed or not.

alexweininger: I believe `pwa-edge` is built-in via https://github.com/microsoft/vscode-js-debug. So we'd need to provide a setting for users to control which they prefer, edge or chrome.

How important is this? Is it worth the added complexity to support the other browsers?

nturinski: Right now, I'd lean towards "no" though I do think it's safer to assume `pwa-edge` since it's built in, as you mentioned.

I guess I am a little concerned if we would get criticism for forcing users on a "Microsoft" path especially since the tasks are dynamic and they wouldn't be able to easily change it.

A check to see if they have the chrome debugger might be worth it though because it'll just fail if they don't have it installed, right?

alexweininger: I think it's actually built-in, the Chrome Debugger extension is deprecated in favor of the built-in ones.

See deprecated Chrome debugger: https://marketplace.visualstudio.com/items?itemName=msjsdiag.debugger-for-chrome

nturinski: Ok, disregard then

Issue #549, opened 2021-10-20
[bugbash] Functions extension isn't installed when moving to WSL


fiveisprime:
The core problem is that installing the extension in WSL doesn't bring Functions along. Not sure if this is something we can even do something to fix..

1. Install SWA
2. Go to the extension viewlet and install SWA to WSL (at this point only SWA is installed to WSL)
3. Run the initialization command

Result:

![image](https://user-images.githubusercontent.com/1186948/138180580-6754fcb4-a6ea-4865-9c00-2729aa322e80.png)

Expected:  to get the full debug config setup.

Additional details: when I realized what was going on, I switched back to the extension viewlet to install functions. When I install Functions, I get this notification:

![image](https://user-images.githubusercontent.com/1186948/138180744-e98c9eb5-86ea-4cf5-a28b-7d21b85cf1f8.png)

I know that I don't want to click that, but customers may be confused by this since the commands are similar (though the message is specific to Functions).

alexweininger:
@fiveisprime  Is the expected behavior that Functions should automatically be installed whenever a user installs SWA?

I don't think this is the current behavior on any platform.

PR #559, opened 2021-11-01
Ensure functions is installed when debugging API

alexweininger:
If they don't have it installed it'll prompt them to install it.

Fixes: #549

### DIFF
--- a/src/debug/StaticWebAppDebugProvider.ts
+++ b/src/debug/StaticWebAppDebugProvider.ts
@@ -65,7 +65,7 @@ export class StaticWebAppDebugProvider implements DebugConfigurationProvider {
 
                 if ((await tryGetApiLocations(context, folder))?.length) {
                     // make sure Functions extension is installed
-                    await getFunctionsApi(context);
+                    await getFunctionsApi(context, localize('funcInstallForDebugging', 'You must have the "Azure Functions" extension installed to debug a Functions API.'));
 
                     const configName = this.parseDebugConfigurationName(debugConfiguration);
                     const swaCliConfigFile = await tryGetStaticWebAppsCliConfig(folder.uri);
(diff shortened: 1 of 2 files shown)


src/debug/StaticWebAppDebugProvider.ts
```diff
@@ -63,6 +64,9 @@ export class StaticWebAppDebugProvider implements DebugConfigurationProvider {
                 }
 
                 if ((await tryGetApiLocations(context, folder))?.length) {
```
nturinski, line 66: Unrelated, but `tryGetApiLocations` can take a second or two (sometimes even double digits) to run.  If this is having every time the user tries to debug, I suggest that you look to see if they have an API path in the `swa.cli.config` rather than running `tryGetApiLocations`. 

The only case we should do that is if we were going to set the API path for them (but I assume that you only do that on the `swa.cli.config` init

src/debug/StaticWebAppDebugProvider.ts
```diff
@@ -63,6 +64,9 @@ export class StaticWebAppDebugProvider implements DebugConfigurationProvider {
...
+                    // make sure Functions extension is installed
+                    await getFunctionsApi(context);
```
nturinski, line 68: It might be better to have a custom error for this.  I personally think it could be confusing to debug and see this error `'You must have the "Azure Functions" extension installed to perform this operation.'`  

Isn't the func core tools cli the requirement here?  Does the user actually need the Functions Extension?

alexweininger: Yeah we should use a custom error message. And we need Functions because we are relying on the tasks that the Functions extension resolves.

```json
{
  "type": "func",
  "command": "host start",
  "problemMatcher": "$func-node-watch",
  "isBackground": true,
  "dependsOn": "npm build (functions)",
  "options": {
    "cwd": "${workspaceFolder}/api"
  }
}
```

src/debug/StaticWebAppDebugProvider.ts
alexweininger: Ok, that's a very reasonable concern. 

Right now I invoke it every time, but I think what I should do is check to see if they have a swa-cli.config.json and if it has the apiLocation property (just like you mentioned).

And I'll only invoke it if they don't have a swa-cli.config.json file at all.

Currently I'm always invoking it because I want to help users add the apiLocation property to their swa-cli.config.json file if they don't already have it set up, but I think it's ok to remove that feature for sake of performance.

nturinski (approved):

PR #564, opened 2021-11-03
Add postfix to debug configs

alexweininger:
Option A postfixing both:

image

Option B: postfixing only the ones picked up from the swa-cli.config.json file

image

I went with option B because A looked really busy but I'm open to feedback.

### DIFF
--- a/src/debug/StaticWebAppDebugProvider.ts
+++ b/src/debug/StaticWebAppDebugProvider.ts
@@ -31,7 +31,7 @@ export class StaticWebAppDebugProvider implements DebugConfigurationProvider {
 
             const swaCliConfigFile = await tryGetStaticWebAppsCliConfig(folder.uri);
             if (swaCliConfigFile) {
-                result.push(...Object.entries(swaCliConfigFile?.configurations ?? []).map(([name, options]) => this.createDebugConfiguration(name, options.appLocation)));
+                result.push(...Object.entries(swaCliConfigFile?.configurations ?? []).map(([name, options]) => this.createDebugConfiguration(name, options.appLocation, swaCliConfigFileName)));
             }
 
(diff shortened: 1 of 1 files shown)


bwateratmsft (approved):

fiveisprime:
Can they be grouped the way we do recommended SKUs?

image

Show the config from the file at the top and a group below with "detected automatically" with those. Is there a preference for what people should use? It would be helpful to make it obvious which one they _should_ pick with a recommendation.

alexweininger:
> Can they be grouped the way we do recommended SKUs?
> 
> image
> 
> Show the config from the file at the top and a group below with "detected automatically" with those. Is there a preference for what people should use? It would be helpful to make it obvious which one they _should_ pick with a recommendation.

This currently isn't possible but it's a great idea so I'll be filing a request to VS Code soon. 

As for which people should use, if they do have a swa-cli.config.json file setup for debugging I think using that would be recommended. If the swa-cli.config.json file is setup incorrectly for debugging we do have some basic checks

alexweininger:
Quickly exploring if we could just use our own quick pick for selecting debug configs. It works well, the only issue is this wouldn't work with the native debugging menu. But that's not very discoverable anyway so it might be ok.

I think we can explore more of this later on, and for now the built-in menu is fine. But we could get fancy rolling our own quick pick, use the learn more button, etc.

image

alexweininger:
> Is there a preference for what people should use? It would be helpful to make it obvious which one they should pick with a recommendation.

I thought about this for a while, and I don't think we can justify putting recommended by any of the picks here. It would get awkward if they have multiple apps, then which one do we say is recommended? One of them? All of them? I think for now showing where it comes from is sufficient until we can design a better experience.

PR #575, opened 2021-11-10
Add devserver-timeout flag to SWA CLI tasks

### LINKED ISSUE #574: There is an error when debugging the "angular-basic" project
**OS:** All
**Build Version:** 20211105.14
**Regression:** Not a regression

**Repro Steps:**

1. Clone https://github.com/staticwebdev/angular-basic -> Open the project in VS Code.
2. Open VS Code terminal and run "npm install" -> Go to the "Run and Debug" menu -> Select the "Show all automatic debug configurations." -> Select "Azure Static Web Apps…" -> Select "SWA: Run angular-basic".
3. Check whether there is no error when debugging.

**Expect:**
There is no error when debugging.

**Actual:**
There is an error when debugging.
MicrosoftTeams-image (8)

**More Info:**
The opened browser window as below:
MicrosoftTeams-image (7)

alexweininger:
Fixes #574

I'd like at least @wwlorey to verify that this works since he can repro #574 

I think setting the timeout to 90000 ms (90 seconds) is ok since it's 3x the default of 30000, but nothing else went into that choice.

### DIFF
--- a/src/cli/SwaCliTaskProvider.ts
+++ b/src/cli/SwaCliTaskProvider.ts
@@ -92,6 +92,9 @@ export class SwaTaskProvider implements TaskProvider {
         };
 
         const args: string[] = [addArg(options, 'appLocation', 'app-location'), addArg(options, 'apiLocation', 'api-location'), addArg(options, 'run', 'run', true)];
+
+        // Increase devserver timeout to 3x default. See https://github.com/microsoft/vscode-azurestaticwebapps/issues/574#issuecomment-965590774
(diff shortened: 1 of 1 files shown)


bwateratmsft (approved):

wwlorey (approved):
This worked locally for me 👍

nturinski (commented):
Could you make this a setting in the workspace with a default of 90000?

alexweininger:
> Could you make this a setting in the workspace with a default of 90000?

Genius, yup I'll do that.

PR #576, opened 2021-11-10
Create .gitattributes

alexweininger:
Marked `test/testProject/**/*` and `NOTICE.html` as linguist-generated

This will just change some visual things on GitHub

1. The files matched by the globs will be collapsed by default in PRs
2. The files won't be taken into account when generating this graphic  

image

Read more about this here:

https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github

https://github.com/github/linguist/blob/master/docs/overrides.md#generated-code

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+test/testProject/**/* linguist-generated=true
+NOTICE.html linguist-generated=true
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,2 @@
-test/testProject/**/* linguist-generated=true
+test/testProject/**/* linguist-vendored=true
 NOTICE.html linguist-generated=true


.gitattributes
```diff
@@ -0,0 +1,2 @@
+test/testProject/**/* linguist-generated=true
```
wwlorey, line 1: I *think* this should be `linguist-vendored` since it's not generated code, but just code that we didn't write. From https://github.com/github/linguist/blob/master/docs/overrides.md#vendored-code:
> Checking code you didn't write, such as JavaScript libraries, into your git repo is a common practice, but this often inflates your project's language stats and may even cause your project to be labeled as another language.

wwlorey (approved):

.gitattributes
alexweininger: Well it's up to whether we want these files to show up in diffs by default or not. I thought it'd be nice if these weren't in the diffs, what do you think?

bwateratmsft:
Wow, I never knew about this feature. I'm gonna do this for the Docker extension too!

.gitattributes
wwlorey: I'd assume that if we're updating the test projects it would be in a separate PR so those diffs wouldn't really clutter things up. But I could see them being annoying otherwise. It's up to you!

alexweininger:
image

nice 😍

PR #577, opened 2021-11-11
Prep for release - 0.9.0

alexweininger:
The NOTICE.html generated is way smaller than normal, I think it's due to what it says in the "FYI" of the image. Iirc that's new.

image

I ended up adding the gif because it won't actually go into the packaged VSIX.

> If you add a repository field to your package.json and it is a public GitHub repository, vsce will automatically detect it and adjust relative links accordingly, using the master branch by default. You can override the GitHub branch with the --githubBranch flag when running vsce package or vsce publish.

from https://code.visualstudio.com/api/working-with-extensions/publishing-extension#marketplace-integration


CHANGELOG.md
```diff
@@ -1,5 +1,24 @@
 # Change Log
 
+## 0.9.0 - 2021-11-11
+
+### Added
+- Added dynamic debug configurations for debugging a static web app using the Azure Static Web Apps CLI [#557](https://github.com/microsoft/vscode-azurestaticwebapps/pull/557)
+
+    ![Authorize Access to through GitHub](resources/changelog/dynamic_debug_configs.png)
+
+- Added autocomplete and validation for [swa-cli.config.json](https://github.com/Azure/static-web-apps-cli#swa-cliconfigjson-file) files
```
wwlorey, line 10: I think you can omit "Added" here and on line 6 above since it's already in the "Added" section

wwlorey (approved):
Linking a PR for each bullet point is a nice touch 👌

CHANGELOG.md
```diff
@@ -1,5 +1,24 @@
 # Change Log
 
+## 0.9.0 - 2021-11-11
+
+### Added
+- Dynamic debugging to easily debug your static web app using the [Azure Static Web Apps CLI](https://github.com/Azure/static-web-apps-cli) [#557](https://github.com/microsoft/vscode-azurestaticwebapps/pull/557)
+
+    ![Debug static web app with a dynamic configuration](resources/changelog/debugging.gif)
```
bwateratmsft, line 8: The .gif is very large (9.5 MB), and they don't compress well so it will significantly increase the VSIX size. In the extension page in VSCode, images in the readme get loaded directly from the web and so can be excluded from the VSIX via the `.vscodeignore` file, however I'm not sure if the same is true for the changelog.

I'd add this .gif (and the other .png image too) into the `.vscodeignore`, and check whether the image shows up when viewing the changelog from within the extension page. If it doesn't show up, I'd just remove it from the changelog, leaving it only in the readme where we know it will show up without needing to be in the VSIX.

bwateratmsft (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,24 @@
 # Change Log
 
+## 0.9.0 - 2021-11-11
+
+### Added
+- Dynamic debugging to easily debug your static web app using the [Azure Static Web Apps CLI](https://github.com/Azure/static-web-apps-cli) [#557](https://github.com/microsoft/vscode-azurestaticwebapps/pull/557)
+
+    ![Debug static web app with a dynamic configuration](resources/readme/debugging.gif)
```
bwateratmsft, line 8: Looks like `resources/readme` is also not yet in the `.vscodeignore`.

PR #584, opened 2021-11-15
Turn on "strict" tsconfig rule

alexweininger:
Had to make a few types-only changes, most of them are just informing TS that we set `parent` in the super constructor using the `!`.

### DIFF
--- a/src/tree/ActionTreeItem.ts
+++ b/src/tree/ActionTreeItem.ts
@@ -18,7 +18,7 @@ import { JobTreeItem } from './JobTreeItem';
 export class ActionTreeItem extends AzExtParentTreeItem implements IAzureResourceTreeItem {
     public static contextValueCompleted: string = 'azureStaticActionCompleted';
     public static contextValueInProgress: string = 'azureStaticActionInProgress';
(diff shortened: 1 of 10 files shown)


src/tree/ActionTreeItem.ts
```diff
@@ -18,7 +18,7 @@ import { JobTreeItem } from './JobTreeItem';
 export class ActionTreeItem extends AzExtParentTreeItem implements IAzureResourceTreeItem {
     public static contextValueCompleted: string = 'azureStaticActionCompleted';
     public static contextValueInProgress: string = 'azureStaticActionInProgress';
-    public parent: ActionsTreeItem;
+    public parent!: ActionsTreeItem;
```
bwateratmsft, line 21: image

bwateratmsft (approved):
LOUD NOISES

src/tree/ActionTreeItem.ts
bwateratmsft: (I know what the `!` does but it's always funny to imagine that we're literally just yelling the variable name louder to drown out TS' complaints)

PR #585, opened 2021-11-16
Fix error when checking if debug config is SWA

alexweininger:
Telemetry is showing that this method is being called with lots of `debugConfiguration`s that have no name property. Which is throwing an error when we do `debugConfiguration.name.startsWith`. I was tempted to suppress all telemetry until after that check, but I am afraid I might miss other errors if I did that.

### DIFF
--- a/src/debug/StaticWebAppDebugProvider.ts
+++ b/src/debug/StaticWebAppDebugProvider.ts
@@ -140,6 +140,6 @@ export class StaticWebAppDebugProvider implements DebugConfigurationProvider {
     }
 
     private isSwaDebugConfig(debugConfiguration: DebugConfiguration): boolean {
-        return debugConfiguration?.name?.startsWith(StaticWebAppDebugProvider.configPrefix);
+        return !!(debugConfiguration?.name?.startsWith(StaticWebAppDebugProvider.configPrefix));
     }
 }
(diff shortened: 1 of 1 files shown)


src/debug/StaticWebAppDebugProvider.ts
```diff
@@ -140,6 +140,6 @@ export class StaticWebAppDebugProvider implements DebugConfigurationProvider {
     }
 
     private isSwaDebugConfig(debugConfiguration: DebugConfiguration): boolean {
-        return debugConfiguration.name.startsWith(StaticWebAppDebugProvider.configPrefix);
+        return debugConfiguration?.name?.startsWith(StaticWebAppDebugProvider.configPrefix);
```
bwateratmsft, line 143: This works and is probably fine since it's a private method, but if the return result was `undefined` (i.e. if `debugConfiguration` or `name` were `undefined`), and the caller tried to do something other than a falsy check, it could error.

Doing a double-negation will ensure that the method always returns a `boolean`, never `null` or `undefined`.

```suggestion
        return !!(debugConfiguration?.name?.startsWith(StaticWebAppDebugProvider.configPrefix));
```

I'm not sure if the extra parentheses are actually needed but they won't hurt; TS will probably show an error if they are.

bwateratmsft (approved):

PR #589, opened 2021-12-01
Generate and upload SBOM

alexweininger:
Using the SBOM generator tool detailed here:
https://www.1eswiki.com/wiki/ADO_sbom_Generator#YAML_Pipelines

* I'm only uploading the sbom on linux, since I don't think we need to upload an sbom for each OS.
* I'm uploading the contents of the `_manifest` directory as an artifact named "sbom".

### DIFF
--- a/.azure-pipelines/common/lint.yml
+++ b/.azure-pipelines/common/lint.yml
@@ -8,8 +8,3 @@ steps:
   - task: ComponentGovernanceComponentDetection@0
     displayName: "Component Detection"
     condition: ne(variables['System.PullRequest.IsFork'], 'True')
-
-  - task: AzureArtifacts.manifest-generator-task.manifest-generator-task.ManifestGeneratorTask@0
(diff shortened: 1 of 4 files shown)


wwlorey (approved):
> I'm uploading the contents of the _manifest directory as an artifact named "sbom".

Was there guidance from the docs about renaming the _manifest directory to sbom? If not, I think it's worth sticking to the default behavior and naming the artifact "_manifest". It looks like an sbom is under the umbrella of a manifest:

> The ADO Manifest Generator task is used to generate an SBOM along with an associated bsi.json that captures internal-only provenance metadata about the build. ([source](https://www.1eswiki.com/wiki/ADO_sbom_Generator#Description))

bwateratmsft:
> * I'm only uploading the sbom on linux, since I don't think we need to upload an sbom for each OS.

I doubt we even need to run it on Windows/Mac...

bwateratmsft (approved):

PR #596, opened 2022-01-04
Delete resource group when deleting swa

### LINKED ISSUE #594: Deleting Static Web App from within VS Code leaves behind Azure Resource Groups
Transferring the issue from [Static Web Apps #676](https://github.com/Azure/static-web-apps/issues/676) by @SheldonHage 

**Deleting Static Web App from within VS Code leaves behind Azure Resource Groups**

**To Reproduce**
1. Use VS Code to checkin code to Github with action to deploy to new Azure Static Web App
2. After successful build and deployment...
3. Go to Azure Portal and see that a dedicated resource group was created for the static web app
4. In VS Code, use the Static Web App plugin to delete the Static Web App
5. Go back to Azure portal and see that the dedicated resource group remains.

**Expected behavior**
I think this should be handled differently. Basically what I saw after a number of trial and error failed deployments and deleting the static web apps from within VS Code, my Azure portal had many resource groups with auto-incremented names. In practice these would leave a person unsure which was still in use and which weren't

alexweininger:
Fixes #594 

Important to see this [comment](https://github.com/microsoft/vscode-azurestaticwebapps/pull/596/files#diff-df419a970c68bdec8ae31d33c884c0d08604f4273a8b5e3001a5fc32fe425a90R103) about some weird behavior I found when listing the resources in the RG before deleting.

### DIFF
--- a/src/tree/StaticWebAppTreeItem.ts
+++ b/src/tree/StaticWebAppTreeItem.ts
@@ -4,11 +4,12 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { WebSiteManagementClient, WebSiteManagementModels } from "@azure/arm-appservice";
+import { ResourceManagementModels } from '@azure/arm-resources';
 import { ProgressLocation, window } from "vscode";
 import { AzExtParentTreeItem, AzExtTreeItem, IActionContext, TreeItemIconPath } from "vscode-azureextensionui";
 import { onlyGitHubSupported, productionEnvironmentName } from '../constants';
 import { ext } from "../extensionVariables";
-import { createWebSiteClient } from "../utils/azureClients";
+import { createResourceClient, createWebSiteClient } from "../utils/azureClients";
 import { getResourceGroupFromId, pollAzureAsyncOperation } from "../utils/azureUtils";
 import { getRepoFullname } from '../utils/gitUtils';
 import { localize } from "../utils/localize";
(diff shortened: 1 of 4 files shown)


src/utils/azureClients.ts
```diff
@@ -4,11 +4,20 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { WebSiteManagementClient } from '@azure/arm-appservice';
-import { AzExtClientContext, createAzureClient } from 'vscode-azureextensionui';
+import { ResourceManagementClient } from '@azure/arm-resources';
```
bwateratmsft, line 7: I'd recommend `import type` to help keep this dev-time.

src/tree/StaticWebAppTreeItem.ts
```diff
@@ -4,11 +4,12 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { WebSiteManagementClient, WebSiteManagementModels } from "@azure/arm-appservice";
+import { ResourceManagementModels } from '@azure/arm-resources';
```
bwateratmsft, line 7: I'd recommend `import type` to help keep this dev-time. Could add for the import above this for `@azure/arm-appservice` too.

bwateratmsft (approved):

src/tree/StaticWebAppTreeItem.ts
```diff
@@ -96,6 +97,16 @@ export class StaticWebAppTreeItem extends AzExtParentTreeItem implements IAzureR
             const deleteResponse = await client.staticSites.deleteStaticSite(this.resourceGroup, this.name);
             await pollAzureAsyncOperation(context, deleteResponse, this.subscription);
 
+            const resourceClient = await createResourceClient([context, this]);
+            const resources: ResourceManagementModels.ResourceListResult = await resourceClient.resources.listByResourceGroup(this.resourceGroup);
+
+            // Recently deleted SWA is returned in the list of resources in the RG, even though it's been deleted.
+            // Only delete if 0 resources, or exactly 1 resource and if the resource is the SWA that has just been deleted.
+            if ((resources.length === 0 || (resources.length === 1 && resources[0].id === this.data.id)) && !resources.nextLink) {
```
nturinski, line 105: Rather than doing this check, we could have a background check running to see if `resources.length === 0` for a few minutes.  (Of course, if `resources.length > 1` we don't have to bother checking).  Definitely not a blocker, but this id check is a bit janky

alexweininger: What if that background check takes 2 to 3 or even 5 minutes? It might get confusing for the user if it doesn't get deleted soon(ish) after they delete their SWA.

nturinski: Well, typically users wouldn't even see that they have a rg to begin with since we don't explicitly list it anywhere.  We also don't mention that we are clearing any empty resource groups so I don't think a user would have that expectation.  

We could prompt the user if they want to delete the rg if the SWA is the only resource we detect during the delete as well?  We do that for app service plans (if we delete the last app in it) for cost reasons

I think we should still show that the SWA has deleted as soon as that async request reports that it is though.

alexweininger: Ok, sounds like we should have some more discussion on this before merging this PR. I'd get input (PM magic) from @fiveisprime on this too.

alexweininger: Ok I've simplified the logic to where I'm much happier with the solution.

Before deleting the SWA, I get a list of resources in the resource group. Then after deleting the SWA I delete the resource group if it had only a single resource in it (the SWA). Basically the same as before but the code is less janky.

PR #601, opened 2022-01-07
Handle reserved app settings prefixes

### LINKED ISSUE #509: Upload app settings fails when there are blocked settings
image

Static Web Apps does not allow [certain settings](https://docs.microsoft.com/en-us/azure/static-web-apps/apis#constraints).

There are settings that are included by default in function apps that cannot be uploaded. Some are required locally to run (like `FUNCTIONS_WORKER_RUNTIME`). However, looks like the settings API is all-or-nothing and fails if any one setting is restricted. Ideally it should upload the settings that are valid and leave the rest. Or perhaps provide an option to select which settings to upload. As it stands right now, I don't think most settings uploads are likely to fail.

alexweininger:
Depends on https://github.com/microsoft/vscode-azurefunctions/pull/3032
Fixes #509 

Excludes app settings that have reserved prefixes from being uploaded. 
https://docs.microsoft.com/en-us/azure/static-web-apps/apis#constraints

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -45,6 +45,9 @@ export const shell: string = 'shell';
 export const minSwaCliVersion = '0.8.0';
 export const swaCliConfigFileName = 'swa-cli.config.json';
 
+// https://docs.microsoft.com/en-us/azure/static-web-apps/apis#constraints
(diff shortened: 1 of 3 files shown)


src/constants.ts
```diff
@@ -45,6 +45,9 @@ export const shell: string = 'shell';
 export const minSwaCliVersion = '0.8.0';
 export const swaCliConfigFileName = 'swa-cli.config.json';
 
+// https://docs.microsoft.com/en-us/azure/static-web-apps/apis#constraints
+export const reservedSettingsPrefixes: (RegExp | string)[] = [/APPSETTING_/, /AZUREBLOBSTORAGE_/, /AZUREFILESSTORAGE_/, /AZURE_FUNCTION_/, /CONTAINER_/, /DIAGNOSTICS_/, /DOCKER_/, /FUNCTIONS_/, /IDENTITY_/, /MACHINEKEY_/, /MAINSITE_/, /MSDEPLOY_/, /SCMSITE_/, /SCM_/, /WEBSITES_/, /WEBSITE_/, /WEBSOCKET_/, /AzureWeb/];
```
nturinski, line 49: Since these are prefixes, you should probably include a ^ in the RegExps

bwateratmsft (approved):

PR #602, opened 2022-01-11
Bump required vscode version to 1.62.1

### LINKED ISSUE #579: [Suggestion] It would be better to update the VS Code version in extension/package.json file
**OS:** Mac
**Build Version:** 20211112.1
**Regression:** Not a regression

**Repro Steps:**
1. Install VS Code with 1.57.0 vetsion according to the version in extension/package.json file.
image
2. Debug a Reactsite project with "SWA: Run app" -> Fail to debug with an error.
3. Check whether the issue reproduces for higher version of VS Code.

**Actual:**
This issue doen't reproduce for 1.62.1 (or higher) version of VS Code.

**Suggestion:**
It would be better to update the VS Code version in extension/package.json file.

alexweininger:
The debugging work added in [`v0.9.0`](https://github.com/microsoft/vscode-azurestaticwebapps/releases/tag/v0.9.0) depends on [a fix](https://github.com/microsoft/vscode/issues/135899) that was released in `1.62.1`, now that VS Code is on `1.63.x` we should be good to bump this.

Fixes #579

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -51,7 +51,7 @@
                 "webpack-cli": "^4.6.0"
             },
             "engines": {
-                "vscode": "^1.57.0"
+                "vscode": "^1.62.1"
             }
         },
         "node_modules/@azure/abort-controller": {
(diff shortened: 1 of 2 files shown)


wwlorey (approved):

bwateratmsft (approved):

nturinski (commented):
Update the types too https://github.com/microsoft/vscode-azurestaticwebapps/blob/main/package.json#L515

PR #604, opened 2022-01-12
Add message for when no locations are found

### LINKED ISSUE #558: Provide message when there are no locations available
Suggested in https://github.com/microsoft/vscode-azurestaticwebapps/issues/536#issuecomment-954676354, it would be nice to have a message instead of just blank/nothing.

alexweininger:
Fixes #558 

Might need feedback on the wording, "No available regions."

### DIFF
--- a/src/tree/SubscriptionTreeItem.ts
+++ b/src/tree/SubscriptionTreeItem.ts
@@ -88,7 +88,8 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
 
         LocationListStep.setLocationSubset(wizardContext, Promise.resolve(locations), webProvider);
         LocationListStep.addStep(wizardContext, promptSteps, {
-            placeHolder: localize('selectLocation', 'Select a region for Azure Functions API and staging environments')
+            placeHolder: localize('selectLocation', 'Select a region for Azure Functions API and staging environments'),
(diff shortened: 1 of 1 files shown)


wwlorey (approved):

wwlorey:
> Might need feedback on the wording, "No available regions."

Sounds good to me 👍

bwateratmsft (approved):

src/tree/SubscriptionTreeItem.ts
```diff
@@ -88,7 +88,8 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
 
         LocationListStep.setLocationSubset(wizardContext, Promise.resolve(locations), webProvider);
         LocationListStep.addStep(wizardContext, promptSteps, {
-            placeHolder: localize('selectLocation', 'Select a region for Azure Functions API and staging environments')
+            placeHolder: localize('selectLocation', 'Select a region for Azure Functions API and staging environments'),
+            noPicksMessage: localize('noRegions', 'No available regions.')
```
nturinski, line 92: Do you think we could just have the `noPicksMessage` be the default behavior for `LocationListStep`?  It probably doesn't happen to frequently, but don't really see why we should only have this be limited to Static Web Apps.

PR #616, opened 2022-02-14
Migrate to azext-* packages

alexweininger:


src/extension.ts
```diff
@@ -28,6 +29,7 @@ export async function activateInternal(context: vscode.ExtensionContext, perfSta
     context.subscriptions.push(ext.outputChannel);
 
     registerAppServiceExtensionVariables(ext);
+    registerAzureUtilsExtensionVariables(ext);
```
bwateratmsft, line 32: Does the order of these three calls matter?

bwateratmsft (approved):

src/extension.ts
alexweininger: In this order, `registerAzureUtilsExtensionVariables` does nothing so we could remove it or swap the order. To be verbose I'll swap the order.

PR #627, opened 2022-03-07
Show friendly error for unsupported clouds

### LINKED ISSUE #624: There is an error in the subscription node after signing in with Azure China/Germany account
**OS:** All
**Build Version:** 20220228.4

**Repro Steps:**
1. Sign in with Azure China/Germany account.
2. Expand the subscription node in Azure view.
3. Check whether there is no error in the subscription node.

**Expect:**
There is no error in the subscription node.

**Actual:**
There is an error in the subscription node.
image

alexweininger:
Fixes #624 

image

### DIFF
--- a/src/tree/SubscriptionTreeItem.ts
+++ b/src/tree/SubscriptionTreeItem.ts
@@ -5,7 +5,7 @@
 
 import { StaticSiteARMResource, WebSiteManagementClient } from '@azure/arm-appservice';
 import { LocationListStep, ResourceGroupCreateStep, ResourceGroupListStep, SubscriptionTreeItemBase, uiUtils, VerifyProvidersStep } from '@microsoft/vscode-azext-azureutils';
-import { AzExtTreeItem, AzureWizard, AzureWizardExecuteStep, AzureWizardPromptStep, IActionContext, ICreateChildImplContext } from '@microsoft/vscode-azext-utils';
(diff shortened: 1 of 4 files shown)


src/tree/SubscriptionTreeItem.ts
```diff
@@ -40,7 +41,15 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
 
     public async loadMoreChildrenImpl(_clearCache: boolean, context: IActionContext): Promise<AzExtTreeItem[]> {
         const client: WebSiteManagementClient = await createWebSiteClient([context, this]);
-        const staticWebApps: StaticSiteARMResource[] = await uiUtils.listAllIterator(client.staticSites.list());
+        let staticWebApps: StaticSiteARMResource[] = []
+
+        try {
+            staticWebApps = await uiUtils.listAllIterator(client.staticSites.list());
+        } catch (e) {
+            if (isRestError(e) && e.code === 'InvalidResourceType') {
```
wwlorey, line 49: I'd recommend using `parseError` here. Then we don't need to create the `isRestError` function because checking if it's an `InvalidResourceType` is enough, right?

Also make sure we aren't relying on the error code casing. Something like: 

```ts
if (/InvalidResourceType/i.test(parseError(e).errorType)) {
```

alexweininger: Thanks, I didn't know the power of `parseError`  😄 🚀

wwlorey (approved):

PR #635, opened 2022-03-24
Bump version

alexweininger:


package.json
```diff
@@ -2,7 +2,7 @@
     "name": "vscode-azurestaticwebapps",
     "displayName": "Azure Static Web Apps",
     "description": "%staticWebApps.description%",
-    "version": "0.10.0",
+    "version": "0.10.1-alpha",
```
bwateratmsft, line 5: Whoa, how did `package-lock.json` end up with 12K line deletions?

alexweininger: I found that the lockfile version changed from 2 to 1. This has never happened before but I realized I created this PR in Codespaces and I assume there was an older version of npm installed on the machine.

bwateratmsft (approved):

PR #639, opened 2022-04-05
Upgrade simple-git to fix CVE-2022-24066

alexweininger:
Fixes CVE-2022-24066

### DIFF
--- a/package.json
+++ b/package.json
@@ -543,6 +543,7 @@
         "moment": "^2.27.0",
         "open": "^8.0.4",
         "semver": "^7.3.5",
+        "simple-git": "^3.5.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -543,6 +543,7 @@
         "moment": "^2.27.0",
         "open": "^8.0.4",
         "semver": "^7.3.5",
+        "simple-git": "^3.5.0",
```
bwateratmsft, line 546: Wouldn't the better approach be to update the `simple-git` dependency in `@microsoft/vscode-azext-azureappservice` and then update `@microsoft/vscode-azext-azureappservice` here?

PR #656, opened 2022-04-15
Use activity api for swa create and delete

alexweininger:
* Delete now uses a wizard

image


src/utils/activityUtils.ts
```diff
@@ -0,0 +1,10 @@
+import { ExecuteActivityContext } from "@microsoft/vscode-azext-utils";
```
nturinski, line 1: Add the license to the top

src/commands/deleteStaticWebApp/DeleteResourceGroupStep.ts
```diff
@@ -0,0 +1,27 @@
+/*---------------------------------------------------------------------------------------------
```
alexweininger, line 1: It might be good to add this to the azureutils package later

src/commands/createStaticWebApp/createStaticWebApp.ts
```diff
@@ -91,12 +93,17 @@ export async function createStaticWebApp(context: IActionContext & Partial<ICrea
...
+        accessToken: await getGitHubAccessToken(),
+        client,
+        ...context,
+        ...node.subscription,
+        ...(await createActivityContext())
+    };
 
     const title: string = localize('createStaticApp', 'Create Static Web App');
-    const promptSteps: AzureWizardPromptStep<IStaticWebAppWizardContext>[] = [];
+    const promptSteps: AzureWizardPromptStep<IStaticWebAppWizardContext & ExecuteActivityContext>[] = [];
```
nturinski, line 106: I don't think you need this cast btw

alexweininger: That's an [intersection type](https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html#intersection-types). 😄 But I think you're right, we don't need it.

PR #658, opened 2022-04-15
Add Next.JS build presets

### LINKED ISSUE #612: Include Next in the list of preset options
Include Next as an option during create static web apps step 3/5. 

Default values: 

app_location: '/'
api:''
output_location: ''

If users are deploying a static export next.js app, they need to be able to specify 'out' as an output folder https://nextjs.org/docs/advanced-features/static-html-export 

For users that are deploying dynamic Next.js apps (e.g. using SSR, API Routes, etc), they should leave api and output_location empty.

alexweininger:
Closes #612

As @simonaco presents in #612, Next.JS has two ways to create your app: 
1. the standard way, which includes server-side rendering (SSR). This bundles a custom express.js server into your app as an api
2. static export, which doesn't support SSR, and exports as, as the name implies, static html + js + css

We can add these both as different presets to make it easy for users, and I think this would be a good way to show them:

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "vscode-azurestaticwebapps",
-    "version": "0.10.1-alpha.1",
+    "version": "0.10.1-alpha.2",
     "lockfileVersion": 2,
     "requires": true,
     "packages": {
         "": {
             "name": "vscode-azurestaticwebapps",
-            "version": "0.10.1-alpha.1",
+            "version": "0.10.1-alpha.2",
             "license": "SEE LICENSE IN LICENSE.md",
             "dependencies": {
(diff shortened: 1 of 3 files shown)


nturinski (approved):
Though would it make sense to put the static export under the static generator group?

anthonychu:
@alexweininger can you confirm the branding for Next.js? I think it should be lowercase js.

alexweininger:
> can you confirm the branding for Next.js? I think it should be lowercase js.

@anthonychu 
Yup, I'll change it. I swear it used to be Next.JS 😅

alexweininger:
> Though would it make sense to put the static export under the static generator group?

I had this, but it looks better with the Next.js options grouped together. And technically it's a framework so...

nturinski (approved):

PR #659, opened 2022-04-15
Get all workflowruns and filter based off branch

nturinski:
After some investigation, seems like new repos don't support branch as a query param (it returns with an empty array).  I'm not surprised since GitHub itself doesn't have a view to only look at actions based on branch, so I think it makes more sense to just grab all the actions and filter based off the branch instead.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
     "name": "vscode-azurestaticwebapps",
     "displayName": "Azure Static Web Apps",
     "description": "%staticWebApps.description%",
-    "version": "0.10.1-alpha.1",
+    "version": "0.10.1-alpha.2",
     "publisher": "ms-azuretools",
     "icon": "resources/azure-staticwebapps.png",
     "aiKey": "AIF-d9b70cd4-b9f9-4d70-929b-a071c400b217",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -2,7 +2,7 @@
     "name": "vscode-azurestaticwebapps",
     "displayName": "Azure Static Web Apps",
     "description": "%staticWebApps.description%",
-    "version": "0.10.1-alpha.1",
+    "version": "0.10.1-alpha.2",
```
bwateratmsft, line 5: inb4 it wants the change in `package-lock.json` too

nturinski: Don't give Alex any ideas...

alexweininger: You should run `npm version prerelease` and it'll take care of everything for you and it'll commit.

https://docs.npmjs.com/cli/v8/commands/npm-version

bwateratmsft (approved):

package.json
alexweininger: > Don't give Alex any ideas...

Oh, we def need to add that... 😆

alexweininger:
Oh we can release this next week too! Perfect timing lol

PR #708, opened 2022-06-20
Disable Next.js (SSR)

### LINKED ISSUE #695: Support only static export in Next.js Build Presets
Currently the VSCode build preset has two options - one for SSR and one for Static Export. 
Until the SSR changes are in picture, we need to disable that option as a build preset and only have **static export** with the existing preset values
image
app-location: "/"
api-location: "api"
output-location: "out"

cc// @alexweininger @nturinski

MicroFish91:
Closes #695

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
     "name": "vscode-azurestaticwebapps",
     "displayName": "Azure Static Web Apps",
     "description": "%staticWebApps.description%",
-    "version": "0.11.1-alpha.1",
+    "version": "0.11.1-alpha.2",
     "publisher": "ms-azuretools",
     "icon": "resources/azure-staticwebapps.png",
     "aiKey": "AIF-d9b70cd4-b9f9-4d70-929b-a071c400b217",
--- a/src/commands/createStaticWebApp/BuildPresetListStep.ts
+++ b/src/commands/createStaticWebApp/BuildPresetListStep.ts
@@ -12,7 +12,12 @@ import { IStaticWebAppWizardContext } from './IStaticWebAppWizardContext';
 export class BuildPresetListStep extends AzureWizardPromptStep<IStaticWebAppWizardContext> {
     public async prompt(context: IStaticWebAppWizardContext): Promise<void> {
         const placeHolder: string = localize('choosePreset', 'Choose build preset to configure default project structure');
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -2,7 +2,7 @@
     "name": "vscode-azurestaticwebapps",
     "displayName": "Azure Static Web Apps",
     "description": "%staticWebApps.description%",
-    "version": "0.11.1-alpha.1",
+    "version": "0.11.1",
```
nturinski, line 5: The main reason we don't want to change this version yet is because 

a.) We might try to include more fixes before we release
b.) When CTI tests our builds, it'll create telemetry entries based off of that. We don't want to start collecting telemetry because this might not be what actually gets released in the wild

src/commands/createStaticWebApp/BuildPresetListStep.ts
```diff
@@ -12,7 +12,12 @@ import { IStaticWebAppWizardContext } from './IStaticWebAppWizardContext';
 export class BuildPresetListStep extends AzureWizardPromptStep<IStaticWebAppWizardContext> {
     public async prompt(context: IStaticWebAppWizardContext): Promise<void> {
         const placeHolder: string = localize('choosePreset', 'Choose build preset to configure default project structure');
-        const picks: IAzureQuickPickItem<IBuildPreset | undefined>[] = buildPresets.map((pb) => { return { label: pb.displayName, data: pb, group: pb.group === 'ssg' ? localize('ssg', 'Static site generator') : localize('framework', 'Framework') }; });
+
+        // Todo: Remove SSR filter when support is added
+        // https://github.com/microsoft/vscode-azurestaticwebapps/issues/709
+        const picks: IAzureQuickPickItem<IBuildPreset | undefined>[] = buildPresets
+            .filter((pb) => pb.id !== "nextjs")
+            .map((pb) => { return { label: pb.displayName, data: pb, group: pb.group === 'ssg' ? localize('ssg', 'Static site generator') : localize('framework', 'Framework') }; });
```
alexweininger, line 20: Nit: I think the variable name on these lines should be `bp` short for build preset instead of `pb`. I think we made a typo when this code was written originally.

alexweininger (dismissed):

alexweininger (approved):

PR #710, opened 2022-06-20
Remove bad semver check to fix CLI update notification

alexweininger:
1. This check currently prevents the notification to update the CLI from appearing if the installed version and the newest version don't have the same major version number 🤦 .
2. The logic was most likely [copied from here in Functions](https://github.com/microsoft/vscode-azurefunctions/blob/411ece5f9453af075c1ff48c70aec349f5942a47/src/funcCoreTools/validateFuncCoreToolsIsLatest.ts#L79), and shouldn't be needed for SWA. Func core tools follows a different versioning method, but the SWA CLI follows semver.
3. Should increase the % of users on v1 of the CLI, which will allow us to remove/improve a lot of code.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "vscode-azurestaticwebapps",
-    "version": "0.11.1-alpha.1",
+    "version": "0.11.1-alpha.2",
     "lockfileVersion": 2,
     "requires": true,
     "packages": {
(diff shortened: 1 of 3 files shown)


nturinski (approved):
Good catch. That'll be great for the release

alexweininger:
I'll merge after #708

PR #727, opened 2022-07-14
Add Action to create version bump PR after releasing

alexweininger:
Triggered when we publish a release on GitHub (can also manually trigger it). Creates a PR [like this one](https://github.com/alexweininger/bumper-test/pull/25) that bumps the prerelease version.

It uses [my own Action task](https://github.com/marketplace/actions/bump-prerelease-version) I published since I couldn't find one that already did what I needed. It has no dependencies.

If we like this, then I'll add it to all the other repos.


bwateratmsft (approved):
Neat!

PR #759, opened 2022-11-10
Allow creating .NET 7.0 Functions

### LINKED ISSUE #758: Add support for Node.js 18 and .NET 7
Discussed in our sync. Trying to land this before the Thanksgiving break.

alexweininger:
This will allow users to create .NET 7.0 Functions. However, if they have .NET 6.0 and .NET 7.0 installed locally, then we will still default to .NET 6.0.

Fixes #758

### DIFF
--- a/src/commands/createHttpFunction.ts
+++ b/src/commands/createHttpFunction.ts
@@ -34,6 +34,6 @@ export async function createHttpFunction(context: IActionContext): Promise<void>
         templateId: 'HttpTrigger',
         languageFilter: /Python|C\#|^(Java|Type)Script$/i,
         functionSettings: { authLevel: 'anonymous' },
-        targetFramework: ['netcoreapp3.1', 'net6.0'] // Will only work on functions api v1.4.0, but won't hurt on v1.3.0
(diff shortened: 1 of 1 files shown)


bwateratmsft (dismissed):
Depending which you merge first, this will definitely yield a merge conflict with #760.

nturinski:
> This will allow users to create .NET 7.0 Functions. However, if they have .NET 6.0 and .NET 7.0 installed locally, then we will still default to .NET 6.0.
> 
> Fixes #758

Does this mean the user never gets prompted for which .NET they want to use? And we'll only use .NET 7 if they have it installed?

If that's the case, I think we would want to change this behavior. I believe the Func CLI lets you create .NET 7 triggers with a .NET 6 runtime.

alexweininger:
> Does this mean the user never gets prompted for which .NET they want to use? And we'll only use .NET 7 if they have it installed?

Yes.

> If that's the case, I think we would want to change this behavior. I believe the Func CLI lets you create .NET 7 triggers with a .NET 6 runtime.

I agree, but wanted to at least enable it for now. Letting users pick will be a larger change (and need to happen in Functions). I will create a new issue for that.

nturinski (approved):

PR #797, opened 2023-03-03
Use remote repositories API to enable create static web app in vscode.dev

nturinski:
Remote Repositories lets users open GitHub repos in vscode.dev or in a virtual workspace. By leveraging their API, I was able to basically enable full SWA create support on vscode.dev with minimal refactoring.

There are a lot of files related to the git API, but I will move them in another PR since it would make reviewing changes a pain.

`path.join` won't work in vscode.dev, so I replaced it where relevant. `createHttpFunction` still uses it since it can only be used on desktop, though we can probably just remove it all eventually.

### DIFF
--- a/src/RemoteRepoApi.ts
+++ b/src/RemoteRepoApi.ts
@@ -5,9 +5,10 @@
 
 import { AzureExtensionApi, createApiProvider } from '@microsoft/vscode-azext-utils';
 import * as vscode from 'vscode';
+import { AzureExtensionApiProvider } from '../azext-utils-api';
 import { revealTreeItem } from './commands/api/revealTreeItem';
-import { APIState, PublishEvent } from './git';
-import { API, IGit, PostCommitCommandsProvider, Repository } from './rrapi';
+import { APIState, IGit, PublishEvent } from './git';
+import { API, PostCommitCommandsProvider, Repository } from './rrapi';
 
 export const enum RefType {
     Head,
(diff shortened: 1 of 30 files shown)


nturinski:
Note: Still to do
- When there is no open folder in vscode.dev, the prompt still asks to open an existing projecting, which leads to your local directory. It should lead to a remote repo.
- There is a Git Clone button that doesn't work
- If you have a local folder opened in vscode.dev, it will fail with the error `Cannot read properties of undefined (reading 'openRepository')`

src/RemoteRepoApi.ts
```diff
@@ -0,0 +1,179 @@
...
+        const handle = this._nextHandle();
+        this._providers.set(handle, provider);
+
+        this._disposables.push(provider.onDidCloseRepository(e => this._onDidCloseRepository.fire(e)));
+        this._disposables.push(provider.onDidOpenRepository(e => {
+            this._updateReposContext();
+            this._onDidOpenRepository.fire(e);
+        }));
+        if (provider.onDidChangeState) {
+            console.log(provider.onDidChangeState);
```
bwateratmsft, line 127: Test line?

nturinski: Oh yeah, that API is pulled from somewhere else. Will clean it up 👍

bwateratmsft: Oh, this isn't our source? The original source should be noted in a comment

nturinski: Thanks for reminding me; I was meaning to ask.

The repo that this was pulled from is private, so I wasn't sure if I should list the original source. It's also not exactly the same since we didn't need everything that they implemented, but just needed a sample for the `registerGitProvider` function.

PR #801, opened 2023-03-13
Migration work for vscode.dev

nturinski:
I'm planning on moving the git API files, but I didn't want to change the files and move them in the same PR.


azext-utils-api.d.ts
```diff
@@ -0,0 +1,23 @@
...
+// For now this file needs to be copied/pasted into your repo if you want the types. Eventually we may put it somewhere more distributable.
+
+export interface AzureExtensionApi {
+    /**
+     * The API version for this extension. It should be versioned separately from the extension and ideally remains backwards compatible.
+     */
+    apiVersion: string;
+}
+
+export interface AzureExtensionApiProvider {
```
bwateratmsft, line 15: Can we just get these declarations straight from the package? (Once we've actually merged and published)

nturinski: Oh, it's actually in `apiUtils` from `@microsoft/vscode-azureresources-api`

package.json
```diff
@@ -485,29 +485,31 @@
         "mocha": "^10.0.0",
         "mocha-junit-reporter": "^2.0.0",
         "mocha-multi-reporters": "^1.1.7",
+        "process": "^0.11.10",
         "ts-node": "^7.0.1",
         "typescript": "^4.3.5",
         "vsce": "^2.11.0",
+        "vscode-azurekudu": "^0.2.3",
```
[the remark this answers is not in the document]
nturinski: I can actually get rid of this! Since I got rid of appservice package from here 😀

src/vscode-azext-azureappservice/AppSettingTreeItem.ts
```diff
@@ -0,0 +1,134 @@
...
+import { AzExtTreeItem, DialogResponses, IActionContext, TreeItemIconPath, createContextValue } from '@microsoft/vscode-azext-utils';
+import { ThemeIcon } from 'vscode';
+import { ext } from '../extensionVariables';
+import { localize } from '../utils/localize';
+import { AppSettingsTreeItem, validateAppSettingKey } from './AppSettingsTreeItem';
+
+/**
+ * NOTE: This leverages a command with id `ext.prefix + '.toggleAppSettingVisibility'` that should be registered by each extension
+ */
+export class AppSettingTreeItem extends AzExtTreeItem {
```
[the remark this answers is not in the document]
nturinski: The only code the SWA extension was using from the `appservice` tools package were the `AppSettingTreeItems`. So I just went ahead and added those files to the extension itself.

package.json
```diff
@@ -485,29 +485,31 @@
...
-        "@microsoft/vscode-azext-azureutils": "^0.3.4",
-        "@microsoft/vscode-azext-utils": "^0.3.21",
+        "@microsoft/vscode-azext-azureutils": "0.4.0-alpha.1",
+        "@microsoft/vscode-azext-utils": "0.4.3-alpha",
+        "@microsoft/vscode-azureresources-api": "^2.0.3",
         "@octokit/rest": "^18.5.2",
+        "buffer": "^6.0.3",
         "dayjs": "^1.11.0",
         "git-url-parse": "^13.1.0",
         "open": "^8.0.4",
```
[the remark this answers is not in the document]
nturinski: No, we have to provide the polyfill for the webpack config. I didn't put it in the dev package because not every extension needs it.

package.json
```diff
@@ -494,20 +496,20 @@
...
         "semver": "^7.3.5",
         "vscode-nls": "^4.1.1",
         "vscode-uri": "^3.0.2",
-        "yaml": "^1.10.2"
+        "yaml": "^2.2.1"
     },
     "extensionDependencies": [
-        "ms-vscode.azure-account",
```
nturinski, line 510: I need to figure out how to have a separate package.json to have different extensionDependencies for .dev and desktop

alexweininger: I thought we determined this wasn't possible, did something change?

src/vscode-azext-azureappservice/AppSettingTreeItem.ts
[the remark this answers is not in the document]
bwateratmsft: That doesn't seem like a great reason to stop sharing code...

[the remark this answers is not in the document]
alexweininger: I figure we can extract the app settings code into a new package. I don't think we need to be shy about creating new packages. Having more smaller packages is better in many ways.

package.json
nturinski: I have a feeling that there has to be someway to do it if they recommend that you don't 🤔

bwateratmsft: > We do not recommend that extensions ship a different package.json for the web platform or to remove parts of the VSIX that do not work in the web.

Can't we just turn Azure Account into a soft-hard dependency?

bwateratmsft: Or potentially, just auto-install it after activation when on desktop (if it's not present already).

alexweininger: > Or potentially, just auto-install it after activation when on desktop (if it's not present already).

Oh right, I forgot we could do that.

nturinski: > Or potentially, just auto-install it after activation when on desktop (if it's not present already).

I'm okay to do that. It's a slightly worse user-experience, but not that bad. Also, we can include (if it's not already) accounts in the Azure extension pack.

src/IGit.ts
```diff
@@ -0,0 +1,26 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { Disposable, Event, Uri } from "vscode";
+import { APIState, PublishEvent, Repository } from "./git";
+import { Metadata, PostCommitCommandsProvider } from "./rrapi";
+
```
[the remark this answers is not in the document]
nturinski: The types are actually from here: https://github.com/microsoft/vscode/blob/48eea9e7bc7882a0bc44ac0e4ab6f63df05910f5/extensions/git/src/api/git.d.ts#L284C8-L303

They just call it an API-- it's the same code that used to be here: https://github.com/microsoft/vscode-azurestaticwebapps/blob/main/src/git.d.ts#L255-L272 (which is copied from vscode)

I renamed it to IGit though because RemoteRepos implements the same API and I believe that they implemented it as `IGit`.

That being said, I can still link the vscode `git.d.ts` file.

src/RemoteRepoApi.ts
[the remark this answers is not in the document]
nturinski: So we wouldn't want to actually use their code exactly. It's adapted for our use-- GitHub Pull Request extension handles the providers a little bit differently than we are: https://github.com/microsoft/vscode-pull-request-github/blob/b6c48e64b65d5d2934391a796ad0f5a81b7c12be/src/api/api1.ts#L104

The main thing is that the extension needs to export `registerGitProvider` that the RemoteRepo extension can call so that we have access to their API.

package.json
nturinski: I figured there's no point in having Accounts as a dependency here since Resources is already a dependency.

bwateratmsft: Agreed

bwateratmsft (approved):

alexweininger (approved):

PR #836, opened 2023-06-08
Update shared packages to v2

alexweininger:
The v2 versions of these packages are fully migrated to Track 2 SDKs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -467,7 +467,7 @@
     },
     "devDependencies": {
         "@microsoft/eslint-config-azuretools": "^0.1.0",
-        "@microsoft/vscode-azext-dev": "^1.0.3",
+        "@microsoft/vscode-azext-dev": "^2.0.0",
         "@octokit/types": "^6.13.0",
         "@types/fs-extra": "^8.1.1",
         "@types/git-url-parse": "^9.0.0",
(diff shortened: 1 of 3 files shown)


package.json
bwateratmsft, line 1: Don't forget `@microsoft/vscode-azext-dev`!

src/commands/deleteStaticWebApp/IDeleteWizardContext.ts
```diff
@@ -3,8 +3,7 @@
 *  Licensed under the MIT License. See License.txt in the project root for license information.
 *--------------------------------------------------------------------------------------------*/
 
-import { ISubscriptionContext } from "@microsoft/vscode-azext-dev";
-import { ExecuteActivityContext, IActionContext } from "@microsoft/vscode-azext-utils";
+import { ExecuteActivityContext, IActionContext, ISubscriptionContext } from "@microsoft/vscode-azext-utils";
```
bwateratmsft, line 6: Both of these imports could just be `import type` though it's not that important.

alexweininger:
In theory we can now use the 
@azure-rest/arm-appservice SDK https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/appservice/arm-appservice-rest/README.md

bwateratmsft (approved):

Issue #761, opened 2022-11-10
Let users specify .NET version when adding an HTTP Function


alexweininger:
See #759 for details, but tl;dr is we need to not only allow creating .NET 7.0 Functions, we also need to give users a choice between .NET 6.0 and 7.0 instead of defaulting to 6.0.

thomasgauvin:
Now with .NET 8, we should allow users the ability to choose .NET 8, and default to .NET 8 (currently we default to .NET 6) @nturinski

PR #893, opened 2024-05-15
Allow .NET Runtime Selection When Creating Http Function

Timothyw0:
Resolves GitHub issue: 
GitHub  https://github.com/microsoft/vscode-azurestaticwebapps/issues/761

By removing the `targetFramework` option when creating an Azure Function using the extension, the context will be initialized with a `null` workerRuntime property which will then prompt the user to select which .NET runtime stack to use as the next step.
image

### DIFF
--- a/src/commands/createHttpFunction.ts
+++ b/src/commands/createHttpFunction.ts
@@ -33,7 +33,6 @@ export async function createHttpFunction(context: IActionContext): Promise<void>
         suppressCreateProjectPrompt: true,
         templateId: 'HttpTrigger',
         languageFilter: /Python|C\#|^(Java|Type)Script$/i,
-        functionSettings: { authLevel: 'anonymous' },
-        targetFramework: ['net6.0', 'net7.0'] // Will only work on functions api v1.4.0, but won't hurt on v1.3.0
+        functionSettings: { authLevel: 'anonymous' }
     });
 }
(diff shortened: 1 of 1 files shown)


src/commands/createHttpFunction.ts
```diff
@@ -33,7 +33,6 @@ export async function createHttpFunction(context: IActionContext): Promise<void>
         suppressCreateProjectPrompt: true,
         templateId: 'HttpTrigger',
         languageFilter: /Python|C\#|^(Java|Type)Script$/i,
-        functionSettings: { authLevel: 'anonymous' },
-        targetFramework: ['net6.0', 'net7.0'] // Will only work on functions api v1.4.0, but won't hurt on v1.3.0
+        functionSettings: { authLevel: 'anonymous' }
```
motm32, line 36: Thanks for putting up this PR! Looking a little more into this I am not sure this will be the best solution since there are .NET runtimes that are not supported. We would likely just want to add .NET 8 to the `targetFramework` and then possibly make that the default runtime like suggested in this [comment](https://github.com/microsoft/vscode-azurestaticwebapps/issues/761#:~:text=Now%20with%20.NET%208%2C%20we%20should%20allow%20users%20the%20ability%20to%20choose%20.NET%208%2C%20and%20default%20to%20.NET%208%20(currently%20we%20default%20to%20.NET%206)%20%40nturinski).  Feel free to make those changes yourself or we can try and get to it sometime in the future 😊

Timothyw0: Ah, that makes sense! Totally understood, I will change this PR to default to .NET 8 as the `targetFramework` for now and continue to look into the HTTP Function selection changes. That will require some change to be done in the VS Code Azure Function Extension to populate the Subwizard for our specific .NET runtime selector then :)

Thank you for your review!

Timothyw0:
Abandoning this PR for now to approach the .NET runtime selection at a later time as @motm32 suggested. I will raise a new PR to default the .NET runtimes to 8.0

PR #908, opened 2024-07-25
Update @vscode/extension-telemetry package

alexweininger:


nturinski (dismissed):
You may need to update the dev package too to get rid of these build error typings.

motm32 (approved):

PR #1006, opened 2025-11-11
eng: Linting: manual fixes

bwateratmsft:


src/git.ts
bwateratmsft: This file had to be renamed as it was not actually just declarations.

PR #1010, opened 2025-11-12
Use the fancy new eng package!

bwateratmsft:
~~TODO: Release the eng package and install it~~


.vscodeignore
bwateratmsft: TODO: Need to verify VSIX before/after.

